PR #62, opened 2018-10-26
Switch to weather subscriptions

cbarber:
This PR switches the front end weather widget from polling the weather graphql endpoint to receive weather updates via a web socket subscription. A backend worker is added to run every 5 minutes to poll the locations that clients are currently subscribed to.

Caveats:
* Some of the graphql channel still uses class instance variables to store data. This should eventually be switched to redis.
* Poller interval is currently hardcoded to 5 minutes.

### DIFF
--- a/spec/workers/weather_poller_worker_spec.rb
+++ b/spec/workers/weather_poller_worker_spec.rb
@@ -1,0 +1,38 @@
+require 'rails_helper'
+RSpec.describe WeatherPollerWorker, type: :worker do
+  before {
+    allow(ForecastIO).to receive(:forecast) { "" }
+    allow_any_instance_of(Redis).to(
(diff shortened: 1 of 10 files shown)


spec/workers/weather_poller_worker_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  describe "a poller" do
+    it "will query uniq locations from redis " do
+      expect(WeatherPollerWorker.new.locations).to(
+        eq([WeatherPollerWorker::Location.new("41.823979", "-71.412834")])
+      )
+    end
+
+    it "polling a location will publish the darkshy response" do
```
MojoJosh: "darkshy"

PR #68, opened 2018-11-06
fix: configure sidekiq to use redis_url env

cbarber:
Configures sidekiq and sidekiq-repeat to connect to redis using the REDIS_URL env variable (used in Heroku).


cbarber:
https://github.com/mojotech/helios2/blob/master/docker-compose.yml#L21

PR #69, opened 2018-11-26
falling blocks

chewnoill:
dec-10-2018 16-19-01


app/javascript/components/widgets/numbers/fallingblocks.jsx
```diff
@@ -0,0 +1,62 @@
...
+  for (let i = array.length - 1; i > 0; i -= 1) {
+    const j = Math.floor(Math.random() * (i + 1));
+    [array[i], array[j]] = [array[j], array[i]];
+  }
+}
+
+class FallingBlocks extends React.Component {
+  constructor(props) {
+    super(props);
+    this.blocks = [];
```
cbarber: I think we could skip the array creation and shuffle by keeping pr, commit, slack counts in state and randomly decrementing any of the fields that is greater than zero.

app/javascript/components/widgets/numbers/fallingblocks.jsx
```diff
@@ -0,0 +1,62 @@
...
+  }
+  componentDidMount() {
+    this.fallingBlocks = UnityLoader.instantiate(
+      'falling-blocks-canvas',
+      'unity/webgl.json',
+      {
+        Module: {
+          // wait for unity splash screen to go away before
+          // adding blocks
+          preRun: [() => setTimeout(this.addBlocks, 4500)],
```
cbarber: It looks like Unity provides a flag to indicate if the splash screen is finished: https://docs.unity3d.com/ScriptReference/Rendering.SplashScreen-isFinished.html

app/javascript/components/widgets/numbers/panel.jsx
```diff
@@ -1,6 +1,11 @@
 import React from 'react';
 import Wrapper from '../../panel-wrapper';
+import FallingBlocks from './fallingblocks';
 
-export const Numbers = () => <Wrapper />;
+export const Numbers = () => (
+  <Wrapper>
+    <FallingBlocks pr={143} commit={160} slack={500} />
```
cbarber: This can wrapped with a graphql query. Have a look at `app/javascript/components/widgets/numbers/carousel.jsx`

There are rake tasks to simulate pr and commits:
```
rake simulate:commit[count]             # Simulates a new commit on github
rake simulate:pull_request              # Simulates opening a new pull request on github

```

public/unity/webgl.json
```diff
@@ -0,0 +1,13 @@
+{
+"companyName": "DefaultCompany",
+"productName": "falling squares",
+"dataUrl": "webgl.data.unityweb",
+"wasmCodeUrl": "webgl.wasm.code.unityweb",
```
cbarber: Is there source code for this?

cbarber:
We might have to adjust the layout as well. The unity canvas has a big gap from the top left at 100% zoom.

app/javascript/components/widgets/numbers/fallingblocks.jsx
will-cohen-ck:
```suggestion
          preRun: [() => setTimeout(this.addBlocks, 1500)],
```

app/javascript/components/widgets/numbers/fallingblocks.jsx
chewnoill: I like it, much simpler. 

This ends up changing the distribution of the block types so each group is weighted evenly instead of having them weighted by the number of blocks of each type. If one group is much larger then the others, once the other types have been depleted you get a bunch of that type all at once.

app/javascript/components/widgets/numbers/fallingblocks.jsx
```diff
@@ -0,0 +1,62 @@
...
+        },
+      },
+    );
+  }
+
+  addBlocks = () => {
+    const block = this.blocks.pop();
+    this.fallingBlocks.SendMessage('background', 'addObject', block);
+    if (this.blocks.length > 0) {
+      setTimeout(this.addBlocks, 10);
```
cbarber:
```suggestion
      const TEN = 10;
      setTimeout(this.addBlocks, TEN);
```

app/javascript/components/widgets/numbers/fallingblocks.jsx
cbarber: You can maintain standard distribution by selecting a random integer from the range of 0..sum(counts) and selecting the decrement index based on the ranges.

chewnoill:
thanks @cbarber, updated

app/javascript/components/widgets/numbers/fallingblocks.jsx
chewnoill: I'll stick with the simpler way for now, if we decide its important we can change it later

public/unity/webgl.json
chewnoill: there is. Not sure where it should live; I'm thinking a separate repo

PR #70, opened 2018-12-07
fix: bump rails to 5.2.2 for security updates

cbarber:


Gemfile.lock
```diff
@@ -172,37 +172,37 @@ GEM
...
-      actioncable (= 5.2.0)
-      actionmailer (= 5.2.0)
-      actionpack (= 5.2.0)
-      actionview (= 5.2.0)
-      activejob (= 5.2.0)
-      activemodel (= 5.2.0)
-      activerecord (= 5.2.0)
-      activestorage (= 5.2.0)
-      activesupport (= 5.2.0)
+    rails (5.2.2)
```
will-cohen-ck: Should I expect a change for this in `Gemfile` also? Updating just a lockfile seems unexpected.

cbarber: Nope, the Gemfile is a set of hints and Gemfile.lock is the dependency resolution. `'~> 5.2.0'` equates to >= 5.2.0 *and* < 5.3.0 which is satisified by 5.2.2. I could bump the Gemfile to `~> 5.2.2` but it seems redundant with the Gemfile.lock.

cbarber: Oh, and the command is just `bundle update <optional gem|omit for all gems>`

PR #81, opened 2019-05-12
Updates to Dockerfile, docker-compose.yml

mfrost:
Ended up rewriting a lot of the Docker setup. I would still like to figure out how to move the `node_modules` out of the project directory so we don't have to deal with shuffling those packages around in the volume.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,11 +9,12 @@ RUN useradd -ms /bin/bash -d /app app
 RUN mkdir /app/secure && chown app:app /app/secure
 RUN mkdir /app/tmp && chown app:app /app/tmp
 WORKDIR /app
+
+COPY Gemfile /app/Gemfile
+COPY Gemfile.lock /app/Gemfile.lock
+RUN bundle install --jobs=4
+
 USER app
 RUN bundle config git.allow_insecure true
-COPY --chown=app vendor /app/vendor
-COPY --chown=app Gemfile /app/Gemfile
-COPY --chown=app Gemfile.lock /app/Gemfile.lock
-RUN bundle install --jobs=4 --path vendor/bundle
 RUN yarn install
 COPY --chown=app . /app
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,29 @@
 FROM ruby:2.5.1
+
+ARG USER_ID=1000
+ARG GROUP_ID=1000
+
 RUN curl -sL https://deb.nodesource.com/setup_8.x | bash &&\
     apt-get update -qq && apt-get install -yq build-essential nodejs unzip &&\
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -8,13 +8,12 @@ RUN curl -sL https://deb.nodesource.com/setup_8.x | bash &&\
...
 RUN groupadd -g ${GROUP_ID} developers &&\
     useradd -l -u ${USER_ID} -g ${GROUP_ID} developer &&\
-    install -d -m 0755 -o developer -g developers /home/developer
-
-RUN mkdir /helios && chown developer:developers /helios
-RUN mkdir /helios/secure && chown developer:developers /helios/secure
-RUN mkdir /helios/tmp && chown developer:developers /helios/tmp
+    install -d -m 0755 -o developer -g developers /home/developer &&\
+    mkdir -p /helios/secure /helios/tmp &&\
+    chown -R developer:developers /helios
```
cbarber: Should we also add:
```
      args:
        USER_ID: $(id -u ${USER})
        GROUP_ID: $(id -g ${USER})
``` 
to docker-compose.yml?

mfrost: I thought about that, but I'm not sure what OS X support is like there. Also, some research suggests that they are not really bind mounts in OS X, and assume the permissions of the host user. Needs testing.

djkirby:
verified user perms looked good for me on os x but it seems to have a problem w/ the migrations

Screen Shot 2019-05-15 at 10 12 39 AM

mfrost:
@djkirby you can bring the database up to speed with `docker-compose exec web rails db:migrate` if the container is already running.

cbarber:
Oooohhhh. That's my bad @djkirby. I thought `rake db:migrate` was still executed via `docker-compose up` but I didn't RTF-PR :)

docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3'
 services:
   web:
     build: .
-    command: bash -lc "(rm -rf tmp/pids/ || true) && yarn install && bundle install --path vendor/bundle && bin/rails db:migrate db:seed RAILS_ENV=development && bundle exec foreman start -f Procfile.development"
+    command: bash -lc "(rm -rf tmp/pids/ || true) && yarn install && bundle exec foreman start -f Procfile.development"
```
chewnoill: why are we removing the migration from here?

aaddaamm:
@cbarber @mfrost has there been any traction on this PR? is it waiting on something?

docker-compose.yml
mfrost: I believe there are cases where you might want to boot the app without always running migrations. They are easily executed with `docker-compose exec` so I removed them here.

PR #82, opened 2019-05-15
slack event numbers

mariana0pachon:
- Added ruby controller for slack message events to be added to the database
- Added graphql query to count events in the current week and display the number in the panel and carousel
- Added env variables and authentication controller for slack mojotech workspace

### DIFF
--- a/.env.local.sample
+++ b/.env.local.sample
@@ -3,3 +3,7 @@ PRIMARY_CITY_NAME=Providence
 WIFI_NAME=mojotech-guest
 WIFI_PASSWORD=p@ssw0rd
 BATHROOM_CODE=0000*
+SLACK_CLIENT_ID=
+SLACK_API_SECRET==
+SLACK_REDIRECT_URI=https://mojotech-helios2.herokuapp.com/web_hooks/create_slack_auth
+SLACK_VERIFICATION_TOKEN=
--- a/.env.local.sample
+++ b/.env.local.sample
@@ -4,6 +4,6 @@ WIFI_NAME=mojotech-guest
 WIFI_PASSWORD=p@ssw0rd
 BATHROOM_CODE=0000*
 SLACK_CLIENT_ID=
-SLACK_API_SECRET==
+SLACK_API_SECRET=
 SLACK_REDIRECT_URI=https://mojotech-helios2.herokuapp.com/web_hooks/create_slack_auth
 SLACK_VERIFICATION_TOKEN=
--- a/app/javascript/components/widgets/numbers/panel.jsx
+++ b/app/javascript/components/widgets/numbers/panel.jsx
@@ -1,6 +1,110 @@
 import React from 'react';
-import Wrapper from '../../panel-wrapper';
+import PropTypes from 'prop-types';
+import { Query } from 'react-apollo';
+import gql from 'graphql-tag';
+import styled from 'styled-components';
+import { over, lensPath, inc } from 'ramda';
+import pluralize from 'pluralize';
+import {
+  colors,
+  weights,
+  fontSizes,
+  spacing,
+  fonts,
+  leftPanelWidth,
+} from '../../../lib/theme';
 
-export const Numbers = () => <Wrapper />;
+const LoadingMessage = () => <p>Loading...</p>;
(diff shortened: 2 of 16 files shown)


app/javascript/lib/datetime.js
```diff
@@ -67,3 +67,10 @@ export const parseDay = datetime => {
     weekday: 'long',
   });
 };
+
+export const latestDay = day => {
+  const t = new Date();
+  return new Date(t.setDate(t.getDate() - t.getDay() + day))
+    .toISOString()
+    .slice(0, 10);
+};
```
djkirby: wonder if at this point we should introduce something like `date-fns`

app/javascript/components/widgets/numbers/panel.jsx
```diff
@@ -1,6 +1,119 @@
...
+`;
+
+const NumberWrapper = styled.div`
+  display: flex;
+  flex-wrap: no-wrap;
+  justify-content: space-between;
+  width: 5em;
+`;
+
+class SubscribedEvents extends React.PureComponent {
```
geraldalewis: I'd use a `function` here since this class has no state/lifecycle methods.

```javascript
const SubscribedEvents = ({ githubPull, githubCommit, slackMessage }) => ...
```

app/javascript/components/widgets/numbers/panel.jsx
```diff
@@ -1,6 +1,119 @@
...
+  render() {
+    const { githubPull, githubCommit, slackMessage } = this.props;
+
+    const commitText = pluralize('Commit', githubCommit);
```
geraldalewis: The inconsistent case (titlecase vs. the following, which are lowercase) here bumped me when I first saw it. Not asking for a change, but food for thought:

```javascript
const commitText = pluralize('commit', githubCommit);
...
<BigNumberSubtitle style={{ color: colors.pink }}>
  {titlecase(commitText)} // not sure if this proj has a dep that can provide `titlecase`...
</BigNumberSubtitle>
```

or even:

```javascript
<BigNumberSubtitle style={{ color: colors.pink }}>
  {pluralize('Commit', githubCommit)}
</BigNumberSubtitle>
```

The code as-is looks very clean, so I'm not advocating for those changes (just something that crossed my mind).

app/javascript/components/widgets/numbers/panel.jsx
```diff
@@ -1,6 +1,119 @@
...
+        <NumberWrapper>
+          <div>
+            <BigNumber>{githubCommit}</BigNumber>
+            <BigNumberSubtitle style={{ color: colors.pink }}>
+              {commitText}
+            </BigNumberSubtitle>
+          </div>
+          <div>
+            <BigNumber>{githubPull}</BigNumber>
+            <BigNumberSubtitle style={{ color: colors.yellow }}>
```
geraldalewis: The `chore: add colors for number panel to theme` commit (where these colors are added) comes _after_ this commit. I'd just reorder these two commits.

Alternatively: Some devs have strong feelings that a commit that introduces new code should also _use_ that new code, so some would advise that these two commits be squashed together.

I personally feel that keeping them separate helps me review and is arguably more atomic.

mariana0pachon: Yeah, doing it before makes more sense. We talked through some commit styling things with Craig that I am working on. 👍

app/javascript/lib/datetime.js
aaddaamm: we could probably circle back and retro fit this with date-fns right? I can add a ticket for that in the backlog.

djkirby: yea not blocking

.env.local.sample
```diff
@@ -3,3 +3,7 @@ PRIMARY_CITY_NAME=Providence
 WIFI_NAME=mojotech-guest
 WIFI_PASSWORD=p@ssw0rd
 BATHROOM_CODE=0000*
+SLACK_CLIENT_ID=
+SLACK_API_SECRET=
+SLACK_REDIRECT_URI=https://mojotech-helios2.herokuapp.com/web_hooks/create_slack_auth
```
djkirby: I think we should leave this out of the sample .env since it's a public repo

PR #89, opened 2019-05-24
ui(stats): format stat numbers w/ commas

djkirby:
Screen Shot 2019-05-24 at 8 30 02 AM

### DIFF
--- a/app/javascript/components/widgets/numbers/carousel.jsx
+++ b/app/javascript/components/widgets/numbers/carousel.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Query } from 'react-apollo';
 import gql from 'graphql-tag';
+import numeral from 'numeral';
 import { over, lensPath, inc } from 'ramda';
 import pluralize from 'pluralize';
 import { getMostRecentDay } from '../../../lib/datetime';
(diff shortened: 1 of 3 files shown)


app/javascript/components/widgets/numbers/carousel.jsx
```diff
@@ -32,19 +33,24 @@ const getEventCounts = gql`
...
 class SubscribedEvents extends React.Component {
   componentDidMount() {
     this.props.subscribeToPublishedEvents();
   }
 
   render() {
     const { githubPull, githubCommit, slackMessage } = this.props;
+    const githubPullCount = numeral(githubPull).format(STAT_FORMAT);
+    const githubCommitCount = numeral(githubCommit).format(STAT_FORMAT);
+    const slackMessageCount = numeral(slackMessage).format(STAT_FORMAT);
```
aaddaamm: what is the reason for using this numeral library? could we accomplish this without adding the dependency?

djkirby: 🤷‍♂ probably but that's more code we'd have to maintain, I'd rather let numeral do the formatting personally

aaddaamm: could we use this? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString#Using_toLocaleString

cbarber: I'm ok with 3.8kB added for an expressive way to format numbers
https://bundlephobia.com/result?p=numeral@2.0.6

PR #90, opened 2019-05-24
Feat: Added improved error and loading message handling

theroyalblah:
This PR adds a file render-message that displays a message accompanied by an image of a sun and moon. It also adds default-messages which comes with two preset messages which most of the files were edited to call instead of displaying the actual error raw. It's important to note that render-message always renders to the panel, so the calls to RenderMessage are currently limited to the panels.

Screen Shot 2019-05-24 at 2 48 30 PM
Screen Shot 2019-05-24 at 2 51 12 PM


aaddaamm:
Needs a rebase 🙏

djkirby:
@theroyalblah can you add a screenshot the PR? curious how this looks

app/javascript/components/widgets/messages/render-message.jsx
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+import { GreySubText } from '../../typography';
+import { fontSizes, leftPanelWidth } from '../../../lib/theme';
+import sunMoon from '../../../../assets/images/sunmoon.png';
+
+export const RenderLoadingMessage = () => (
```
djkirby: I feel like we can ditch the `Render` prefix on these components -- I think the rendering is implied

app/javascript/components/widgets/messages/render-message.jsx
```diff
@@ -0,0 +1,71 @@
...
+          x={((widthFloat - parseFloat(imageWidth)) / 2).toString()}
+          y={(heightFloat - parseFloat(imageHeight) + 1).toString()}
+          width={imageWidth}
+          height={imageHeight}
+          xlinkHref={sunMoon}
+          position="absolute"
+        />
+      </svg>
+      <MessageDisplay>
+        <Message />
```
djkirby: should be able to just `{message}` here

theroyalblah:
Here's some screenshots of the features
Screen Shot 2019-05-24 at 2 48 30 PM
Screen Shot 2019-05-24 at 2 51 12 PM

app/javascript/components/widgets/messages/render-message.jsx
geraldalewis: I think that's a good idea. @theroyalblah friendly reminder that it's a good idea to rename the `render-message.jsx` filename as well, and to squash those changes into the original commit.

app/javascript/components/widgets/messages/default-messages.jsx
```diff
@@ -0,0 +1,2 @@
+export const LoadingMessage = () => 'Loading...';
```
thedelchop: I'm still semi-new to React myself, but it strikes me as odd that we'd want to use a Component to return a simple string here.  I think this would be better handled with just straight JS.

geraldalewis: @thedelchop From a consistency perspective, I like using `<LoadingMessage />` instead of needing to interpolate (`{LoadingMessage}`). Another advantage is that it hides the details of implementation from the consumer side, allowing us to update `LoadingMessage` later to be more robust without modifying the call sites (e.g., `export const LoadingMessage = () => <BigLetters>Loading</BigLetters>`)

PR #93, opened 2019-05-28
chore: add Slack integration documentation

mariana0pachon:


INTEGRATIONS.md
```diff
@@ -1,11 +1,37 @@
-### Integrations
+# Integrations
```
geraldalewis: Can we split out the header-level and newline changes into their own commit? That way, it's obvious which changes apply to the cleanup, and which apply to the feature.

PR #94, opened 2019-05-29
Convert datetime.js functions to use date-fns for date calculations

mariana0pachon:
- Wrote tests and implemented functions for datetime using date-fns
- The `timeAndDateForTimezone` and `getStartOfWeek` functions don't have tests because CircleCI runs tests at timezone UTC +0, which is different than the local timezone, and this became a problem when `new Date()` was called. I tried specifying timezone as a parameter but this was only ever used for testing purposes.

### DIFF
--- a/app/javascript/components/widgets/numbers/carousel.jsx
+++ b/app/javascript/components/widgets/numbers/carousel.jsx
@@ -4,7 +4,7 @@ import { Query } from 'react-apollo';
 import numeral from 'numeral';
 import { over, lensPath, inc } from 'ramda';
 import pluralize from 'pluralize';
-import { getMostRecentDay } from '../../../lib/datetime';
+import { getStartOfWeek } from '../../../lib/datetime';
 import { getEventCounts, subscribeEventPublished } from './queries';
 import { LoadingMessage, ErrorMessage } from '../messages/default-messages';
 
@@ -36,10 +36,7 @@ SubscribedEvents.propTypes = {
 };
 
 const Numbers = () => (
-  <Query
-    query={getEventCounts}
-    variables={{ after: getMostRecentDay('Monday') }}
-  >
+  <Query query={getEventCounts} variables={{ after: getStartOfWeek() }}>
     {({ loading, error, data, subscribeToMore }) => {
       if (loading) {
         return <LoadingMessage />;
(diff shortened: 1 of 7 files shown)


thedelchop:
Looks good to me, nice work @mariana0pachon!

app/javascript/lib/datetime.test.js
```diff
@@ -1,11 +1,101 @@
...
+describe('#parseDay', () => {
+  const weekdays = {
+    0: 'Sunday',
+    1: 'Monday',
+    2: 'Tuesday',
+    3: 'Wednesday',
+    4: 'Thursday',
+    5: 'Friday',
+    6: 'Saturday',
+  };
```
djkirby: nit: I think this could be an array,

```js
  const weekdays = [
    'Sunday',
    'Monday',
    'Tuesday',
    'Wednesday',
    'Thursday',
    'Friday',
    'Saturday'
  ];
```

app/javascript/components/widgets/numbers/carousel.jsx
```diff
@@ -5,7 +5,7 @@ import gql from 'graphql-tag';
 import numeral from 'numeral';
 import { over, lensPath, inc } from 'ramda';
 import pluralize from 'pluralize';
-import { getMostRecentDay } from '../../../lib/datetime';
+import { getStartOfWeek } from '../../../lib/datetime';
```
geraldalewis: Love this commit's atomicity. However, we're importing a function here (`getStartOfWeek`) that doesn't yet exist. If we ran `git bisect`, we'd run into problems at this commit :( It's really important to ensure that each commit is a buildable one. I suggest: 

* Modifying `refactor: implement fns with date-fns and test ` to not (yet) remove `getMostRecentDay`, and remove the definition of `getStartOfWeek` from that commit
* reordering this commit to be the final commit, and move the code where you define the `getStartOfWeek` function into this commit

Make sense?

cbarber:
Two approaches you can take with for controlling test data:
* optional parameters only used in tests e.g. `const omeFunc = (date = new Date()) => {.....}`. this is a kind of stubbing 
* mock `new Data` with jest mocking to control what it resolves. this can be powerful but be weary this tends to tightly couple test and method implementations

PR #96, opened 2019-06-03
Support absolute js and jsx imports

mariana0pachon:
- Defined aliases for React directories in the `webpack.config.custom.js` file.
- Created a transform file using `jscodeshift` to transform relative imports to alias imports when an alias is found.
- Automated changing relative to absolute imports in .jsx files with the `npm run relative-to-alias` script. After running this, `prettier` needs to be run in all files because all imports are done with double quotes instead of single quotes.
- The `styles.js` and `date.jsx` files both imported functions from `lib/datetime`, which is not a relative import and so was not changed when the script was run, but then was not recognized by eslint, so I manually added a `@` at the beginning of the import.

### DIFF
--- a/config/transforms/relative_to_alias_path.js
+++ b/config/transforms/relative_to_alias_path.js
@@ -1,0 +1,50 @@
+import custom from '../webpack/webpack.config.custom';
+
+const path = require('path');
+
+const { alias } = custom.resolve;
+
+const isRelative = (filePath = '') => filePath[0] === '.';
+const filterNonRelative = p => isRelative(p.value.value);
+
+const getKeyByValue = (object, value) =>
+  Object.keys(object).find(key => object[key] === value);
+
+const renameLiteral = (j, newPath) => p => {
+  j(p).replaceWith(() => j.literal(newPath));
+};
+
+export default (fileInfo, api) => {
+  const j = api.jscodeshift;
+  const root = j(fileInfo.source);
+  const fileAbsPath = fileInfo.path;
+
+  const importDeclarations = root
+    .find(j.ImportDeclaration)
+    .find(j.Literal)
+    .filter(filterNonRelative);
+
+  const aliasImports = [];
+  importDeclarations.paths().forEach(importDeclaration => {
+    const relativeImport = importDeclaration.value.value;
+    const correspondingAlias = getKeyByValue(
+      alias,
+      path.dirname(path.resolve(path.dirname(fileAbsPath), relativeImport)),
+    );
+
+    if (correspondingAlias) {
+      aliasImports.push(
+        correspondingAlias.concat('/', path.basename(relativeImport)),
+      );
+    } else {
+      aliasImports.push(relativeImport);
+    }
+  });
+
+  importDeclarations.forEach((importDeclaration, i) => {
+    const rewriter = renameLiteral(j, aliasImports[i]);
+    rewriter(importDeclaration);
+  });
+
+  return root.toSource();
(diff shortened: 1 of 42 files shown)


config/webpack/environment.js
```diff
@@ -1,3 +1,3 @@
-const { environment } = require('@rails/webpacker')
+const { environment } = require('@rails/webpacker');
 
-module.exports = environment
+module.exports = environment;
```
geraldalewis: ~~:100:~~ :1000:

config/webpack/webpack.config.custom.js
```diff
@@ -0,0 +1,34 @@
...
+      ),
+      '@messages': path.resolve(
+        __dirname,
+        '../../app/javascript/components/widgets/messages',
+      ),
+      '@numbers': path.resolve(
+        __dirname,
+        '../../app/javascript/components/widgets/numbers',
+      ),
+      '@weather': path.resolve(
```
geraldalewis: I think these aliases are a good balance of general (`@components`) and specific (`@weather`). 👍

package.json
```diff
@@ -19,8 +19,9 @@
     "graphql-ruby-client": "^1.4.0",
     "graphql-tag": "^2.9.2",
     "hoist-non-react-statics": "^3.0.1",
-    "numeral": "^2.0.6",
+    "jscodeshift": "^0.6.4",
```
geraldalewis: Just noting that I'm personally a fan of making the addition of a lib its own commit (as it's done here). Other devs feel like adding the lib should be in the same commit as when it's used, but this way makes it _much_ easier for me to review. (Also, for other kinds of projects like `react-native`, it is _very_ useful to have libs added in their own commits).

config/transforms/relative_to_alias_path.js
```diff
@@ -0,0 +1,50 @@
...
+
+export default (fileInfo, api) => {
+  const j = api.jscodeshift;
+  const root = j(fileInfo.source);
+  const fileAbsPath = fileInfo.path;
+
+  const importDeclarations = root
+    .find(j.ImportDeclaration)
+    .find(j.Literal)
+    .filter(filterNonRelative);
```
geraldalewis: That's some nice self-documenting code 👍

package.json
cbarber: I fall into the latter camp preferring to add the library with its usage. Not requesting a change, but curious how the split is beneficial in `react-native`.

cbarber: Hrm. This should be under devDependencies instead.

package.json
```diff
@@ -54,7 +54,8 @@
   },
   "scripts": {
     "lint": "./node_modules/.bin/eslint --ext js --ext jsx app/javascript/",
-    "test": "jest"
+    "test": "jest",
+    "relative-to-alias": "jscodeshift --extensions jsx -t ./config/transforms/relative_to_alias_path.js ./ "
```
cbarber: I think this can be consolidated with the addition of the transformer script. I'm surprised you didn't have to specify js and jsx in the extensions list. Is js a default?

mariana0pachon: Yeah, js is the default

aaddaamm:
This needs a rebase 🙏 🙇

chewnoill:
looks like there are conflicts again; not sure I agree with @geraldalewis, would merge ASAP

cbarber:
Just waiting on the solar cycle PR to be merged and then I was going to focus on getting this in.

PR #97, opened 2019-06-03
Count only certain Slack message events

mariana0pachon:
- Make sure only regular messages (with no subtype) and certain subtypes are counted in the numbers. All subtypes can be found here: https://api.slack.com/events/message
- The relevant ones for now are: `file_comment`, `file_mention`, `file_share`, `me_message`, `message_replied`, `thread_broadcast` 
- Tested with rake `simulate:slack_message` where the event would be of a relevant subtype, an irrelevant subtype or would have no subtype

### DIFF
--- a/app/controllers/web_hooks/slack_controller.rb
+++ b/app/controllers/web_hooks/slack_controller.rb
@@ -1,5 +1,6 @@
 class WebHooks::SlackController < ApplicationController
   protect_from_forgery with: :null_session
+
   def create
     halt 403, "Invalid Slack verification token: #{params[:token]}" if ENV['SLACK_VERIFICATION_TOKEN'] != params[:token]
     render plain: build_create_response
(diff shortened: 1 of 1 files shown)


app/controllers/web_hooks/slack_controller.rb
```diff
@@ -18,14 +18,20 @@ def create
     render plain: build_create_response
   end
 
+  # rubocop:disable Metrics/AbcSize
```
geraldalewis: Personally totally fine with disabling this here; could anyone with more Ruby experience shed some light on ways to structure code in a more `sqrt(A*A + B*B + C*D)` way?

mariana0pachon: I had this same error in the `create` function when I was working on the slack controller like 2 weeks ago, which is why I created the separate `build_create_response` function and it was reduced. Also found that having if statements in one line helped but that's all the experience I have with it.

For this case, I could create another function `check_message_subtype` or something like that

app/controllers/web_hooks/slack_controller.rb
```diff
@@ -23,9 +23,17 @@ def build_create_response
...
     else
       "Error"
     end
   end
+
+  def check_event_subtype(params)
+    subtypes = %w[file_comment file_mention file_share me_message message_replied thread_broadcast]
+    if params[:event][:type] == 'message' &&
+       (params[:event][:subtype].nil? || subtypes.include?(params[:event][:subtype]))
+      "Add Slack message event to db #{slack_message(params)}"
```
cbarber: Ah ok. Missed this originally. I think its a little odd to trigger a database insertion from within string interpolation. What about something like:
```
when 'event_callback'
   event = params[:event]
   return "Skipping disallowed event #{event[:type]:#{event[:subtype]" unless allow_event?(event)
   result = create_slack_message(event)
   "Add Slack message event to db #{result}"
```

app/controllers/web_hooks/slack_controller.rb
```diff
@@ -10,12 +11,19 @@ def build_create_response
     when 'url_verification'
       params[:challenge]
     when 'event_callback'
-      slack_event(params, params[:event][:type])
+      return "Skipping disallowed event #{params[:type][:subtype]}" unless allow_event?(params)
```
cbarber: Looks like a typo here with `params[:type][:subtype]`. May want to just grab `params[:event]` as a variable.

PR #98, opened 2019-06-03
Weather Panel Design Update

theroyalblah:
This is the first step towards getting the redesign merged with master, with this PR focusing on the weather portions of the redesign.
The message on the side of the screen with its own icon will disappear with a future PR relating to the side panel, so the added image on the side isn't permanent, however for now it's intended behavior.

Screen Shot 2019-06-03 at 2 41 39 PM


app/javascript/components/widgets/weather/minutely-weather.jsx
```diff
@@ -1,19 +1,34 @@
...
+const IconWrapper = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+`;
+
+const MinutelyWeather = ({ weather, width, height }) => {
   const { summary } = weather.minutely;
-  return <span>{summary}</span>;
+  const { icon } = weather.minutely;
```
cbarber: Can be constructed in the line above.

app/javascript/components/widgets/weather/minutely-weather.jsx
```diff
@@ -1,19 +1,34 @@
...
+  return (
+    <IconWrapper>
+      <SkyIcon icon={icon} width={width} height={height} />
```
cbarber: I feel like there is an opportunity here to specify intent rather than height/width e.g. `<MinutelyWeather useLargeIcon={true|false} />`

```
{ useLargeIcon ? <SmallSkyIcon icon={icon} : <LargeSkyIcon icon={icon} }
```

app/javascript/components/widgets/weather/sky-icons.jsx
```diff
@@ -11,25 +11,60 @@ import CloudyIcon from '../../../../assets/images/icons/icon-1-cloud.svg';
...
-    <img src={PartlyCloudyDayIcon} alt="partly-cloudy-day" />
-  ),
-  'partly-cloudy-night': (
-    <img src={PartlyCloudyNightIcon} alt="partly-cloudy-night" />
-  ),
+export const SkyIcon = ({ icon, width, height }) => {
+  switch (icon) {
+    case 'clear-day':
+      return (
+        <img src={ClearDayIcon} alt="clear-day" width={width} height={height} />
```
cbarber: There some duplication here. Maybe a mapping of the `icon` prop to the `src` with a final return of:
```
<img src={iconMapping[icon]} alt={icon} width={width} height={height} />
```

app/javascript/components/widgets/weather/sky-icons.jsx
```diff
@@ -43,6 +78,13 @@ SkyIcon.propTypes = {
     'partly-cloudy-day',
     'partly-cloudy-night',
   ]).isRequired,
+  width: PropTypes.string,
+  height: PropTypes.string,
+};
+
+SkyIcon.defaultProps = {
+  width: '32px',
+  height: '32px',
```
cbarber: What do you think about:
```
export const SmallSkyIcon = (props) => <SkyIcon width={'32px'} height={'32px}' {...props} />;
export const LargeSkyIcon = (props) => <SkyIcon width={'72px'} height={'72px'} {...props} />;

export default SkyIcon
```
rather than specifying defaults?

app/javascript/components/widgets/weather/daily-weather.jsx
```diff
@@ -14,7 +14,7 @@ const Wrapper = styled(Row)`
   color: ${colors.grey};
   font-weight: ${weights.light};
   margin-left: ${spacing.xl};
-  margin-top: ${spacing.xxl};
+  margin-top: 65px;
```
cbarber: Is this a one off or should it become the new `spacing.xxl` value?

geraldalewis: I'm not in favor of modifying spacing constants unless I get buy-in from a designer and carefully check all instances in the app/site that would be affected. (Also, if it were simply added, it'd be need to be greater than the current largest `xxxl: '40px',` -- `xxxxl: '65px'` feels unweildy to me).

theroyalblah: I was thinking that maybe sometime we could talk with the designers about that, because a lot of the spacing given in the design does not lend itself to reusability well.

cbarber: Agreed. I meant this as more of a question to discover design intent versus asking for a change in the PR.

app/javascript/components/widgets/weather/daily-weather.jsx
```diff
@@ -7,14 +7,14 @@ import { colors, fontSizes, weights, spacing } from '../../../lib/theme';
 import { parseDay } from '../../../lib/datetime';
 import { WhiteText } from '../../typography';
 import { Row } from '../../row';
-import SkyIcon from './sky-icons';
+import SmallSkyIcon from './sky-icons';
```
cbarber: This should be `{ SmallSkyIcon }`. Right now is the default export, `SkyIcon` which requires `width` / `height`

PR #99, opened 2019-06-03
Loading Bar to Automatically Switch Tabs

theroyalblah:
This automatically switches tabs every 20 seconds, but can be stopped by pressing the spacebar. 
I also removed the tabs that were only temporary, as automatically switching to them would be pointless.

Screen Shot 2019-06-03 at 4 16 02 PM

### DIFF
--- a/app/javascript/components/tabs.jsx
+++ b/app/javascript/components/tabs.jsx
@@ -26,11 +26,58 @@ export class Tab extends React.Component {
     selected: PropTypes.bool.isRequired,
     text: PropTypes.string.isRequired,
     widgetId: PropTypes.number.isRequired,
+    totalTime: PropTypes.number.isRequired,
+    tabDown: PropTypes.func.isRequired,
   };
 
   constructor(props) {
     super(props);
-    this.state = {};
+    const { totalTime } = props;
+    this.state = {
+      timeBarProgress: totalTime,
+      maxTime: totalTime,
+    };
+    this.timeRemaining = totalTime;
+  }
+
+  componentDidMount() {
+    if (this.props.selected) {
+      this.timer();
+    }
+  }
+
+  componentDidUpdate(prevProps) {
+    const isSelected = this.props.selected;
+    if (isSelected !== prevProps.selected) {
+      if (isSelected && !this.intervalId) {
+        this.timer();
+      } else if (!isSelected) {
+        clearInterval(this.intervalId);
+        this.intervalId = null;
+        this.timeRemaining = this.state.maxTime;
+      }
+    }
+  }
+
+  componentWillUnmount() {
+    clearInterval(this.intervalId);
+    this.intervalId = undefined;
+  }
+
+  timer() {
+    const max = this.state.maxTime;
+    this.intervalId = setInterval(() => {
+      if (this.props.selected) {
+        this.timeRemaining -= 50;
+        if (this.state.timeBarProgress <= 0) {
+          this.timeRemaining = max;
+          this.props.tabDown();
+        }
+      } else {
+        this.timeRemaining = max;
+      }
+      this.setState({ timeBarProgress: this.timeRemaining });
+    }, 50);
   }
 
   render() {
--- a/app/javascript/components/tabs.jsx
+++ b/app/javascript/components/tabs.jsx
@@ -4,6 +4,8 @@ import PropTypes from 'prop-types';
 import { colors, fontSizes } from '@lib/theme';
 import { CurrentTabBar, OtherTabBar } from '@components/tab-bar';
 
+const SPACEBAR = 32;
+
 const CurrentTabText = styled.div`
   margin-top: 19px;
   margin-bottom: 19px;
@@ -19,6 +21,10 @@ const OtherTabText = styled.div`
   color: ${colors.white};
   opacity: 0.5;
   cursor: pointer;
+`;
+
+const TabButton = styled.div`
+  outline: none;
 `;
 
 export class Tab extends React.Component {
@@ -38,6 +44,7 @@ export class Tab extends React.Component {
       maxTime: totalTime,
     };
     this.timeRemaining = totalTime;
+    this.isPaused = false;
   }
 
   componentDidMount() {
@@ -55,6 +62,7 @@ export class Tab extends React.Component {
         clearInterval(this.intervalId);
         this.intervalId = null;
         this.timeRemaining = this.state.maxTime;
+        this.setPause(false);
       }
     }
   }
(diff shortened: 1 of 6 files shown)


cbarber:
I spoke with Simon in person regarding updating state frequently in a top level component and passing down a frequently changing prop through the component tree. He's going to work on switching this to passing the total time down and then having the bar either callback or fire a redux action on completion.

app/javascript/components/widget-bar.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { colors } from '../lib/theme';
+
+export const CurrentWidgetBar = ({ time }) => (
+  <svg width={568 * (time / 20000)} height="4">
```
chewnoill: what do you think about giving names to these numbers so they seem more on purpose?

app/javascript/components/widget-controller.jsx
```diff
@@ -38,6 +38,7 @@ export class WidgetController extends React.Component {
     this.state = {
       index: 0,
       time: TWENTYSECONDS,
+      isPaused: 0,
```
chewnoill: lets make this a boolean `true`/`false`

app/javascript/components/widget-controller.jsx
```diff
@@ -97,6 +104,14 @@ export class WidgetController extends React.Component {
     }));
   }
 
+  pause() {
```
chewnoill: maybe call this `togglePause`?

app/javascript/components/widget-controller.jsx
```diff
@@ -97,6 +104,14 @@ export class WidgetController extends React.Component {
...
+    if (this.state.isPaused) {
+      this.setState({ isPaused: 0 });
```
chewnoill: best practice is to use a call back

ie `this.setState(({isPaused}) => ({isPaused: !isPaused}))`

app/javascript/components/widget-bar.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { colors } from '../lib/theme';
+
+export const CurrentWidgetBar = ({ time }) => (
+  <svg width="568" height="4">
+    <rect width={568 * (time / 20000)} height="4" fill={colors.white} />
```
geraldalewis: If we replace the hardcoded `20000` here by importing the `const`, we can ensure any changes to the `time` value are reflected everywhere.

app/javascript/components/widget-bar.jsx
```diff
@@ -0,0 +1,20 @@
...
+    <rect width="568" height="1" fill={colors.white} />
+  </svg>
+);
+
+export const OtherWidgetBar = () => (
```
geraldalewis: Is there a more specific word we can use than `Other`? Or a pair of adjectives we can use instead (Active/Inactive)?

app/javascript/components/widget-controller.jsx
```diff
@@ -1,120 +1,108 @@
...
-
-const Icon = styled.button`
-  height: 21px;
-  width: 16px;
-  border-style: none;
-  background-color: ${colors.black};
-`;
-
-const iconId = 'lock-icon';
+const TWENTYSECONDS = 20000;
```
geraldalewis: Love using a constant here for this value, but the ref should describe what the value _does_ rather than what it _is_. E.g., `const SWITCH_INTERVAL = ...` or something

app/javascript/components/widget-controller.jsx
```diff
@@ -1,120 +1,108 @@
...
     const upKey = 38;
     const downKey = 40;
+    const spacebar = 32;
     if (keyCode === downKey) {
       this.moveDown();
     }
     if (keyCode === upKey) {
       this.moveUp();
     }
+    if (keyCode === spacebar) {
```
geraldalewis: nice 👍

app/javascript/components/widget-controller.jsx
```diff
@@ -1,120 +1,108 @@
...
+    if (targetId.startsWith('widget_')) {
       const widgetId = parseInt(targetId.replace('widget_', ''), 10);
       this.switchToPage(widgetId);
     }
   };
 
   startCarousel() {
-    setInterval(() => {
-      if (!this.state.isLocked) {
+    this.intervalID = setInterval(() => {
```
geraldalewis: Before creating a new interval, let's remove the old interval:

```javascript
  if (this.intervalId) clearInterval(this.intervalId);
  this.intervalID = setInterval(() => { ...
```

app/javascript/components/widget-controller.jsx
```diff
@@ -1,120 +1,108 @@
...
   moveUp() {
+    this.setState({ time: TWENTYSECONDS });
     this.setState(({ index }) => ({
       index: mathMod(index - 1, widgets.length),
     }));
   }
 
   moveDown() {
+    this.setState({ time: TWENTYSECONDS });
     this.setState(({ index }) => ({
```
geraldalewis: Let's merge these two setStates together `this.setState(({ index, time: TWENTYSECONDS }) => ({...`

app/javascript/components/widget-bar.jsx
theroyalblah: I like Active and Inactive

aaddaamm:
@theroyalblah could you give this a rebase off latest master? 🙏

app/javascript/components/tabs.jsx
```diff
@@ -107,7 +127,7 @@ export class Tab extends React.Component {
     }
     return (
       // eslint-disable-next-line
-      <div>
+      <div onKeyDown={this.spaceBarPress} tabIndex="0">
```
thedelchop: So `spaceBarPress` is a tough name for that function, since you reject a whole bunch of calls if the button pressed wasn't the spacebar.  I think this would be much more clear if you inlined this code inside of `onKeyDown`, since then its less misleading.

`onKeyDown={({keyCode}) => if (keyCode === 32) this.togglePause()}`

theroyalblah: @thedelchop what if we had `if (keyCode === spacebar)`
I don't think it's inherently obvious keyCode 32 is the spacebar, so maybe that would make it clearer?

app/javascript/components/tabs.jsx
```diff
@@ -20,35 +23,116 @@ const OtherWidgetText = styled.div`
...
+      );
+    }
     return (
-      <span>
-        <CurrentWidgetText id={`widget_${props.widgetId}`}>
-          {props.text}
-        </CurrentWidgetText>
-        <TabBar />
-      </span>
+      // eslint-disable-next-line
```
geraldalewis: I'd just move that `onKeyDown` inline fn expression into a class method. It's a little more readable to me when it's not inline in the render method, and it's arguably a good practice (since creating a new method here will invalidate this part of the render tree in React's eyes).

```render
class Tabs {
  handleKeyDown = ({ keyCode }) => ...;

  render() {
    <TabButton onKeyDown={} />
  }
```

PR #100, opened 2019-06-04
fix: remove trailing 0 from time

mariana0pachon:
Change `hh` to `h` for time, which was accidentally switched when implementing everything with `date-fns`


joealba:
@mariana0pachon I think your change is impacting the _leading_ zero instead of _trailing_ zero, right?  Can you please update the commit and PR descriptions?

PR #102, opened 2019-06-04
Add widgets table and model

mariana0pachon:
- Backend to store Widgets that will be rendered in the frontend
- Stores: `name` which matches frontend naming, `enabled` flag, `duration_seconds` of the display, `order` of appearance on the side bar
- Scope: `all_in_order` sorted by `order`, and `enabled` which is also sorted by `order`
- Seeds: `Guests`, `Weather`, `Twitter`, `Numbers` in that order

### DIFF
--- a/app/models/widget.rb
+++ b/app/models/widget.rb
@@ -1,0 +1,4 @@
+class Widget < ApplicationRecord
+  validates :name, presence: true, uniqueness: true
+  validates :position, presence: true, uniqueness: true
+end
--- a/app/models/widget.rb
+++ b/app/models/widget.rb
@@ -1,4 +1,7 @@
 class Widget < ApplicationRecord
   validates :name, presence: true, uniqueness: true
   validates :position, presence: true, uniqueness: true
+
+  default_scope { order(position: :asc) }
+  scope :enabled, -> { where(enabled: true) }
 end
--- a/db/migrate/20190604172236_create_widgets.rb
+++ b/db/migrate/20190604172236_create_widgets.rb
@@ -1,0 +1,12 @@
+class CreateWidgets < ActiveRecord::Migration[5.2]
+  def change
+    create_table :widgets do |t|
+      t.string :name, null: false
+      t.boolean :enabled, null: false
+      t.integer :duration_seconds, null: false
+      t.integer :position, null: false
+
+      t.timestamps
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,15 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_05_31_061410) do
-
-  create_table "daily_event_summaries", force: :cascade do |t|
-    t.string "source", null: false
-    t.date "day", null: false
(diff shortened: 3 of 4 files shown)


db/schema.rb
```diff
@@ -10,7 +10,19 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_05_31_061410) do
+ActiveRecord::Schema.define(version: 2019_06_04_172236) do
+
+  create_table "announcements", force: :cascade do |t|
```
cbarber: That's weird. Did this not save from the announcements PR? Or does this need a rebase onto master?

app/models/widget.rb
```diff
@@ -1,4 +1,7 @@
 class Widget < ApplicationRecord
   validates :name, presence: true, uniqueness: true
   validates :order, presence: true, uniqueness: true
+
+  scope :all_in_order, -> { order(order: :asc) }
+  scope :enabled, -> { where(enabled: true).order(order: :asc) }
```
cbarber: I don't think it makes sense to chain order into the enabled scope since we'd want it to be explicit. May instead want to use `default_scope` to define the sort order for all queries.

geraldalewis: @cbarber can you elucidate more what you mean here? I'm not following entirely.

db/schema.rb
mariana0pachon: Doesn't it make sense that it's yesterday's date since that's when I last ran `rake db:migration`?

db/migrate/20190604172236_create_widgets.rb
```diff
@@ -0,0 +1,12 @@
+class CreateWidgets < ActiveRecord::Migration[5.2]
+  def change
+    create_table :widgets do |t|
+      t.string :name, null: false
+      t.boolean :enabled, null: false
+      t.integer :duration_seconds, null: false
+      t.integer :order, null: false
```
thedelchop: This is a bit of a nit, _but_ usually, at least in my experience with Rails that people call this field `position`, not `order`.  I think its better as well, because `order` is a description of a list, but `position` is the placement of a _single_ object in the list.  I'd rename.

app/models/widget.rb
cbarber: There's two parts:
* scopes should be very explicit since they are chainable e.g. you wouldn't write `scope :enabled, -> { where(enabled: true).order(order: :asc) }`. Instead you would write `scope :enabled, -> { where(enabled: true) }` and use `Widget.enabled.all_in_order.all`.
* we likely always want our widgets to be selected in order. in that case, we can omit defining `scope :all_in_order ...` and simply define `default_scope => { order(order: :asc) }`. Now *every* query for widgets will be ordered.

db/seeds.rb
```diff
@@ -25,3 +25,31 @@
   city_name: 'NYC',
   time_zone: 'America/New_York'
 )
+
+Widget.create(
+  name: "Guests",
+  enabled: false,
+  duration_seconds: 20,
+  order: 0
```
thedelchop: @mariana0pachon I think you missed these in the renaming from `order` to `position`.

PR #104, opened 2019-06-05
chore: upgrade dependencies for node.js v12

sracca:
-updated package.json and yarn.lock using 'yarn upgrade'
- is compatible with node.js v10


joealba:
@sracca Have you tried `yarn upgrade-interactive` yet?  It's a really handy way to see outdated dependencies and upgrade just the ones you select.  YOLO upgrades like this may be fine for an internal project, but it's good to pick up the habit of reviewing changelogs on your upgraded packages and wrapping up upgrades into smaller commits that are easier to review.

Not a blocker at all, just wanted to bring it up when I saw the very large change to `yarn.lock` -- plus it was something I was just doing earlier today.  :)

sracca:
@joealba No I didn't realize that was a command. I'll definitely use that in the future!

aaddaamm:
Another way that i like to do upgrades is to just run `yarn outdated` and upgrade each package individually. i've found that with anytime i've run all the upgrades, things start breaking and it's difficult to discern what specific upgrade caused the issue. you then get nice commits talking about each upgrade and what the previous version is. It's slower for sure but usually has less surprises.

aaddaamm:
@sracca Sorry to be a pain but would you mind breaking the upgrades into individual commits? That way we know exactly what is getting upgraded in this PR. 🙏 🙇 🥇

sracca:
@aaddaamm Of course! Thanks so much!

aaddaamm:
also, if you are feeling brave, we can upgrade some of the dependencies to the latest versions. 😬 not a requirement though, that could create a lot more work. anything that would be a hellish upgrade we can make tickets for and do it later.

sracca:
@aaddaamm sorry for the wait but I fixed it based on what you said. I updated almost all of the dependencies need to the latest except for ramda and styled-components, which would need some changes to the code, which I thought would be better as separate tickets. Thanks!

PR #105, opened 2019-06-06
chore: upgrade to react v16.8.0

sracca:
- includes an addition to eslint to include rules on react-hooks


aaddaamm:
checked it out locally and fired it up, seems to run A-OK. interested to see what hooks will be used for.

PR #107, opened 2019-06-06
Side Panel Updates

theroyalblah:
This includes general formatting and spacing edits of the contents in the side panel, including the time date and widget bar. The biggest change being to the time, as the secondary times were removed. As a result of no longer needing those times, some code was removed and times's confusing name was renamed to time-hero, especially because now it only fetches one time.

Screen Shot 2019-06-13 at 11 40 01 AM


app/javascript/components/time.jsx
```diff
@@ -1,54 +1,37 @@
...
+const HourMin = styled.div`
+  font-size: ${fontSizes.xlarge};
+  font-family: ${fonts.extended};
+  font-weight: ${weights.regular};
+  position: relative;
+  top: 8px;
+`;
 
-const AMPM = styled.div``;
+const AMPM = styled.div`
```
cbarber: Nit: could be named `Meridiem`. Might just be more confusing but I initially thought `AMPM` was an acronym.

PR #108, opened 2019-06-06
Fix Slack announcement bugs

mariana0pachon:
- Change simulation templates to match a complete incoming slack event and not just the event parameter.
- Parse incoming message for guests correctly. Was not realizing that when the event is an `app_mention` a first word corresponding to the bot id is added to the text parameter.
- If an announcement that was today but in the past was added it would be appended to the list, so fixed that by adding another datetime function.

### DIFF
--- a/app/controllers/web_hooks/slack_controller.rb
+++ b/app/controllers/web_hooks/slack_controller.rb
@@ -13,7 +13,7 @@ class WebHooks::SlackController < ApplicationController
     when 'event_callback'
       event = params[:event]
       return "Skipping disallowed event #{event[:subtype]}" unless allow_event?(event)
-      result = slack_event(params, event[:type])
+      result = slack_event(event, event[:type])
       "Add Slack message event to db #{result}"
     else
       "Error"
@@ -26,12 +26,12 @@ class WebHooks::SlackController < ApplicationController
       (event[:subtype].nil? || subtypes.include?(event[:subtype]))
   end
 
-  protected def slack_event(params, type)
+  protected def slack_event(event, type)
     case type
     when 'message'
-      "Add Slack message to db #{publish_event(Event.slack_messages.with_external_id(params[:event_id]))}"
+      "Add Slack message to db #{publish_event(Event.slack_messages.with_external_id(event[:event_ts]))}"
     when 'app_mention'
-      "Add Slack announcement to db #{publish_announcement(params)}"
+      "Add Slack announcement to db #{publish_announcement(event)}"
     else 'Error'
     end
   end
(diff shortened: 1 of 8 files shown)


app/javascript/components/widgets/guests/tab.jsx
```diff
@@ -121,7 +121,13 @@ const Guests = () => (
                     prev,
                   );
                 }
-                return prev;
+
+                const isFutureToday = item => isInFutureToday(item.publishOn);
```
cbarber: It seems like we could still have old announcements if we add a new announcement that is in the future. The solution might be chaining it all together: append new event, filter, sort.

app/controllers/web_hooks/slack_controller.rb
```diff
@@ -44,19 +44,18 @@ def allow_event?(event)
...
     Helios2Schema.subscriptions.trigger("announcementPublished", {}, announcement)
   end
 
-  private def parse_announcement(event_callback)
-    # Ex: "guests: Welcome to Roy and Amanda from Under Armour on May 23rd 2019 at 11:00 am in Providence"
-    type, content = event_callback[:event][:text].split(/: /)
-
+  private def parse_announcement(event)
+    # Ex: "<@U0LAN0Z89> guests= Welcome to Roy and Amanda from Under Armour on May 23rd 2019 at 11:00 am in Providence"
+    _bot_id, type, content = event[:text].split(/>|: /).map(&:strip)
```
cbarber: I'm a little worried this may match other parts of the string. You can use a more exact regex to avoid that:
```
if match = event[:text].match(/^<\S+> (\S+) (.*)$/)
  type, content = match.captures
  # stuffs
else
  # potential help message
end
```

app/lib/simulate/slack-announcement.json.erb
```diff
@@ -1,19 +1,36 @@
 {
+  "token": "Sim-123456",
```
cbarber: Hrm. Should this be in the same commit as the slack simulate changes?

mariana0pachon: You mean squash this commit with the next one where I make changes to the `slack_controller` ?

app/javascript/components/widgets/guests/tab.jsx
mariana0pachon: So I tested this and if a future announcement added and there is one in the list that is in the past then the filter will take care of it. The problem I see is that a past announcement will remain in the list until the next poll (which happens every hour) or until the next published announcement. So maybe the `pollInterval` could be shorter or we would be ok with having an event that happened an hour ago until the next poll.

mariana0pachon:
The `SLACK_WEBHOOK_URL` can be found here https://mojotech.slack.com/services/BK1HVGXEW

app/lib/simulate/slack-announcement.json.erb
cbarber: I think so. It seems like the bug fix and slack message format should be part of the same commit.

PR #109, opened 2019-06-07
Improved Falling Blocks

theroyalblah:
The general idea here was to improve performance and fix the problem where the blocks would crush one another. 

On a screen transition, all of the previous list of bodies are removed, which improves performance. Additionally, the blocks that had little movement get locked in place which helps with a couple things.
- Blocks locked in place while sleeping make no calculations until they wake.
- This prevents the blocks from crushing one another

The old version would drop around 500 or so and then collapse and drop to ~5 fps. This updated version got ~57fps at 1000 blocks, and to 33fps at around 3000 blocks or so.

Most of the performance problems at this stage are just from trying to draw so many objects, the physics aren't the issue anymore. Specifically render.js's drawImage is taking more than half of the cpu per frame.

Screen Shot 2019-06-13 at 11 19 43 AM


app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -40,7 +40,15 @@ class Scene extends React.Component {
       Engine.merge(this.engine, { world: loadedWorld });
     }
     this.addBlocks();
-    this.freezeBlocks();
+    let i;
+    const allBodies = Composite.allBodies(this.engine.world);
+    if (allBodies.length > 1500) {
+      for (i = 3; i < allBodies.length; i += 1) {
```
chewnoill: I assume we start at 3 to skip the walls? lets try and find a better way to identify and filter those out. I imagine it would look something like `allBodies.filter(isWall).map(...)`.

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -195,6 +202,17 @@ class Scene extends React.Component {
     });
   }
 
+  freezeBlocks() {
+    const allBodies = Composite.allBodies(this.engine.world);
+    let i = 0;
+    for (i = 0; i < allBodies.length; i += 1) {
+      if (allBodies[i].motion < 0.4 && allBodies[i].position.y > 10) {
+        Body.setStatic(allBodies[i], true);
+        Body.set(allBodies[i], 'isSleeping', true);
```
cbarber: What is the difference between `isSleeping` and `static`?

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -37,14 +37,21 @@ class Scene extends React.Component {
...
+    this.freezeBlocks();
   }
 
   componentWillUnmount() {
     clearTimeout(this.timer);
     this.save();
+    if (this.engine.world) {
+      const allBodies = Composite.allBodies(this.engine.world);
+      let i = 0;
+      for (i = 0; i < allBodies.length; i += 1) {
```
cbarber: Likely identical to `Composite.clear`

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -27,7 +27,8 @@ class Scene extends React.Component {
       ...currentState,
     };
     this.scene = React.createRef();
-    this.timer = null;
+    this.addBlockTimer = null;
+    this.freezeTimer = null;
```
cbarber: Why iterate on the collection of bodies outside of the render loop? It seems to me a likely culprit could be the matter js builin render method. Excerpt from https://github.com/liabru/matter-js/wiki/Rendering

> There is an included debug renderer called Matter.Render. This module is an optional canvas based renderer for visualising instances of Matter.Engine. It is mostly intended for development and debugging purposes, but may also be suitable starting point for simple games.

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -168,13 +168,13 @@ class Scene extends React.Component {
...
-      Bodies.circle(this.state.width / 2 + (Math.random() - 0.5), -50, 17, {
+      Bodies.circle(this.state.width / 2 + (Math.random() - 0.5), -50, 6, {
         restitution: 0.5,
         render: {
           sprite: {
             texture: blockTypes[block],
-            xScale: 0.07,
-            yScale: 0.07,
+            xScale: 0.025,
+            yScale: 0.025,
```
cbarber: This commit is doing two things `and`

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -219,3 +227,45 @@ export default compose(
   withLocalState,
   withLocalMutation,
 )(Scene);
+
+// eslint-disable-next-line no-underscore-dangle
+Sleeping._motionSleepThreshold = 0.5;
+
+// eslint-disable-next-line func-names
+Sleeping.update = function(bodies, timeScale) {
```
geraldalewis: Not seeing the need for copying over this code -- doesn't look like anything changes except our linting rules?

If that's the case, I think `Sleeping._motionSleepThreshold = 0.5;` should be enough.

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -106,7 +106,7 @@ class Scene extends React.Component {
         wallWidth / 2,
         this.state.height / 2,
         wallWidth,
-        this.state.height,
+        this.state.height + 50,
```
geraldalewis: How'd you land on `50` here?

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -75,8 +75,8 @@ class Scene extends React.Component {
       element: this.scene.current,
       engine: this.engine,
       options: {
-        width: window.innerWidth,
-        height: window.innerHeight,
+        width: this.state.width,
+        height: this.state.height,
```
geraldalewis: I think these `width` and `height` changes can be their own commit.

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -166,7 +166,7 @@ class Scene extends React.Component {
     // directly on top of each other.
     World.add(
       this.engine.world,
-      Bodies.polygon(this.state.width * Math.random(), -50, 8, 8, {
+      Bodies.polygon((this.state.width - 50) * Math.random() + 50, -10, 8, 8, {
```
chewnoill: hardcoding this might cause a problem if the width is < 50

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -166,7 +166,7 @@ class Scene extends React.Component {
     // directly on top of each other.
     World.add(
       this.engine.world,
-      Bodies.circle(this.state.width / 2 + (Math.random() - 0.5), -50, 8, {
+      Bodies.polygon(this.state.width * Math.random(), -50, 8, 8, {
```
chewnoill: looks like more then just a shape change here. I imagine some parts of this belong in 387bd0ec52add4fdc36587810dca1446887bef7d

app/javascript/components/widgets/numbers/falling-blocks.jsx
chewnoill: lets also update the comment on L165 so its not a lie

app/javascript/components/widgets/numbers/falling-blocks.jsx
theroyalblah: Sorry, putting that function there was a bit unclear on my part, but it needs to be overwritten to enforce setting the bodies to static instead of putting them to sleep.
```suggestion
        if (body.sleepCounter >= body.sleepThreshold) {
          Body.setStatic(body, true);
        }
```
versus the original ... 
```suggestion
                if (body.sleepCounter >= body.sleepThreshold)
                    Sleeping.set(body, true);
```

cbarber: In that case, lets name the function for what it does differentlyand then assign it to `Sleeping.update` which would also address the func-names eslint violation.

theroyalblah: How does this sound?
`Sleeping.update = function freezeOnSleep(bodies, timeScale) {`

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -85,7 +85,7 @@ class Scene extends React.Component {
     });
 
     const wallWidth = 50;
-
+    this.engine.world.gravity.y = 0.5;
```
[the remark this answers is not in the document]
theroyalblah: I don't think so, or at least it isn't listed in the options field in Engine.js
`        var defaults = {
            positionIterations: 6,
            velocityIterations: 4,
            constraintIterations: 2,
            enableSleeping: false,
            events: [],
            plugin: {},
            timing: {
                timestamp: 0,
                timeScale: 1
            },
            broadphase: {
                controller: Grid
            }
        };`
https://github.com/liabru/matter-js/blob/master/src/core/Engine.js

app/javascript/components/widgets/numbers/falling-blocks.jsx
theroyalblah: I wanted the walls to have a height greater than the screen height, as the blocks are created at a height above the screen as well. They need to be at least a tad higher up than their creation height, as two can be created at the same-ish point and shove each other in a random direction. However regardless of what number I chose I felt like my height choice was a bit arbitrary so I just went with 50. I'm definitely open to suggestions on that though.

app/javascript/components/widgets/numbers/falling-blocks.jsx
geraldalewis: maybe `staticOnSleep`? (unless freeze and static are used interchangeably in matterjs code/docs). Also, we're already patching `Sleeping.update` -- could we do:

```javascript
Sleeping._motionSleepThreshold = 0.5;
Sleeping.set = (param1, param2)=> Body.setStatic(body, true);
```
(that closure above^ is _probably_ necessary and I can go through it if it's not clear why)

app/javascript/components/widgets/numbers/falling-blocks.jsx
geraldalewis: Cool -- a comment there could help a future dev feel confident about changing that value in the future if need-be. You might also consider something like:

```javascript
// Padding applied to the height of the walls to prevent blocks created contemporaneously from colliding.
const padding = 50;
```

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -1,5 +1,36 @@
...
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+ */
```
cbarber: Can we make a separate file for this license and the relevant function to closely relate them and reduce noise here?

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -189,8 +189,9 @@ class Scene extends React.Component {
     // directly on top of each other.
     World.add(
       this.engine.world,
-      Bodies.circle(this.state.width / 2 + (Math.random() - 0.5), -50, 17, {
-        restitution: 0.7,
+      Bodies.circle(this.state.width / 2 + (Math.random() - 0.5), -50, 8, {
+        restitution: 0.25,
+        friction: 0.8,
```
cbarber: `change block properties` is kind of vague. Could you indicate intent or provide some more context in the commit body following the first line?

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -195,8 +195,6 @@ class Scene extends React.Component {
         render: {
           sprite: {
             texture: blockTypes[block],
-            xScale: 0.07,
-            yScale: 0.07,
```
cbarber: Commit message should suggest `scaling` or `resizing` the block images.

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -185,19 +185,25 @@ class Scene extends React.Component {
...
     World.add(
       this.engine.world,
-      Bodies.polygon(this.state.width / 2 + (Math.random() - 0.5), -50, 8, 8, {
-        restitution: 0.25,
-        friction: 0.8,
-        render: {
-          sprite: {
-            texture: blockTypes[block],
+      Bodies.polygon(
+        (this.state.width - wallWidth) * Math.random() + wallWidth,
```
cbarber: This is a good example where naming something can improve quality more than a comment:
```
const randomPositionBetweenWalls = (this.state.width - wallWidth) * Math.random() + wallWidth;
....
Bodies.polygon(
  randomPositionBetweenWalls,
```

The `so that they don't fall directly on top of each other` could stay or move into the commits body beneath the first line (better describing the background of the change)

app/javascript/components/widgets/numbers/falling-blocks.jsx
cbarber: Why are width and height in state?

app/javascript/components/widgets/numbers/sleeping-blocks.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+ */
+import { Sleeping, Body } from 'matter-js';
```
chewnoill: are we using `Body` somewhere?

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Engine, Render, World, Bodies, Sleeping, Composite } from 'matter-js';
```
chewnoill: probably this should be squashed down into 67813e392d22077148d29b89ba87214e1192dae9, since that was the last reference to `Sleeping`

app/javascript/components/widgets/numbers/falling-blocks.jsx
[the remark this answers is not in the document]
chewnoill: https://github.com/liabru/matter-js/blob/master/src/core/Engine.js#L86? Its not super important though, this should be fine the way it is

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -13,6 +13,10 @@ import '@numbers/sleeping-blocks';
 import { withLocalMutation, withLocalState } from '@numbers/ducks';
 
 const blockTypes = { githubPull, githubCommit, slackMessage };
+const wallWidth = 50;
+
+// Padding applied to the height of the walls to prevent blocks created contemporaneously from colliding.
```
chewnoill: not sure I understand this comment. Why are blocks colliding with walls and why was it a problem?

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -159,8 +159,9 @@ class Scene extends React.Component {
     // directly on top of each other.
     World.add(
       this.engine.world,
-      Bodies.circle(this.state.width / 2 + (Math.random() - 0.5), -50, 17, {
-        restitution: 0.7,
+      Bodies.circle(this.state.width / 2 + (Math.random() - 0.5), -50, 8, {
```
chewnoill: might be a good idea to squash c56b48b15934e0ec8d1ecf0cc501e36a6e63190e into this commit, as this changes the size of the circle but doesn't touch the texture.

app/javascript/components/widgets/numbers/falling-blocks.jsx
theroyalblah: The theory was there was this edge case where blocks could be created close to inside one another and as a result push each other unexpectedly. In the event that this took place on the edge of the world, they could be pushed over the wall and out of bounds. With a taller wall we can try prevent them from escaping without some expensive loop making sure that none have left the screen.

app/javascript/components/widgets/numbers/sleeping-blocks.jsx
theroyalblah: We use Body on line 60 to set static instead of sleep.
`Body.setStatic(body, true);`

PR #110, opened 2019-06-11
Update graphql schema for widget querying

mariana0pachon:
- Created a `Widget` graphql interface to be used in the frontend to do a single query to get all `Weather`, `Guests`, `Numbers` and `Twitter` information for a given location
- Example query and result: 
Screen Shot 2019-06-11 at 4 22 41 PM
- Eventually all existing queries in the frontend could be replaced with a single query like this, and some graphql query type definitions might become unnecessary

### DIFF
--- a/app/graphql/helios2_schema.rb
+++ b/app/graphql/helios2_schema.rb
@@ -4,4 +4,6 @@ class Helios2Schema < GraphQL::Schema
   mutation(Types::MutationType)
   query(Types::QueryType)
   subscription(Types::SubscriptionType)
+
+  orphan_types [Types::GuestsWidget, Types::WeatherWidget, Types::TwitterWidget, Types::NumbersWidget]
 end
--- a/app/javascript/schema.json
+++ b/app/javascript/schema.json
@@ -353,6 +353,24 @@
               "deprecationReason": null
             },
             {
+              "name": "widgets",
+              "description": null,
+              "args": [
+
+              ],
+              "type": {
+                "kind": "NON_NULL",
+                "name": null,
+                "ofType": {
+                  "kind": "OBJECT",
+                  "name": "WidgetCollection",
+                  "ofType": null
+                }
+              },
+              "isDeprecated": false,
+              "deprecationReason": null
+            },
+            {
               "name": "wifiName",
               "description": null,
               "args": [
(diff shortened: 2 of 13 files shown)


db/migrate/20190604172236_create_widgets.rb
```diff
@@ -5,6 +5,8 @@ def change
       t.boolean :enabled, null: false
       t.integer :duration_seconds, null: false
       t.integer :position, null: false
+      t.string :widget_id, null: false, unique: true
```
cbarber: Hrm. What is this used for?

app/javascript/schema.json
```diff
@@ -36,51 +36,6 @@
           "name": "Query",
           "description": null,
           "fields": [
-            {
-              "name": "announcements",
```
cbarber: This is a side effect of changing the schema so I feel like it should be part of the schema change commit.

app/models/widget.rb
```diff
@@ -6,4 +6,16 @@ class Widget < ApplicationRecord
 
   default_scope { order(position: :asc) }
   scope :enabled, -> { where(enabled: true) }
+  scope :next, ->(current_id) {
+    begin
+      current_position = enabled.index(enabled.find_by(widget_id: current_id))
```
cbarber: Ah ok. What is the benefit of using widget_id over id?

The general rule of thumb is you want a scope to return a scope so they are chainable. I'd expect `Widget.next` to always resolve to a model instance so `def self.next(current_id)` would be preferred as a hint its not chainable.

I this is a good case for adding a spec.

It also looks like this commit may be out of order from the schema change. There are a couple of "fix github sort" one liners floating around. The simplest would be to rebase with each commit marked for reword and making no wording changes.

app/models/location.rb
```diff
@@ -5,6 +5,7 @@ class Location < ApplicationRecord
   validates :time_zone, presence: true
 
   has_many :announcements, inverse_of: :location, dependent: :destroy
+  has_many :widgets, inverse_of: :location, dependent: :destroy
```
chewnoill: not sure I understand why locations own widgets, I imagined the location the client wanted would only have to live on the client in a param

app/graphql/helios2_schema.rb
```diff
@@ -4,4 +4,6 @@ class Helios2Schema < GraphQL::Schema
   mutation(Types::MutationType)
   query(Types::QueryType)
   subscription(Types::SubscriptionType)
+
+  orphan_types [Types::GuestsWidget, Types::WeatherWidget, Types::TwitterWidget, Types::NumbersWidget]
```
chewnoill: this is interesting. I didn't know orphan types were a thing. Is this just a rails idea? I guess because ruby doesn't do any type inference?

app/models/location.rb
mariana0pachon: I think because different locations have different data for the weather and for the announcements specifically, so when we do a single query for every widget we want to get only the relevant information for the location running the app. The primary location is in the env variable and then the query would just be something like the screenshot: `{primaryLocation{widgets{...}}` and then no other location would need to be specified. Eventually, we could add another query that would be something like: `{locations(city_name: "Boulder"){widgets{ ... }}}` if the location wanted is not the primary.

app/graphql/helios2_schema.rb
mariana0pachon: Not 100% sure, @cbarber probably knows more, but I think it is only a [gql-ruby](https://graphql-ruby.org/type_definitions/interfaces#orphan-types) interface idea: 
> If you add an object type which implements an interface, but that object type doesn't properly appear in your schema, then you need to add that object to the interfaces's orphan_types

Maybe a consequence of not having type inference?

aaddaamm:
@mariana0pachon would you be able to rebase this plz? 🙏

app/models/location.rb
chewnoill: I guess what I assumed was that the widgets would be the same for all locations. The content would change based on a query param that would be passed into the weather widget.

db/migrate/20190604172236_create_widgets.rb
```diff
@@ -5,6 +5,7 @@ def change
       t.boolean :enabled, null: false
       t.integer :duration_seconds, null: false
       t.integer :position, null: false
+      t.references :location
```
chewnoill: probably we should not be changing an existing migration like this, better to create a new migration instead

app/models/location.rb
chewnoill: the way you have it should be fine though.

cbarber:
@chewnoill The widgets will very likely be different based on location. Future plans are:
* use the queried location to replace `primaryLocation` so we can do things in the url like `/Providence` or `/Boulder` to allow the Boulder helios to run off of heroku
* widgets may be configured differently between the offices (announcements and visitors)
* by the numbers may end up with office specific events.
* a webcam widget would need it

PR #112, opened 2019-06-13
feature: add count arg to simulate PR/slack

cbarber:
Adds `rake simulate:pull_request[COUNT]` and `rake simulate:slack_message[COUNT]`

### DIFF
--- a/lib/tasks/simulate.rake
+++ b/lib/tasks/simulate.rake
@@ -2,11 +2,13 @@ namespace :simulate do
   desc "Simulates opening a new pull request on github"
   task pull_request: :environment do
     post_github_event('pull_request', render_new_pull_request_json)
+    disable_github_authentication
   end
 
   desc "Simulates a new commit on github"
   task :commit, [:count] => :environment do |_t, args|
     args.with_defaults(count: 1)
+    disable_github_authentication
     post_github_event('push', render_new_push_json(args[:count].to_i))
   end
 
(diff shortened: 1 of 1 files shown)


lib/tasks/simulate.rake
```diff
@@ -1,18 +1,22 @@
...
+  task :pull_request, [:count] => :environment do |_t, args|
+    args.with_defaults(count: 1)
+    disable_github_authentication
+    args[:count].to_i.times { post_github_event('pull_request', render_new_pull_request_json) }
   end
 
   desc "Simulates a new commit on github"
   task :commit, [:count] => :environment do |_t, args|
     args.with_defaults(count: 1)
+    disable_github_authentication
```
geraldalewis: Would this get its own commit?

cbarber: Hrm. Maybe. The issue was calling `disable_github_authentication` multiple times was throwing any error since it can't remove a controller lifecycle callback more than once. That is only a bug if you want to `post_github_event` more than once.

cbarber: Yeah, ok. I'm sold after typing that all out.

cbarber:
@geraldalewis updated.

lib/tasks/simulate.rake
geraldalewis: Interesting -- I'dve been happy either way so 💯 🏅

PR #115, opened 2019-06-14
fix: remove bad Column styling 

theroyalblah:
Message used to rely on margins to center itself, which caused problems, especially
on the guest tab. Now it uses the screen height to center itself.

### DIFF
--- a/app/javascript/components/widgets/messages/message.jsx
+++ b/app/javascript/components/widgets/messages/message.jsx
@@ -19,9 +19,6 @@ const MessageDisplay = styled(GreySubText)`
 const Column = styled.div`
   display: flex;
   flex-direction: column;
-  margin-top: 100px
-  margin-bottom: 500px
-  height: 100%;
 `;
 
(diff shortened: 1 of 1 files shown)


app/javascript/components/widgets/messages/message.jsx
```diff
@@ -5,9 +5,14 @@ import { GreySubText } from '@components/typography';
 import { fontSizes, leftPanelWidth } from '@lib/theme';
 import sunMoon from '@images/sunmoon.png';
 
-export const LoadingMessage = () => <Message message="Loading..." />;
+export const LoadingMessage = () => (
+  <Message message="Loading..." height={(window.innerHeight - 100) / 2} />
```
cbarber: Although this likely works, I suspect its solving the symptom rather than the disease. I'd rather resolve the css issue that causes are containing width to be wrong than add an inline style

cbarber:
Is this intended to address the scrollbar issues? I'm still seeing them in chrome

PR #117, opened 2019-06-17
fix: address issue with solarcycle proxy class not working with graphql and limit solarycycles to after yesterday

cbarber:
address issue with solarcycle proxy class not working with graphql and limit solarycycles to after yesterday

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -13,7 +13,7 @@ class Location < ApplicationRecord
 
   def weather
     response = Clients::DarkskyClient.forecast(self).dup
-    response['solarcycles'] = solarcycles
+    response['solarcycles'] = solarcycles.all.to_a
     response.freeze
     response
   end
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -13,7 +13,7 @@ class Location < ApplicationRecord
 
   def weather
     response = Clients::DarkskyClient.forecast(self).dup
-    response['solarcycles'] = solarcycles.all.to_a
+    response['solarcycles'] = solarcycles.after_beginning_of_yesterday.to_a
     response.freeze
     response
   end
--- a/app/models/solarcycle.rb
+++ b/app/models/solarcycle.rb
@@ -1,4 +1,6 @@
 class Solarcycle < ApplicationRecord
   self.inheritance_column = :_type_disabled
   belongs_to :location
+
+  scope :after_beginning_of_yesterday, -> { where('time > ?', 1.day.ago.beginning_of_day) }
 end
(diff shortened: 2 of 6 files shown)


app/models/solarcycle.rb
```diff
@@ -1,4 +1,6 @@
 class Solarcycle < ApplicationRecord
   self.inheritance_column = :_type_disabled
   belongs_to :location
+
+  scope :after_beginning_of_yesterday, -> { where('created_at < ?', 1.day.ago.beginning_of_day) }
```
sracca: Just for clarification, is this to limit checking solarcycles after yesterday, or to only getting the solarcycles from the day before to limit the amount in the db?

cbarber: The latter. Otherwise we'll always be sending all solarcycles in the past and the network response will become slow.

mariana0pachon: Shouldn't this be `where('created_at > ?', 1.day.ago.beginning_of_day)` ?

app/models/location.rb
```diff
@@ -13,7 +13,7 @@ def self.primary(city_name = ENV['PRIMARY_CITY_NAME'])
 
   def weather
     response = Clients::DarkskyClient.forecast(self).dup
-    response['solarcycles'] = solarcycles
+    response['solarcycles'] = solarcycles.after_beginning_of_yesterday.all.to_a
```
mariana0pachon: What is the benefit of doing `.all.to_a` ? Wouldn't `solarcycles.after_beginning_of_yesterday` already return a shape that can be indexed?

app/models/solarcycle.rb
sracca: nit but the function name is a little confusing to me? It could be a little less wordy

cbarber: Doh! You're write. If only I had written unit tests ;)

app/models/location.rb
cbarber: Needed `.all.to_a` to fix the graphql gem not working with the active record proxy record. Probably should split these out into two separate commits.

app/models/solarcycle.rb
cbarber: Hrm. My concern was `after_yesterday` may not suggest we want `midnight` of yesterday rather than current time yesterday.

spec/models/solarcycle_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+      type: 'sunrise',
+      time: '2019-06-16T05:11:31-04:00'
+    )
+  end
+
+  after do
+    Timecop.return
+  end
+
+  it "should limit cycles to after beginning of yesterday" do
```
mariana0pachon: 🚀 yay tests

PR #118, opened 2019-06-18
Remove Scrollbar from Numbers

theroyalblah:
Changing the position from absolute to fixed prevents the vertical scrollbar from appear on the Numbers panel

### DIFF
--- a/app/javascript/components/widgets/numbers/falling-blocks.jsx
+++ b/app/javascript/components/widgets/numbers/falling-blocks.jsx
@@ -207,7 +207,7 @@ class Scene extends React.Component {
     return (
       <div
         style={{
-          position: 'absolute',
+          position: 'fixed',
           top: 0,
           left: 0,
         }}


app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -61,19 +61,19 @@ class Scene extends React.Component {
...
 
+    const wallWidth = 50;
+
     this.renderMatter = Render.create({
       element: this.scene.current,
       engine: this.engine,
       options: {
         width: window.innerWidth,
-        height: window.innerHeight,
+        height: window.innerHeight - wallWidth,
```
geraldalewis: Does the `Bodies.rectangle` code below need to be something like:

```javascript
Bodies.rectangle(
  wallWidth / 2,
  (this.state.height - wallWidth) / 2,
  wallWidth,
  this.state.height - wallWidth,
```

If so, would it make sense to add another `const` that's eq to `this.state.height - wallWidth`?

theroyalblah: @geraldalewis The actual problem is that the canvas itself is too big. If you place a body outside the canvas bounds, the canvas doesn't stretch to accommodate it, it just isn't visible. So the idea with `height: window.innerHeight - wallWidth,` is that is lowers the size of the canvas to be within the screen.

chewnoill: probably don't want to associate wall width canvas height, as these things are not related.

theroyalblah:
This PR is made irrelevant with my now, more encompassing Scrollbar Fixes PR, which includes the numbers fix.

PR #121, opened 2019-06-19
Right Side Weather View

theroyalblah:
When you aren't on the weather tab, on the right side instead of the time you can now see a small snippet of the weather.
This is handled by the top-corner-controller which selects which view, time or weather, that you should see dependent on the tab.
Screen Shot 2019-07-15 at 3 36 31 PM

### DIFF
--- a/app/javascript/components/time-hero.jsx
+++ b/app/javascript/components/time-hero.jsx
@@ -8,8 +8,9 @@ import { spacing } from '@lib/theme';
 import { LoadingMessage, ErrorMessage } from '@messages/default-messages';
 
 const Wrapper = styled(SpacedRow)`
+  margin-top: ${spacing.m};
   align-items: flex-end;
-  margin-bottom: ${spacing.xxl};
+  margin-bottom: 44px;
 `;
 
 const getLocation = gql`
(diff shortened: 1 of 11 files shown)


app/javascript/components/widgets/weather/hourly-temps.jsx
```diff
@@ -67,12 +67,12 @@ const getHourlyWeather = gql`
   }
 `;
 
-const HourlyTemps = ({ weather }) => {
+const HourlyTemps = ({ weather, hours }) => {
   const { data: hourlyWeathers } = weather.hourly;
-
+  const numHours = parseInt(hours, 10);
```
cbarber: Why not integer?

app/javascript/components/time-hero.jsx
```diff
@@ -8,8 +8,9 @@ import { spacing } from '@lib/theme';
 import { LoadingMessage, ErrorMessage } from '@messages/default-messages';
 
 const Wrapper = styled(SpacedRow)`
+  margin-top: ${spacing.m};
   align-items: flex-end;
-  margin-bottom: ${spacing.xxl};
+  margin-bottom: 26px;
```
cbarber: Does this orphan spacing.xxl?

aaddaamm:
@theroyalblah just a rebase off latest master here once the comments are addressed.

app/javascript/components/time-hero.jsx
theroyalblah: Nope, though it's only used in one other place -- fixed-content.jsx

PR #122, opened 2019-06-20
Get latest tweet from Mojotech Twitter

sracca:
- Excludes Retweets and Replies for now, harder to implement
- Retrieves id and creation date from twitter api
- IMPORTANT: Please update env.local with Helios Twitter tokens or else it will not work. (E.g. log into Twitter Devlopers under Helios acount, go under 'Apps' to 'Helios' then find under 'Tokens & Permissions')


app/graphql/types/tweet_type.rb
```diff
@@ -0,0 +1,6 @@
+class Types::TweetType < Types::BaseObject
+  graphql_name "MojoTweet"
+
+  field "id", Integer
+  field "created_at", String
```
thedelchop: I'm still new-ish to GraphQL, but aren't you missing the contents of the tweet here?

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_06_04_172236) do
+ActiveRecord::Schema.define(version: 2019_06_18_200946) do
```
thedelchop: It strikes me as odd that this value changed yet this PR doesn't seem to add/run any migrations.  Its possible I'm missing a piece WRT to GraphQL, but I think this can be removed.

app/graphql/types/tweet_type.rb
sracca: I originally had it this way as we planned on using the id to embed the tweets, but it didn't work out. In the other PR I added more of the contents (e.g. favorites, retweets, and text), but I can add in more of the fields if you think it would be better.

sracca:
@thedelchop I added in the text, favorites, retweets, etc. into this PR, as well as had it so the back-end chops off any links (api returns link to tweet w/in text) during the query because it's not necessary

app/lib/clients/twitter_client.rb
```diff
@@ -0,0 +1,21 @@
+require 'twitter'
+
+class Clients::TwitterClient
+  def self.authenticate
+    client = Twitter::REST::Client.new do |config|
+      config.consumer_key = ENV['TWITTER_CONSUMER_KEY']
+      config.consumer_secret = ENV['TWITTER_CONSUMER_SECRET']
+      config.access_token = ENV['TWITTER_ACCESS_TOKEN']
+      config.access_token_secret = ENV['TWITTER_ACCESS_SECRET']
```
geraldalewis: nit: the name mismatch bumped me, but I think this is okay 👍

app/lib/clients/twitter_client.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+    client
+  end
+
+  def self.latest_tweet
+    client = authenticate
+    # API gets count amt, then filters out rts and replies
+    # This guarantees there will always be a tweet
+    tweet = client.user_timeline('mojotech', count: 5, include_rts: false, exclude_replies: true).first
```
geraldalewis: great 👍

Gemfile
```diff
@@ -86,3 +86,6 @@ end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
+
+# twitter API
+gem "twitter", "~> 6.2"
```
geraldalewis: Let's move this commit before `chore: add twitter authentication `

app/graphql/types/unix_date_time_type.rb
```diff
@@ -1,4 +1,3 @@
-
```
geraldalewis: This snuck in here somehow

app/lib/clients/twitter_client.rb
```diff
@@ -11,11 +11,36 @@ def self.authenticate
     client
   end
 
+  TwitterQuery = Struct.new(:created_at, :text, :favorite_count, :retweet_count, :reply_count)
+
+  def self.reply_count(id, client)
+    replies = client.search("to:mojotech", since_id: id, include_entities: true)
+                    .select { |res| res.in_reply_to_status_id == id }
+    replies.count
```
geraldalewis: very nice

app/lib/clients/twitter_client.rb
```diff
@@ -11,11 +11,36 @@ def self.authenticate
...
+    clean_text
+  end
+
   def self.latest_tweet
     client = authenticate
     # API gets count amt, then filters out rts and replies
     # This guarantees there will always be a tweet
-    tweet = client.user_timeline('mojotech', count: 5, include_rts: false, exclude_replies: true).first
-    tweet
+    latest = client
```
geraldalewis: I might be missing something, but I think you could remove that initial commit entirely -- having this code change makes it a little harder for me to get a model of what it's doing. (If I missed something, pls lemme know!)

spec/graphql/events_spec.rb
```diff
@@ -35,6 +35,7 @@
       variables: variables
     )
     raise res["errors"].to_s if res["errors"]
+
```
geraldalewis: Did this sneak in?

Gemfile
sracca: Oh whoops I think my commits were out of order :/

spec/graphql/events_spec.rb
sracca: Yes :(

PR #123, opened 2019-06-20
Scrollbar Fixes

theroyalblah:
Fixes scrollbar issues surrounding weather, twitter, live-stream and numbers.

Weather panel's scrollbar problems were fixed by properly resizing the panel components so they correctly moved in relation to the screen size. By having a smaller screen size, the components will get closer together to fit, or grow further apart as the screen height increases.

Live-stream and Twitter were getting using Wrapper twice. Those two are currently placeholders, so having them export null seems appropriate.

Numbers was creating a scrollbar with its absolute positioning. Changing its positioning to fixed solves the scrollbar issue.

### DIFF
--- a/app/javascript/components/side-panel.jsx
+++ b/app/javascript/components/side-panel.jsx
@@ -11,6 +11,7 @@ const Wrapper = styled.div`
   padding: ${spacing.xxxl};
   top: 0;
   width: 600px;
+  overflow: hidden;
 `;
 
 export const SidePanel = ({ widgets, selectedWidget }) => (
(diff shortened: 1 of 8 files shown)


app/javascript/components/side-panel.jsx
```diff
@@ -11,6 +11,7 @@ const Wrapper = styled.div`
   padding: ${spacing.xxxl};
   top: 0;
   width: 600px;
+  overflow: hidden;
```
chewnoill: there we go, thats what I wanted to see

PR #125, opened 2019-06-21
fix: switch apollo graphql fetchPolicy to network-only

cbarber:
The default cache policy is resulting in queries cached on page refresh rather than querying the server. `network-only` will force new requests on page load and then rely on cache updates.


app/javascript/components/app.jsx
```diff
@@ -58,6 +58,11 @@ persistCache({
 const client = new ApolloClient({
   link,
   cache,
+  defaultOptions: {
+    query: {
+      fetchPolicy: 'network-only',
```
geraldalewis: I would think `'cache-and-network'` would be preferable, no? This will never hit the cache...

aaddaamm:
this needs a rebase @cbarber 🙏

app/javascript/components/app.jsx
cbarber: https://www.apollographql.com/docs/react/api/react-apollo/#optionsfetchpolicy

Nope. Just first request. Then we're living the caching dream.

cbarber:
@aaddaamm `git rebase -i adams-face`

aaddaamm:
did that work? i'd be flattered if you renamed your local master to `adams-face`

cbarber:
https://github.com/mojotech/helios2/pull/126

app/javascript/components/app.jsx
chewnoill: agreed, `cache-and-network` seems like the play.

PR #128, opened 2019-06-21
Blocks are drawn on a second canvas on sleep

theroyalblah:
This should allow a much larger amount of blocks to be on screen at the same time by swapping out actual blocks with a static image only rendered a single time that overlays the first canvas. So when a block goes to sleep, it first becomes static, invisible, and finally is added to the second canvas.

In order for this to occur when the tab is loaded up again, when componentDidMount is called we have to iterate over the bodies and ensure they are re-added to the second canvas.


app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -43,6 +45,10 @@ class Scene extends React.Component {
       const loadedWorld = this.res.resurrect(world);
       Engine.merge(this.engine, { world: loadedWorld });
     }
+    const allBodies = Composite.allBodies(this.engine.world);
```
chewnoill: probably don't need to keep track of allBodies here, can just filter/map from here

app/javascript/components/widgets/numbers/sleeping-blocks.jsx
```diff
@@ -55,13 +55,18 @@ Sleeping.update = function freezeOnSleep(bodies, timeScale) {
       ) {
         body.sleepCounter += 1;
 
-        if (body.sleepCounter >= body.sleepThreshold) {
+        if (body.sleepCounter >= body.sleepThreshold && body.render.visible) {
           // Bodies here set to static instead of being put to sleep
           Body.setStatic(body, true);
+          body.render.visible = false;
+          addToOverlay(body);
```
cbarber: This function is doing a little more than just `freezeOnSleep` now. Its also hiding the bodies and calling a callback. I think function composition may be a useful pattern here.

app/javascript/components/widgets/numbers/sleeping-blocks.jsx
```diff
@@ -55,13 +55,18 @@ Sleeping.update = function freezeOnSleep(bodies, timeScale) {
...
       } else if (body.sleepCounter > 0) {
         body.sleepCounter -= 1;
       }
     }
   }
-};
+}
+
+export default addToOverlay => (bodies, timeScale) =>
+  freezeOnSleep(bodies, timeScale, addToOverlay);
```
cbarber: `curry` and `compose` may simplify this

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -204,14 +230,26 @@ class Scene extends React.Component {
...
     return (
-      <div
-        style={{
-          position: 'fixed',
-          top: 0,
-          left: 0,
-        }}
-        ref={this.scene}
-      />
+      <div>
```
cbarber: Prefer React.fragment to avoid unnecessary div

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -143,6 +148,27 @@ class Scene extends React.Component {
     'count',
   ]);
 
+  addToOverlay = body => {
+    if (body && body.render.sprite.texture) {
+      const ctx = this.overlay.current.getContext('2d');
```
cbarber: https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Optimizing_canvas#Turn_off_transparency

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -204,14 +230,26 @@ class Scene extends React.Component {
...
+        />
+        <canvas
+          width={window.innerWidth}
+          height={window.innerHeight}
+          style={{
+            position: 'fixed',
+            top: 0,
+            left: 0,
+          }}
+          ref={this.overlay}
```
cbarber: Nit: this is technically an underlay.

app/javascript/components/widgets/numbers/sleeping-blocks.jsx
```diff
@@ -55,13 +55,18 @@ Sleeping.update = function freezeOnSleep(bodies, timeScale) {
       ) {
         body.sleepCounter += 1;
 
-        if (body.sleepCounter >= body.sleepThreshold) {
+        if (body.sleepCounter >= body.sleepThreshold && body.render.visible) {
```
cbarber: Why is `body.render.visible` necessary here? Is `sleepCounter` lost during serialization?

cbarber:
I read this in a little more detail and I think the issue is the freezeOnSleep is doing a little too much since its currently a trivial way to extend the matterjs engine. It may be worthwhile to investigate the events emitted from the engine: `Events.trigger(engine, 'afterUpdate', event);` and `Events.trigger(engine, 'beforeUpdate', event);` to observe for a series of composed behaviors:
```
Event.observe(engine, ...., compose(
    setBodiesStaticIfSleeping,
    hideStaticBodies,
    copyStaticBodiesToUnderlay,
))
```

https://github.com/liabru/matter-js/blob/2ec247b7af1c6b5da6ee05c73274ed5822c73503/src/core/Engine.js#L230

app/javascript/components/widgets/numbers/falling-blocks.jsx
theroyalblah: Doing this seems to destroy everything on the first canvas. Note the lack of counts at the top, and the blocks that are falling are not rendered either
Screen Shot 2019-06-24 at 2 55 50 PM

app/javascript/components/widgets/numbers/sleeping-blocks.jsx
theroyalblah: The objects that are currently "sleeping" pass the check, as they still exist but are simply invisible and static. Given that they pass this check, they would be redrawn to the second canvas as well as making unnecessary calculations. Technically this could be checking for if the body is static instead, as the body being invisible implies the body is static in our usage.

theroyalblah:
@cbarber I tried implementing this, but splitting the functions into 3 and then composing them really damaged performance. What do you think about leaving sleeping-blocks and `freezeBlocks()` alone and creating a single member function that is used as follows
`Events.on(this.engine, 'beforeUpdate', b => this.prepareOverlay(b));`
This member function would make static blocks invisible and pass on singular blocks to ` addToOverlay()`. I have a working version of this and it seems to be rather performant and I think it solves the issue of freezeBlocks doing too much. Thoughts?

cbarber:
Yeah, that sounds like an improvement. I'd like to dig into the performance issues with you for the composed functions. Could you grab the commit from the reflog and push it out on a separate branch?

app/javascript/components/widgets/numbers/falling-blocks.jsx
theroyalblah: I think it is an overlay, as setting the background of the fallingBlocks to opaque canvas does not destroy this canvas, but the reverse is true. If the overlay's background is opaque, fallingBlocks is not visible. Unless I'm thinking about this backwards

cbarber: Oh derp. You're right. I got it backwards.

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -23,6 +23,8 @@ class Scene extends React.Component {
     mutation: PropTypes.func.isRequired,
   };
 
+  imageCache = {};
```
chewnoill: Since its a separate idea, can we pull image caching into its own commit?

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -177,6 +204,47 @@ class Scene extends React.Component {
...
+    const allBodies = Composite.allBodies(this.engine.world);
+    for (let i = 0; i < allBodies.length; i += 1) {
+      if (allBodies[i].isStatic) {
+        this.invisify(allBodies[i]);
+        this.addToOverlay(allBodies[i]);
+      }
+    }
+  }
+
+  // eslint-disable-next-line class-methods-use-this
```
chewnoill: what is this for?

theroyalblah: I wanted `invisify()` to be class member function but it doesn't need to be because it never uses `this`. So if you think it shouldn't be a member function I can change it.

cbarber: You can just move it to the class. There isn't any value for a funciton that does nothing with the class to be namespaced within the class.

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -43,10 +46,20 @@ class Scene extends React.Component {
...
+      .map(b => this.addToOverlay(b));
+    Events.on(this.engine, 'beforeUpdate', b => this.prepareOverlay(b));
     this.addBlocks();
   }
 
   componentWillUnmount() {
+    const allBodies = Composite.allBodies(this.engine.world);
+    for (let i = 0; i < allBodies.length; i += 1) {
+      if (allBodies[i].onOverlay) {
+        allBodies[i].onOverlay = false;
```
cbarber: This is a strange name for a boolean. `on<something>` is generally a callback. `is<something>` / `has<something>` are better boolean indicators

app/javascript/components/widgets/numbers/falling-blocks.jsx
cbarber: Where is this used?

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -177,6 +209,39 @@ class Scene extends React.Component {
...
+    if (!body || !body.render.sprite.texture || body.onOverlay) {
+      return;
+    }
+    const ctx = this.overlay.current.getContext('2d');
+    const { texture } = body.render.sprite;
+    const { sprite } = body.render;
+    ctx.translate(body.position.x, body.position.y);
+    ctx.rotate(body.angle);
+
+    const img = this.getImageFromCache(texture);
```
cbarber: I'm not following the value of this. Module exports are already cached. What is the connection between `image` and `texture`?

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -143,6 +160,33 @@ class Scene extends React.Component {
...
+      b.render.visible = false;
+    }
+  };
+
+  invisify = body => {
+    if (body.render.visible) {
+      const b = body;
+      b.render.visible = false;
+    }
+  };
```
cbarber: This doesn't relate to the class so it can be extracted. 

Also can likely be simplified with a ramda method.

Also duplicated

app/javascript/components/widgets/numbers/falling-blocks.jsx
chewnoill: https://github.com/mojotech/helios2/pull/128/commits/fc0a94a5df76082bb99c0bc477ffee01ea723713#diff-ed8e5fea165eae9a5f4092f20661d47fR164

app/javascript/components/widgets/numbers/falling-blocks.jsx
theroyalblah: It's to try and avoid `image = new Image();` which seemed to be very expensive. Texture is pulled from the body object and contains the filepath for the image.

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -43,10 +51,23 @@ class Scene extends React.Component {
       const loadedWorld = this.res.resurrect(world);
       Engine.merge(this.engine, { world: loadedWorld });
     }
+    Composite.allBodies(this.engine.world)
+      .filter(body => body.isStatic && body.render.sprite.texture)
+      .map(b => this.addToOverlay(b));
```
cbarber: This should be a `forEach`, not `map`

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -43,10 +51,23 @@ class Scene extends React.Component {
...
     this.addBlocks();
   }
 
   componentWillUnmount() {
+    const allBodies = Composite.allBodies(this.engine.world);
+    allBodies
+      .filter(body => body.onOverlay)
+      .map(body => {
+        const b = body;
+        b.onOverlay = false;
```
cbarber: `is<something>`, `has<something>`, `should<something>` are better hints at bools. `on<something` suggests a callback.

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -159,24 +195,61 @@ class Scene extends React.Component {
...
+      img,
+      img.width * -sprite.xOffset,
+      img.height * -sprite.yOffset,
+    );
+
+    ctx.rotate(-body.angle);
+    ctx.translate(-body.position.x, -body.position.y);
+    const b = body;
+    b.isOnOverlay = true;
+    Events.off(body, 'sleepStart', overlayBody =>
```
chewnoill: not sure Events.off gets a callback

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -190,6 +263,15 @@ class Scene extends React.Component {
 
   serialise(object) {
     return this.res.stringify(object, (key, value) => {
+      // serialise fails if it attempts to work on a function, even if it's a member of an object
+      // to prevent this from occuring, we need to check objects for their first member to see if it's a function
+      if (
+        value &&
+        typeof value === typeof object &&
+        typeof value[0] === typeof (() => {})
```
chewnoill: can we check the `key` instead of the `typeof value[0]`?

app/javascript/components/widgets/numbers/falling-blocks.jsx
theroyalblah: Looks like it:
    /**
     * Removes the given event callback. If no callback, clears all callbacks in `eventNames`. If no `eventNames`, clears all events.
     * @method off
     * @param {} object
     * @param {string} eventNames
     * @param {function} callback
     */
`    Events.off = function(object, eventNames, callback) {`

https://github.com/liabru/matter-js/blob/2ec247b7af1c6b5da6ee05c73274ed5822c73503/src/core/Events.js#L45

app/javascript/components/widgets/numbers/falling-blocks.jsx
theroyalblah: This code also works, if you think it would be better:
`if (key === 'sleepStart') {
return null;
      }`

app/javascript/components/widgets/numbers/falling-blocks.jsx
chewnoill: do we need to `addToOverlay` on this callback?

app/javascript/components/widgets/numbers/falling-blocks.jsx
chewnoill: I would prefer to ignore all `key === 'events'` if possible

app/javascript/components/widgets/numbers/falling-blocks.jsx
theroyalblah: nope, it's gone now

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -143,6 +182,28 @@ class Scene extends React.Component {
     'count',
   ]);
 
+  getImageFromCache(imagePath) {
```
chewnoill: looks like this got smooshed into the wrong commit

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -143,6 +182,15 @@ class Scene extends React.Component {
     'count',
   ]);
 
+  sleepStartEvent = block => {
+    Events.on(block, 'sleepStart', () => {
```
cbarber: Javascript events typically bubble up so you may be able to listen on the `world` to reduce creating/destroying events for every body.

cbarber: Just realized this is internal to matterjs and not the javascript Event API so this isn't true.

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -159,24 +207,56 @@ class Scene extends React.Component {
...
+    ctx.drawImage(
+      img,
+      img.width * -sprite.xOffset,
+      img.height * -sprite.yOffset,
+    );
+
+    ctx.rotate(-body.angle);
+    ctx.translate(-body.position.x, -body.position.y);
+    const b = body;
+    b.isOnOverlay = true;
```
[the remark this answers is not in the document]
theroyalblah: eslint doesn't like assigning to function parameters

> Assignment to property of function parameter 'body'.eslint(no-param-reassign)

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -18,6 +26,22 @@ const wallWidth = 50;
 // Padding applied to the height of the walls to prevent blocks created contemporaneously from colliding.
 const padding = 50;
 
+// eslint-disable-next-line no-underscore-dangle
+Sleeping._motionSleepThreshold = 0.5;
```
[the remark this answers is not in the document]
chewnoill: looks like this was moved in from `@numbers/sleeping-blocks'

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -43,10 +68,23 @@ class Scene extends React.Component {
       const loadedWorld = this.res.resurrect(world);
       Engine.merge(this.engine, { world: loadedWorld });
     }
+    Composite.allBodies(this.engine.world)
+      .filter(body => !body.isStatic)
+      .map(block => this.sleepStartEvent(block));
```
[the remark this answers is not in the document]
chewnoill: why is it unnecessary? or do you just prefer foreach?

app/javascript/components/widgets/numbers/falling-blocks.jsx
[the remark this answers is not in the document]
cbarber: I think in this case I'd prefer `// eslint-disable-next-line no-param-reassign` so we're not obscuring that we're mutating the object.

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -190,6 +270,11 @@ class Scene extends React.Component {
 
   serialise(object) {
     return this.res.stringify(object, (key, value) => {
+      // serialise fails if it attempts to work on a function, even if it's a member of an object
+      // to prevent this from occuring, we need to check if there was an event
+      if (key === 'events') {
```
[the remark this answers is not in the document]
chewnoill: This implementation prevents serialize anything from the `events` branches of the object; although the comment suggests a more generalized intent.  I'm fine with either approach, the comment is a little confused though.

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -204,14 +289,26 @@ class Scene extends React.Component {
...
     return (
-      <div
-        style={{
-          position: 'fixed',
-          top: 0,
-          left: 0,
-        }}
-        ref={this.scene}
-      />
+      <React.Fragment>
```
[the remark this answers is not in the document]
chewnoill: agreed, if it works.

app/javascript/components/widgets/numbers/falling-blocks.jsx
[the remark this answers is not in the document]
theroyalblah: it's a foreach now

PR #129, opened 2019-06-25
Numbers Design Update

theroyalblah:
This should make the Numbers Panel more like the new designs. New higher resolution images had to be brought in for usage next to each subtitle, as the old images are 16x16 and these are larger than that.

Screen Shot 2019-06-25 at 9 39 02 AM


app/javascript/components/widgets/numbers/panel.jsx
```diff
@@ -57,15 +76,34 @@ class SubscribedEvents extends React.Component {
...
             <Count>{slackMessageCount}</Count>
-            <CountLabel color={colors.teal}>Slack {messageText}</CountLabel>
+            <IconTextWrapper>
+              <BlockIcon
+                src={slackMessageImage}
+                width="24"
+                height="24"
+                alt=""
+              />
+              <CountLabel color={colors.teal}>Slack {messageText}</CountLabel>
```
sracca: Have you considered using .map on this?

cbarber: I think extracting the duplicated code a named component than a map of props. The goal being:
```
<React.Fragment>
   <CountSummary count={} image={} textColor={} text={} />
   <CountSummary count={} image={} textColor={} text={} />
   <CountSummary count={} image={} textColor={} text={} />
</React.Fragment>
```

app/javascript/components/widgets/numbers/panel.jsx
```diff
@@ -63,15 +76,34 @@ class SubscribedEvents extends React.Component {
...
+              />
+              <CountLabel color={colors.pink}>{commitText}</CountLabel>
+            </IconTextWrapper>
           </div>
           <div>
             <Count>{githubPullCount}</Count>
-            <CountLabel color={colors.yellow}>Pull {requestText}</CountLabel>
+            <IconTextWrapper>
+              <BlockIcon src={githubPullImage} width="24" height="24" alt="" />
+              <CountLabel color={colors.yellow}>Pull {requestText}</CountLabel>
```
cbarber: styled-components can accept props in the style definition. Is this generally preferred to adding inline styles like color, width, height.

PR #131, opened 2019-06-26
Blocks Spawn With Normal Distribution

theroyalblah:
Blocks now spawn exclusively to the left of the side panel, and spawn with a normal distribution to create a nice pile.
Screen Shot 2019-06-26 at 2 54 39 PM


app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -153,8 +154,11 @@ class Scene extends React.Component {
 
   addBlocks = () => {
     const block = this.nextBlock();
-    const randomPositionBetweenWalls =
-      (this.state.width - wallWidth) * Math.random() + wallWidth;
+    const widthBetweenWalls =
+      this.state.width - wallWidth - parseInt(width, 10);
+    const randomDist =
+      (widthBetweenWalls * Math.random() + widthBetweenWalls * Math.random()) /
```
cbarber: This can be further reduced to: `widthBetweenWalls * (Math.random() + Math.random()) / 2`

I think it may also be beneficial to note in the commit message that this is an approximation of a normal distribution, not an actual normal distribution.

app/javascript/components/side-panel.jsx
```diff
@@ -4,13 +4,15 @@ import { spacing } from '@lib/theme';
 import FixedContent from '@components/fixed-content';
 import widgetShape from '@lib/widget-shape';
 
+export const width = '600px';
```
cbarber: Sorry, I wasn't clear before. This can remain an int. You would just need to modify the styles in Wrapper to be `${width}px;` which would mean you don't need the `parseInt`'s further down

PR #133, opened 2019-07-08
feat: quotes, retweets, and multiple images are now supported

sracca:
- All views are supported in primary tweet now
- User profiles are now gathered via Twitter Client
- Note: When reviewing, may need to switch from 'mojotech' to 'helios' user_timeline to test things as mojotech twitter rarely retweets/posts multiple images etc.
- Note: Replies automatically come in as '@UserRepliedTo' via client, so no need to create separate view for them as it is essentially another tweet by user. Replies are different then Quotes as a Quote is essentially a retweet + a reply.
Screen Shot 2019-07-01 at 1 20 34 PM
Screen Shot 2019-07-02 at 1 11 20 PM
Screen Shot 2019-07-08 at 1 29 50 PM


app/lib/clients/twitter_client.rb
```diff
@@ -1,17 +1,23 @@
...
 
-  TwitterQuery = Struct.new(:created_at, :text, :favorite_count, :retweet_count, :media, :link)
+  InteractionQuery = Struct.new(:favorite_count, :retweet_count)
+
+  MediaQuery = Struct.new(:image, :link)
+
+  TwitterQuery = Struct.new(:created_at, :text, :interactions, :media)
+
+  def self.media_query(latest)
+    MediaQuery.new(get_media(latest), get_link(latest))
```
cbarber: It may make more sense for latest to be an object that has these methods rather than class methods in client. Clients responsibility is to request and return the data. Another object could be in charge of how to resolve the data for the graphql query response

app/javascript/schema.json
```diff
@@ -2369,6 +2369,105 @@
               "isDeprecated": false,
               "deprecationReason": null
             },
+            {
+              "name": "interactions",
```
cbarber: The schema change appears to be from the previous commit. I'm ok with a general purpose "adding extended twitter data" commit that covers interaction, profile, media, etc.

app/graphql/types/tweet_type.rb
```diff
@@ -1,10 +1,22 @@
+class Types::InteractionType < Types::BaseObject
+  graphql_name "TweetInter"
```
cbarber: Nit: no need to abbreviate here

app/lib/clients/twitter_client.rb
```diff
@@ -13,7 +12,23 @@ class Clients::TwitterClient
 
   MediaQuery = Struct.new(:image, :link)
 
-  TwitterQuery = Struct.new(:created_at, :text, :interactions, :media)
+  UserQuery = Struct.new(:name, :handle, :avatar)
+
+  TwitterQuery = Struct.new(:status, :created_at, :text, :interactions, :media, :user)
+
+  def self.latest_tweet
```
cbarber: The function move here looks like a rubocop fix up?

app/lib/clients/twitter_client.rb
```diff
@@ -44,17 +59,25 @@ def self.remove_hyperlinks(text)
...
 
-  def self.latest_tweet
-
-    # API gets count amt, then filters out rts and replies
-    # This guarantees there will always be a tweet
-    latest = @client
-             .user_timeline('mojotech', count: 5, include_rts: true,
-             exclude_replies: false, tweet_mode: 'extended')
-             .first
+  def self.det_status(latest)
```
cbarber: Yeah, this feels out of place in `Client`

spec/graphql/tweet_spec.rb
```diff
@@ -68,12 +89,28 @@
...
+        media = result["data"]["tweets"].first["media"]
+        expect(media["image"]).to eq("http://pbs.twimg.com/media/D-KsunCW4AA7uz8.jpg")
+        expect(media["link"]).to eq(nil)
+      end
+
+      it 'will return valid user data' do
+        user = result["data"]["tweets"].first["user"]
+        expect(user["avatar"]).to eq("https://pbs.twimg.com/profile_images/775765308514660352/r1OvXiwV.jpg")
+        expect(user["handle"]).to eq("MojoTech")
+        expect(user["name"]).to eq("MojoTech")
```
cbarber: :100:

app/javascript/components/widgets/twitter/tweet_body.jsx
```diff
@@ -41,21 +41,27 @@ OneMediaType.propTypes = {
   linkUrl: PropTypes.string.isRequired,
 };
 
-const TweetBody = ({ text, mediaUrl, linkUrl }) => {
-  if (mediaUrl === null && linkUrl === null) {
-    return <TweetText text={text} />;
+const removeRT = text => {
+  return text.substr(text.indexOf(' ') + 1);
```
cbarber: Does this mean `removeRetweet`? Prefer avoiding acronyms over full terms in functions

Is the text before the space always dynamic? Could this be a little more explicit?

app/javascript/components/widgets/twitter/panel.jsx
```diff
@@ -85,8 +85,8 @@ const RetweetWrapper = styled.div`
 `;
 
 export const TabBar = () => (
-  <svg width="960" height="4">
-    <rect width="960" height="1" fill={colors.white} opacity="0.2" />
+  <svg width="1020" height="4">
+    <rect width="1020" height="1" fill={colors.white} opacity="0.2" />
```
cbarber: I noticed on the projector that the twitter panel can go behind the tab bar. Is this causing that side effect? Ivan thght it would be best for the tweets to be constrained to the left of tabs.

app/javascript/components/widgets/twitter/tweet_body.jsx
```diff
@@ -1,16 +1,29 @@
...
   color: ${colors.white};
   text-decoration: underline;
   font-size: ${fontSizes.medium};
 `;
 
-const BothMediaTypes = ({ displayText, mediaUrl, linkUrl }) => (
+const quoteStyle = {
+  fontFamily: `${fonts.regular}`,
+  height: '192px',
+  maxWidth: '100%',
```
cbarber: This seems odd.

app/javascript/components/widgets/twitter/tweet_body.jsx
```diff
@@ -41,15 +54,34 @@ OneMediaType.propTypes = {
...
+    </MediaWrapper>
+  </React.Fragment>
+);
+
+QuoteTweet.propTypes = {
+  displayText: PropTypes.string.isRequired,
+  linkUrl: PropTypes.string.isRequired,
+};
+
+// The 'RT' symbol given to retweets by twitter
```
cbarber: Oh ok. I think this may have helped clarify the preceding commit. I think `RetweetSymbol` would also be clearer.

app/javascript/components/widgets/twitter/tweet_body.jsx
```diff
@@ -23,35 +28,45 @@ const quoteStyle = {
...
+BothMediaTypes.defaultProps = {
+  image: null,
+  linkUrl: null,
+};
+
 BothMediaTypes.propTypes = {
   displayText: PropTypes.string.isRequired,
-  mediaUrl: PropTypes.string.isRequired,
-  linkUrl: PropTypes.string.isRequired,
+  image: PropTypes.arrayOf(PropTypes.string),
```
cbarber: Should be plural.

app/javascript/components/widgets/twitter/tweet_content.jsx
```diff
@@ -61,14 +71,29 @@ TweetText.propTypes = {
...
+export const TweetMedia = ({ images, linkUrl }) => {
   return (
     <MediaWrapper>
-      <MediaImage
-        src={mediaUrl}
-        alt="twitter"
-        mediaDisplay={mediaUrl !== null ? 'block' : 'none'}
-      />
+      {images !== null && (
+        <ImageGrid oneImg={images.length === 1}>
```
cbarber: `ImageGrid` does not use `oneImg` prop

app/javascript/components/widgets/twitter/tweet_content.jsx
```diff
@@ -15,22 +15,32 @@ export const MediaWrapper = styled.div`
...
+  height: 487px;
+  width: 100%;
+  grid-gap: ${spacing.s};
+  grid-template-columns: auto auto;
+  overflow: hidden;
+  border-radius: 6px;
+`;
+
+const ImageWrapper = styled.div`
+  grid-row: ${props => (props.main ? ' 1 / 3' : null)};
```
cbarber: I think `auto` is the default value for `grid-row`

app/lib/clients/twitter_client.rb
```diff
@@ -27,15 +27,20 @@ def self.latest_tweet
...
   def self.media_query(latest)
     MediaQuery.new(get_media(latest), get_link(latest))
   end
 
   def self.get_media(latest)
-    latest.attrs[:extended_entities][:media].first[:media_url]
+    images = []
+    latest.attrs[:extended_entities][:media].each do |item|
+      images.push(item[:media_url])
+    end
```
cbarber: Prefer https://ruby-doc.org/core-2.6.3/Enumerable.html#method-i-map

app/graphql/types/tweet_type.rb
```diff
@@ -16,7 +16,7 @@ class Types::InteractionType < Types::BaseObject
 class Types::MediaType < Types::BaseObject
   graphql_name "TweetMedia"
 
-  field "image", String, null: true
+  field "image", [String], null: true
```
cbarber: Prefer plural here as well.

app/javascript/components/widgets/weather/panel.jsx
```diff
@@ -132,7 +132,7 @@ SubscribedWeather.propTypes = {
 };
 
 export default () => (
-  <Query query={getPrimaryLocationWeather}>
+  <Query query={getPrimaryLocationWeather} fetchPolicy="cache-and-network">
```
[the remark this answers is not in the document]
sracca: Oh whoops, I think that was due to me copying the weather widget's panel implimentation before that was fixed.

app/javascript/components/widgets/twitter/tweet_body.jsx
sracca: Odd how?

app/javascript/components/widgets/twitter/panel.jsx
sracca: No, this is merely a divider between the previous and current tweets. This wouldn't cause that behavior. I can constrain it though

sracca:
@cbarber, everything is fixed, and I ended up constraining tweets to only so much of the view width to prevent overflow!

cbarber:
@sracca looks like a unit test failed

cbarber:
@sracca looks like you need to `rake graphql:schema:idl` and `rake graphql:schema:json` on rails graphql schema changes.

I think you can safely squash up the multiple image fixup commit into a single `image` commit.

I'm seeing a weird console error locally so maybe something we can look at monday.

sracca:
@cbarber fixed!!

app/javascript/components/widgets/twitter/panel.jsx
```diff
@@ -15,10 +15,14 @@ const getMojoTweets = gql`
...
-      media
-      link
+      interactions {
+        favoriteCount
+        retweetCount
+      }
+      media {
+        image
+        link
+      }
```
cbarber: This seems like an unnecessary change when you're going to later switch this to multiple images.

app/javascript/components/widgets/twitter/tweet_stats.jsx
```diff
@@ -40,16 +40,21 @@ StatsIcon.propTypes = {
...
 };
 
-const TweetStats = ({ favorites, retweets }) => (
-  <InteractionWrapper>
-    <StatsIcon icon={retweetIcon} count={retweets} />
-    <StatsIcon icon={likeIcon} count={favorites} />
-  </InteractionWrapper>
-);
+const TweetStats = ({ interactions }) => {
+  const { favoriteCount, retweetCount } = interactions;
```
cbarber: Nit: could deconstruct in method signature to reduce diff churn here.

app/lib/clients/twitter_client.rb
```diff
@@ -1,53 +1,20 @@
...
-    client = authenticate
-
     # API gets count amt, then filters out rts and replies
     # This guarantees there will always be a tweet
-    latest = client
-             .user_timeline('mojotech', count: 5, include_rts: true, 
-              exclude_replies: false, tweet_mode: 'extended')
+    latest = @client
+             .user_timeline('mojotech', count: 5, include_rts: true,
+                                        exclude_replies: false, tweet_mode: 'extended')
```
cbarber: This seems like a rubocop fixup from the previous commit so this should be in the RTS commit.

app/workers/tweet_worker.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def text
+    text = @tweet.attrs[:full_text]
+    links = text.split(/\s|\n/).select { |word| word.include? "http" }
+    clean_text = text
+    links.each do |l|
+      clean_text = clean_text.gsub(l, "")
+    end
+    clean_text
+ end
```
cbarber: Indentation is off here.

app/workers/tweet_worker.rb
```diff
@@ -0,0 +1,47 @@
+# Polls weather for subscribed locations
+class TweetWorker
```
cbarber: I like that you split out these methods into a separate object but its not really a worker because its not doing any polling. I'd be fine with this living in `app/clients` for now because its data is closely related to a client.

app/workers/tweet_worker.rb
```diff
@@ -44,4 +44,23 @@ def text
...
+    end
+  end
+
+  def user
+    user = if status == "retweet"
+             @tweet.attrs[:retweeted_status][:user]
+           else
+             @tweet.attrs[:user]
+    end
+    UserQuery.new(user[:name], user[:screen_name], user[:profile_image_url].remove("_normal"))
```
cbarber: What does `remove("_normal")` do here?

Ah ok. Just figured this out. This is a great instance where naming this would be really helpful:
```
large_profile_image_url = user[:profile_image_url].remove("_normal"))
```

app/javascript/components/widgets/twitter/panel.jsx
```diff
@@ -71,13 +92,18 @@ export const TabBar = () => (
 
 const Twitter = ({ tweets }) => {
   const latestTweet = tweets[0];
-  const { createdAt, text, interactions, media, user } = latestTweet;
+  const { status, createdAt, text, interactions, media, user } = latestTweet;
 
   return (
     <PanelWrapper>
+      <RetweetWrapper displayStyle={status === 'retweet' ? 'flex' : 'none'}>
```
cbarber: Nit: prefer this ternary was moved into the styles and you pass `status` as a prop

app/javascript/components/widgets/twitter/panel.jsx
sracca: Even though this is part of a commit to refactor into sub-sections?

cbarber: Hrm. Yeah, that makes sense :+1:

PR #135, opened 2019-07-10
fix: remove blocks when counts are reset

theroyalblah:
The blocks would remain, even if the counts were reset. To account for this, we can refuse to merge with the saved world if the amount of blocks is greater than the count. We also need to reset the state of githubCommit, githubPull and slackMessages to 0, to prevent the saved state from being greater than the count.

### DIFF
--- a/app/javascript/components/widgets/numbers/falling-blocks.jsx
+++ b/app/javascript/components/widgets/numbers/falling-blocks.jsx
@@ -83,7 +83,17 @@ class Scene extends React.Component {
     const { world } = this.getLocalState(this.props);
     if (world) {
       const loadedWorld = this.res.resurrect(world);
-      Engine.merge(this.engine, { world: loadedWorld });
+      const counts = this.getCountsFromProps(this.props);
+
+      if (
+        this.state.githubCommit <= counts.githubCommit &&
+        this.state.githubPull <= counts.githubPull &&
+        this.state.slackMessage <= counts.slackMessage
+      ) {
+        Engine.merge(this.engine, { world: loadedWorld });
+      } else {
+        this.setState({ githubCommit: 0, githubPull: 0, slackMessage: 0 });
+      }
     }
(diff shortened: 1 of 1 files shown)


app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -83,7 +83,17 @@ class Scene extends React.Component {
     const { world } = this.getLocalState(this.props);
     if (world) {
       const loadedWorld = this.res.resurrect(world);
-      Engine.merge(this.engine, { world: loadedWorld });
+      const counts = this.getCountsFromProps(this.props);
+      const totalBlocks =
+        counts.githubPull + counts.githubCommit + counts.slackMessage;
+      const walls = loadedWorld.bodies.filter(
```
chewnoill: counts are stored in local state, next to the world, you shouldn't need to count them again.

theroyalblah: To my understanding what we store in local state is not the total number of messages, but the progress towards hitting that number. So if numbers has created 50 Commit blocks, then we can expect githubCommit in local storage to be 50. What we want here is the total amount, so we can retrieve that number from props.

chewnoill: > What we want here is the total amount, so we can retrieve that number from props.

we do want this number, but thats not what this line does. It rather confusingly counts the number of walls from the total number of bodies, to arrive at the number of bodies in the world. This is unnecessary, because we store the number of bodies that got added to the world in local storage, and store that in the component state.

The reset condition could more simply be 
```javascript
this.state.githubPull > counts.gitHubPull 
  || this.state.gitHubCommit > counts.gitHubCommit 
  || this.state.slackMessages > counts.slackMessages
```

theroyalblah: Yeah that makes a lot more sense

PR #139, opened 2019-07-19
Fix apollo issues

cbarber:
Fixes: setting correct fetchPolicy for <Query> components
Fixes: falling block initialization (engine was cleared on all component mounts)
Fixes: tearing down world correctly (we did not kill the renderer and runner)


app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -219,6 +218,14 @@ class Scene extends React.Component {
     'events',
     'count',
   ]);
+  teardownWorld = () => {
```
theroyalblah: Needs a new line between 220 and 221 here to make CircleCi happy

PR #147, opened 2019-07-26
Fix Message Margins

theroyalblah:
Previously message had some weird quirks that ended in it creating large margins where it shouldn't have been.

### DIFF
--- a/app/javascript/components/widgets/messages/message.jsx
+++ b/app/javascript/components/widgets/messages/message.jsx
@@ -10,10 +10,10 @@ export const DisconnectedMessage = () => (
   <Message message="Looks like we're disconnected" />
 );
 export const WeatherLoadingMessage = () => (
-  <Message imageWidth="0" message="Loading the weather..." />
+  <WeatherMessage message="Loading the weather..." />
 );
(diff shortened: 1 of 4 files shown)


app/javascript/components/widgets/messages/message.jsx
```diff
@@ -27,42 +30,37 @@ const Column = styled.div`
...
+          {showImage && <SunMoon src={sunMoon} />}
+          <MessageDisplay>{message}</MessageDisplay>
+        </Column>
+      )}
+      {!showImage && (
+        <Column>
+          <MessageDisplay>{message}</MessageDisplay>
+        </Column>
+      )}
+    </>
```
cbarber: This seems odd. Couldn't we just use `{showImage && <SunMoon src={sunMoon} />}` and not need the outer `showImage &&` or containing fragment

cbarber: Ah ok. `Column` eventually changes. Is image the only difference between these two or is the split SideBar message vs Panel message? If its the latter, I think it would be better to just name two separate components that are exported.

PR #148, opened 2019-07-29
fix: create DateTimeType for standard iso8601 parsing

cbarber:
Ruby DateTime objects were formatted as `"2019-07-29 17:10:05 UTC"` which worked in Chrome but not Firefox. This PR adds a new type to format as iso8601

https://xkcd.com/927/


aaddaamm:
@cbarber probably don't need that merge commit right?

PR #149, opened 2019-07-31
Update Top Corner Contents

theroyalblah:
This PR brings the weather and time contents in the top corner of helios up to date with the new designs. 
- The time now always displays, regardless of what panel has focus.
- A small weather indicator appears to the right of the time on non-weather panels. 
- This weather indicator does not appear on the weather panel to avoid making a duplicate weather subscription

Screen Shot 2019-07-31 at 9 21 31 AM


app/javascript/components/top-corner-controller.jsx
```diff
@@ -37,7 +37,7 @@ const TopCorner = ({ showWeather }) => {
       </IconWrapper>
       <Row>
         <TimeHero />
-        {!showWeather && <WeatherCorner />}
+        {showWeather && <WeatherCorner />}
```
cbarber: :thinking: looks like this is a bug fix from the previous commit. I think changing the design to properly fit when always displaying the weather makes sense as a single commit.

app/javascript/components/weather-corner.jsx
```diff
@@ -20,19 +17,13 @@ import {
...
     ...SunriseSunsetWeather
+    ...CurrentIcon
   }
 
   ${CurrentTemp.fragments.weather}
-  ${CurrentPercip.fragments.weather}
-  ${HourlyTemps.fragments.weather}
-  ${MinutelyWeather.fragments.weather}
-  ${DailyWeather.fragments.weather}
   ${SunriseSunset.fragments.weather}
```
cbarber: Not technically related to this PR, but SunriseSunset data does not seem to be used anywhere in the corner weather.

PR #152, opened 2019-08-06
Shift hourly weathers

cbarber:
Shift hourly weather to skip current hour

### DIFF
--- a/app/javascript/components/widgets/weather/hourly-temps.jsx
+++ b/app/javascript/components/widgets/weather/hourly-temps.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import gql from 'graphql-tag';
-import { compose, pathOr, take } from 'ramda';
+import { compose, pathOr, slice } from 'ramda';
 import styled from 'styled-components';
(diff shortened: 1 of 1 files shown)


app/javascript/components/widgets/weather/hourly-temps.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import gql from 'graphql-tag';
-import { compose, slice, pathOr } from 'ramda';
+import { compose, pathOr, slice } from 'ramda';
```
theroyalblah: Is there any reason to reorder these here from the previous commit? Does the linter require them to be alphabetical?

aaddaamm: could we remove slice from the last commit and add it as a new import from ramda in this commit? it's added in the previous commit but unused. it is then used in this commit so it should probably only be introduced here instead of the previous commit.

cbarber: Prefer alphabetical. I don't think we have the rule enforcing it. `slice` was a mistake in the first commit.

PR #153, opened 2019-08-07
chore: fix issues with twitter display

sracca:
- refactor `create_tweet` so that attributes are retrieved based on type of status (i.e. images come from `...[:retweeted_status][:images] if retweet`
- decode html entities given from text
- fix size ratio of images in grid


app/javascript/components/widgets/twitter/tweet_body.jsx
```diff
@@ -10,6 +10,8 @@ import {
 } from '@twitter/tweet_content';
 import { colors, fonts, fontSizes } from '../../../lib/theme';
 
+const Entities = require('html-entities').XmlEntities;
```
cbarber: Prefer es6 import style

app/javascript/components/widgets/twitter/tweet_content.jsx
```diff
@@ -23,8 +23,8 @@ export const Text = styled.div`
 `;
 
 const TwitterImage = styled.img`
-  height: ${props => (props.oneImg ? '487px' : 'auto')};
-  width: auto;
+  height: ${props => (props.oneImg ? '487px' : props.isMain ? '100%' : 'auto')};
```
cbarber: This nested ternary is causing a linter failure. Would it make sense to constrain the Wrapper to a specific height if there is only one image?

app/lib/clients/create_tweet.rb
```diff
@@ -49,21 +59,17 @@ def text
   end
 
   def status
-    if @tweet.attrs[:retweeted]
+    if !@tweet.attrs[:retweeted_status].nil?
```
cbarber: `@tweet.attrs[:retweeted_status]` or `@tweet.attrs.has_key(:retweeted_status]`

This and attributes would be a great place for some test coverage.

app/javascript/components/widgets/twitter/tweet_content.jsx
sracca: Do you mean as like passing it as a style in the component?

cbarber: No, I meant if we constrain the ImageWrapper to a height based on `oneImg` you could have the TwitterImage only concern itself with `isMain` or not (assuming `isMain` is true when `oneImg` is true)

sracca: Well imageWrapper is dependent on the amount of images, the size is either the full height or half the height

app/javascript/components/widgets/twitter/tweet_body.jsx
sracca: It seems that import style doesn't seem to work. It completely breaks that way.
https://www.npmjs.com/package/html-entities

sracca:
@cbarber all set!

app/javascript/components/widgets/twitter/tweet_body.jsx
```diff
@@ -10,6 +10,8 @@ import {
 } from '@twitter/tweet_content';
 import { colors, fonts, fontSizes } from '../../../lib/theme';
 
+const Entities = require('html-entities').AllHtmlEntities;
```
cbarber: Prefer es6 import style.

I don't think we want to allow `AllHtmlEntities` here (e.g. something like `<script>doSomethingEvil()</script>` would be bad here). What is this allowing and can we use an explicit whitelist?

app/javascript/components/widgets/twitter/tweet_body.jsx
geraldalewis: @sracca I think there's just a little mismatch between the idioms from each way of importing. 

```javascript
const Entities = require('html-entities').AllHtmlEntities;
```

Is saying: **1** include the `'html-entities'` module, **2** get the `AllHtmlEntities` from it, and **3** _rename_ it to "`Entities`". 

In `import` terms, that'd look like:

**1** Import the module (we won't use it like this; just explaining the mapping):
```javascript
import * as HtmlEntities from 'html-entities';
```
**2** To get the `AllHtmlEntities` member, we'd do: 
```javascript
import { AllHtmlEntities } from 'html-entities'`
```
**3** Finally, to do the rename:
```javascript
import { AllHtmlEntities as Entities } from 'html-entities';
```

**3** _should_ work here

app/javascript/components/widgets/twitter/tweet_body.jsx
sracca: I'm not sure, this is the way it's recommended in https://www.npmjs.com/package/html-entities, and it doesn't seem to work any either way

sracca: Nvm, just saw Gerald's stuff above. I'll look into that

PR #155, opened 2019-08-14
Livestream Panel

theroyalblah:
Here's a start at the Boulder livestream.

The idea here is that each office would have their own unlisted livestream of their respective office, and the other would view it. From my understanding, Youtube livestreams can operate 24/7, so this could be left on if desired.

The livestream url is set in the env file and should ideally follow this pattern:
https://www.youtube.com/embed/live_stream?channel=[channel_id]&autoplay=1&mute=1&loop=1

This should create a permanent youtube livestream link that is associated with the channel, and not that specific livestream. This would still connect if the livestream was taken down and put back up later.

I'm also not sure if we even want to do this, as I have no idea if people are okay with being livestreamed in the first place.

### DIFF
--- a/app/javascript/components/widgets/livestream/panel.jsx
+++ b/app/javascript/components/widgets/livestream/panel.jsx
@@ -1,6 +1,16 @@
 import React from 'react';
 import styled from 'styled-components';
+import { Query } from 'react-apollo';
+import gql from 'graphql-tag';
+import { LoadingMessage, DisconnectedMessage } from '@messages/message';
 
+const getLivestreamUrl = gql`
+  {
+    primaryLocation {
+      livestreamUrl
+    }
+  }
+`;
 const Wrapper = styled.div`
   position: fixed;
   top: 0px;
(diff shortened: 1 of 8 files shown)


app/javascript/components/widgets/livestream/panel.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+const getLivestreamUrl = gql`
+  {
+    primaryLocation {
+      livestreamUrl
+    }
+  }
+`;
+const Wrapper = styled.div`
+  position: fixed;
```
geraldalewis: nice 👍

app/javascript/components/widgets/livestream/panel.jsx
```diff
@@ -0,0 +1,49 @@
...
+      }
+
+      if (error) {
+        return <DisconnectedMessage />;
+      }
+
+      const { livestreamUrl } = data.primaryLocation;
+      return (
+        <Wrapper>
+          <iframe
```
geraldalewis: Don't see much use of `iframe`s these days :)

chewnoill:
closed in favor of not having this feature :sob:

cbarber:
At some point we will have a discussion about privacy concerns. Until then, I'd like this left open so we don't lose track of the work.

cbarber:

PR #156, opened 2019-08-14
Traffic Alerts

theroyalblah:
There was some desire to see the state of traffic when leaving for the day, so this should help notify people where the heaviest traffic is around Providence. 

This panel only appears after 4pm every day, and will disable itself at 11pm, so it only appears during times it would be helpful. 

The images refresh when the panel is loaded, so they are reasonably up to date for viewing.

Screen Shot 2019-08-15 at 2 09 06 PM

### DIFF
--- a/app/javascript/components/widget-controller.jsx
+++ b/app/javascript/components/widget-controller.jsx
@@ -35,6 +35,9 @@ const widgets = [
     panel: <Traffic.Panel />,
     text: 'Traffic',
     showWeather: true,
+    hourStart: 16,
+    hourStop: 23,
+    isVisible: false,
   },
 ];
 
@@ -71,19 +74,35 @@ export class WidgetController extends React.Component {
 
   moveDown = () => {
     this.setState(({ index }) => ({
-      index: (index + 1) % widgets.length,
+      index: (index + 1) % this.getVisibleWidgets().length,
     }));
   };
 
   moveUp = () => {
     this.setState(({ index }) => ({
-      index: mathMod(index - 1, widgets.length),
+      index: mathMod(index - 1, this.getVisibleWidgets().length),
     }));
+  };
+
+  getVisibleWidgets = () => {
+    const date = new Date();
+    const currentHour = date.getHours();
+
+    return widgets.filter(widget => {
+      if (widget.hourStart !== undefined) {
+        if (currentHour >= widget.hourStart && currentHour < widget.hourStop) {
+          return true;
+        }
+        return false;
+      }
+      return true;
+    });
   };
 
   render() {
     const { index } = this.state;
-    const currentWidget = widgets[index];
+    const visibleWidgets = this.getVisibleWidgets();
+    const currentWidget = visibleWidgets[index];
 
     return (
       // eslint-disable-next-line
@@ -95,7 +114,7 @@ export class WidgetController extends React.Component {
       >
         <FullPanel currentWidget={currentWidget.panel} />
         <SidePanel
-          widgets={widgets}
+          widgets={visibleWidgets}
           selectedWidget={index}
           totalTime={SWITCH_INTERVAL}
           tabDown={this.moveDown}
--- a/app/javascript/components/widget-controller.jsx
+++ b/app/javascript/components/widget-controller.jsx
@@ -6,6 +6,7 @@ import SidePanel from '@components/side-panel';
 import Twitter from '@widgets/twitter';
 import Numbers from '@widgets/numbers';
 import Weather from '@widgets/weather';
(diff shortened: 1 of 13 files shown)


cpjolicoeur:
Rt 6 @ Tobey Street please

app/javascript/components/widgets/traffic/panel.jsx
```diff
@@ -1,7 +1,66 @@
...
+  text-align: center;
+  margin-right: 750px;
+`;
+
+const Traffic = () => (
+  <>
+    <Row>
+      <Column>
+        <Title>I-95 at Broad St</Title>
+        <TrafficCam src="http://www.dot.ri.gov/img/travel/camimages/95-21 I-95 S @ Broad St .jpg" />
```
chewnoill: people who put whitespace in their URLs are savages. 

Would we want to keep these on the backend, since this application wants to be location aware?

app/javascript/components/widget-controller.jsx
```diff
@@ -71,19 +79,38 @@ export class WidgetController extends React.Component {
...
+        return false;
+      }
+      return true;
+    });
+  };
+
   render() {
     const { index } = this.state;
-    const currentWidget = widgets[index];
+    this.visibleWidgets = this.setVisibleWidgets();
```
chewnoill: if we call `setVisibleWidgets()` on render we don't really need to do it anywhere else.

Probably we could rename it to `getVisibleWidgets` and skip storing it in the class instance also.

app/javascript/components/widgets/traffic/panel.jsx
```diff
@@ -1,7 +1,66 @@
 import React from 'react';
+import styled from 'styled-components';
+import { colors, weights, fonts, fontSizes } from '@lib/theme';
+import { GreySubText } from '@components/typography';
 
-const Traffic = () => {
-  return <div>Traffic Panel!</div>;
-};
```
cbarber: Can squash these. Not much value to a placeholder in a single PR.

app/javascript/components/widgets/traffic/panel.jsx
cbarber: These should be properly uri encoded. I'm ok with this as an incremental step to putting them in the database.

app/javascript/components/widget-controller.jsx
```diff
@@ -71,19 +79,38 @@ export class WidgetController extends React.Component {
...
   };
 
+  setVisibleWidgets = () => {
+    const date = new Date();
+    const currentHour = date.getHours();
+
+    return widgets.filter(widget => {
+      if (widget.timeStart) {
+        if (widget.timeStart <= currentHour) {
+          if (widget.timeEnd >= currentHour) {
```
cbarber: This can be a single condition if you invert your comparisons.

app/javascript/components/widget-controller.jsx
```diff
@@ -44,6 +47,11 @@ export class WidgetController extends React.Component {
     this.state = {
       index: 0,
     };
+    this.visibleWidgets = this.setVisibleWidgets();
```
cbarber: Don't put state in instance variables.

app/javascript/components/widget-controller.jsx
```diff
@@ -71,19 +79,38 @@ export class WidgetController extends React.Component {
 
   moveDown = () => {
     this.setState(({ index }) => ({
-      index: (index + 1) % widgets.length,
+      index: (index + 1) % this.visibleWidgets.length,
```
cbarber: Probably simplest to just call `getVisibleWidgets().length`

app/javascript/components/widget-controller.jsx
```diff
@@ -65,19 +79,38 @@ export class WidgetController extends React.Component {
...
-      index: mathMod(index - 1, widgets.length),
+      index: mathMod(index - 1, this.visibleWidgets.length),
     }));
   };
 
+  setVisibleWidgets = () => {
+    const date = new Date();
+    const currentHour = date.getHours();
+
+    return widgets.filter(widget => {
```
geraldalewis: @theroyalblah Since it's based on the _hour_, I think renaming `timeStart` and `timeEnd` to `hourStart` and `hourEnd` would make the code more self-documenting.

Also, since `0` is false-y, `if (widget.timeStart) ` will fail if the start time is midnight (`0`).

I think you should be able to rewrite this using a single line and enhance clarity (also, consider destructuring the `widget` props you use).

cbarber: Oh other pet peeve. Start/Stop or Begin/End. Never mixed.

PR #160, opened 2019-09-13
fix: switch local state to filtered world prop

cbarber:
The matterjs world was being serialized as a large javascript object and
compressed with zlib. This switches that serialization to a filtered
array of javascript object properties. The local state of counts is then
inferred from the world when the graphql query is done loading.

### DIFF
--- a/app/javascript/components/widgets/numbers/ducks.jsx
+++ b/app/javascript/components/widgets/numbers/ducks.jsx
@@ -2,22 +2,18 @@
 import React from 'react';
 import { withQuery } from 'react-apollo';
 import gql from 'graphql-tag';
-import { inflateSync, deflateSync } from 'zlib';
 
 export const withLocalMutation = WrappedComponent => props => (
   <WrappedComponent
     {...{
       ...props,
-      mutation: variables => {
+      mutation: ({ world }) => {
         // eslint-disable-next-line react/prop-types
         props.client.writeData({
           data: {
             fallingBlocksState: {
               __typename: 'FallingBlocksState',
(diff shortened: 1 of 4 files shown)


cbarber:
@chewnoill 

 blocks | uncompressed | compressed size 
---------|-------|----|----
 100 | 480K | 80 K | 11 K
 500|  2.3M |403 K | 58 K
1700 | 7.8 M |  ? | 197 K
 2.5k | error?  | 2 M | 348 K

app/javascript/components/widgets/numbers/ducks.jsx
```diff
@@ -27,51 +23,29 @@ export const withLocalMutation = WrappedComponent => props => (
...
-  fallingBlocksState: {
-    githubCommit = 0,
-    githubPull = 0,
-    slackMessage = 0,
-    world = null,
-  } = {},
+  fallingBlocksState: { world = null } = {},
 } = {}) => {
-  if (!world) {
+  if (!world || typeof world !== 'object') {
```
geraldalewis: Nit -- took me a couple reads to understand this -- the `typeof world !== 'object'` is there for some type checking? If so, maybe `Array.isArray(world)` might be more clear (and would obviate the need for `!world ||`).

cbarber: Yeah, that's better. This was in here to safe guard against the old zlib data from sneaking in.

cbarber:
Been playing around with this and found some bugs. Cleaning them up.

PR #161, opened 2019-09-13
Cleanup traffic issues

cbarber:
Addresses traffic cam image sizing and prop key warning.

### DIFF
--- a/app/javascript/components/widgets/traffic/panel.jsx
+++ b/app/javascript/components/widgets/traffic/panel.jsx
@@ -12,6 +12,7 @@ const getTrafficCams = gql`
   {
     primaryLocation {
       trafficCams {
+        id
         title
         url
       }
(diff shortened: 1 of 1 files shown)


app/javascript/components/widgets/traffic/panel.jsx
```diff
@@ -69,25 +83,21 @@ const Traffic = () => (
...
+      const [firstColumn, secondColumn] = splitAt(3, trafficCams);
 
       return (
         <>
           <Row>
-            <Column>{CameraList(0, 3)}</Column>
-            <Column>{CameraList(3, 6)}</Column>
+            <Column>
+              {take(3, firstColumn).map(cam => (
+                <TrafficCam {...cam} />
```
djkirby: `key={/* ... */}` ?

app/javascript/components/widgets/traffic/panel.jsx
```diff
@@ -89,13 +90,13 @@ const Traffic = () => (
         <>
           <Row>
             <Column>
-              {take(3, firstColumn).map(cam => (
-                <TrafficCam {...cam} />
+              {take(3, firstColumn).map(({ id, ...cam }) => (
+                <TrafficCam key={id} {...cam} />
```
djkirby: squash w/ 1st commit?

cbarber: Ah ok. My intent here was the key change was a bugfix that was simplified with the preceding refactor. Maybe that's too sub-atomic. I'll have a look at the squashed result.

djkirby: Maybe I misunderstood the changes -- it looked like there was a `key` being applied originally https://github.com/mojotech/helios2/pull/161/files#diff-9a23cb58dd0b34ddf1471cd7d9aff51eL76

cbarber: Finally got back to this. `<div key={cam}>` was the issue since an object was being passed to key which resulted in the react warnings. I'll make that more clear.

PR #181, opened 2020-03-02
fix: handle state change with identical or missing currentWidgetId

cbarber:
The pi exhibited an issue where a current widget could become stuck and
never continue to the next widget. It seems to be caused by clicking on
the current widget in the tab. Added some logging in case there are
other edge cases.


joealba:
Do you know of any javascript log drains that have a free tier for an easier way to look at these messages?

chewnoill (approved):

app/javascript/components/widget-controller.jsx
```diff
@@ -139,12 +139,16 @@ export class WidgetController extends React.Component {
...
-      nextWidgetId: null,
-      prevWidgetId: null,
-      transitionTime: null,
-    });
+    this.setState(({ currentWidgetId }) =>
+      currentWidgetId !== id
+        ? {
+            currentWidgetId: id,
+            nextWidgetId: null,
+            prevWidgetId: null,
```
chewnoill, line 158: are these all calculated values?

cbarber:
I'm going need some time to remember what this is and if its still neccessary

chewnoill:
closed in favor of https://github.com/mojotech/helios2/pull/252

PR #203, opened 2021-03-26
refactor!: Replace Darksky with OpenWeather API

Aceba1:
Significant changes:
- `.env` requires `WEATHER_API` set to an [OpenWeather API key](https://home.openweathermap.org/api_keys)
- GraphQL schema has been updated to match new API
- Moon-Phase is no longer provided by weather API, it is (currently) calculated within the Location model


app/graphql/types/weather_type.rb
```diff
@@ -121,5 +120,7 @@ class Types::WeatherType < Types::WeatherObject
   field "offset", Int
   field "solarcycles", [Types::SolarcycleType]
 
+  field "moonPhase", Float, deprecation_reason: "Moved to Location"
```
aaddaamm, line 108: nit: This isn't a requirement but it would be helpful for the reviewer if this diff happened back at the original line. When reading the change, seeing them tight together is more helpful to the reader. In this case the change is small so it's not that big of a deal. Just something to think about for future commits/PRs.

aaddaamm: unless there is a good reason as to why this field definition moved from line 18 down to line 123 of course.

app/graphql/types/location_type.rb
```diff
@@ -15,4 +15,6 @@ class Types::LocationType < Types::BaseObject
   field "wifi_password", String
   field "bathroom_code", String
   field "traffic_cams", [Types::TrafficCamType]
+
```
aaddaamm, line 18: probably don't need this whitespace here.

app/graphql/types/icon_id_to_name_type.rb
```diff
@@ -0,0 +1,20 @@
+# Reference: https://openweathermap.org/weather-conditions#Weather-Condition-Codes-2
```
aaddaamm, line 1: This Commit is rather large, is there a way we could break it up?

app/graphql/types/weather_type.rb
Aceba1: The field definition was moved out of `WeatherDailyDetailType`, as the API will no longer be providing it within that scope. Refactoring the schema must also move where queries are being made to, therefore moving moonPhase to WeatherType is to additionally simplify such changes.

aaddaamm: ooooooooh i see now. wow that Diff didn't help me see that at all. Carry on

app/graphql/types/icon_id_to_name_type.rb
Aceba1: Sadly, there's no clear lines to split this commit down and have the program still functional, as all the changes are to transition an API that is depended on for it.

app/graphql/types/weather_type.rb
```diff
@@ -9,114 +9,99 @@ class Types::WeatherObject < Types::BaseObject
...
 
-class Types::WeatherMinutelyDataType < Types::WeatherObject
-  field "time", Types::UnixDateTimeType
-  field "precipIntensity", Float
-  field "precipIntensityError", Float
-  field "precipProbability", Float
-  field "precipType", String
+class Types::WeatherInfoDataType < Types::WeatherObject
+  def initialize(value, context, **kwargs, &block)
+    super(value[0], context, **kwargs, &block)
```
cbarber, line 30: Maybe add comment mentioned we converting the inbound `[{ ... }]` into `{ .... }`. Specifically mentioning that its an array of one object.

app/javascript/components/widgets/weather/hourly-temps.jsx
```diff
@@ -58,29 +58,27 @@ const RainIcon = styled.img`
...
-    pathOr([], ['hourly', 'data']),
+    pathOr([], ['hourly']),
   )(weather);
   return (
     <Wrapper>
-      {hourlyWeathers.map(({ time, temperature, precipProbability }, idx) => (
-        <Item key={time} index={idx}>
-          <Time>{parseHour(time)}</Time>
-          <Temp>{parseInt(temperature, 10)}°</Temp>
+      {hourlyWeathers.map(({ dt, temp, pop }, idx) => (
```
cbarber, line 75: I think `dt` and `pop` are too cryptic in the context of weather and we should alias them.

app/models/location.rb
```diff
@@ -16,10 +16,19 @@ def self.primary(city_name = ENV['PRIMARY_CITY_NAME'])
   def weather
     response = ::Clients::DarkskyClient.forecast(self).dup
     response['solarcycles'] = solarcycles.after_beginning_of_yesterday.to_a
+    response['moonPhase'] = moon_phase
     response.freeze
     response
   end
 
+  MOON_CYCLE_TIME = 2_551_442.8
```
geraldalewis, line 24: Super cool 😎

app/graphql/types/icon_id_to_name_type.rb
```diff
@@ -0,0 +1,20 @@
...
+
+class Types::IconIdToNameType < GraphQL::Schema::Scalar
+  # rubocop:disable Metrics/CyclomaticComplexity
```
geraldalewis, line 4: Not something to be concerned about for this PR, but this reminds me to evaluate the usefulness of the `CyclomaticComplexity` rule...

app/graphql/types/weather_type.rb
```diff
@@ -9,114 +9,100 @@ class Types::WeatherObject < Types::BaseObject
...
 end
 
-class Types::WeatherDailyDetailType < Types::WeatherObject
-  field "summary", String
-  field "icon", String
-  field "data", [Types::WeatherDailyDataType]
+class Types::WeatherDailyDataType < Types::WeatherObject
+  field "time", Types::UnixDateTimeType, hash_key: "dt"
+  field "weather", Types::WeatherInfoDataType
+  field :sunrise, Types::UnixDateTimeType
```
geraldalewis, line 43: Out of curiosity -- is there a reason for some of these fields to be `intern`ed and some expressed as `string`s?

geraldalewis (commented):
Wow -- ambitious PR! I didn't get a chance to really dig into the API update commit, but what I saw looked really good 👍

cbarber (approved):

app/graphql/types/weather_type.rb
Aceba1: Initially the schema was a mix of symbols and strings. There's no benefit or functional difference to strings over symbols, but it would've been out of the scope of this PR to change it all to one type

PR #205, opened 2021-03-29
chore: Resolve security vulnerability for minimist, update package for flaky test case

Aceba1:
- Updates jest to remove outdated transient dependency of minimist
- Remediates flaky "Daylight Savings" tests caused by outdated toolset packages


Aceba1:
Whoops, looks like it only updated yarn.lock, I'll get that fixed now

aaddaamm (approved):

PR #206, opened 2021-03-30
chore: Resolve node-fetch vulnerability

Aceba1:
- Bump eslint-plugin-graphql to 4.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "eslint-config-airbnb": "^17.1.0",
     "eslint-config-prettier": "^4.3.0",
     "eslint-import-resolver-webpack": "^0.11.1",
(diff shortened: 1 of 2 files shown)


aaddaamm:
are there any linting issues after this version bump?

Aceba1:
It functions just the same 👍

Only difference I can find is the change of a npm script here: https://github.com/apollographql/eslint-plugin-graphql/commit/faef1fd1e5c88f16f7b9ccf23da33ca3e138557b

aaddaamm:
so if we run the linter after this version bump we don't get any new linting warnings/errors?

Aceba1:
That's correct

aaddaamm (approved):

PR #207, opened 2021-03-30
chore: Bump styled-components to 4.0.0

Aceba1:
- Resolves (additional) transient dependency of vulnerable node-fetch version
- Updates usage of styled-components to match version 4.0.0

### DIFF
--- a/app/javascript/components/tab-bar.jsx
+++ b/app/javascript/components/tab-bar.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import styled, { keyframes } from 'styled-components';
+import styled, { keyframes, css } from 'styled-components';
 import { colors } from '../lib/theme';
 
 const width = 568;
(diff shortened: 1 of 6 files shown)


app/javascript/components/tab-bar.jsx
```diff
@@ -15,13 +15,17 @@ const progressBarFrames = keyframes`
   }
 `;
 
+const progressBarAnimation = props => css`
```
aaddaamm, line 18: why was this extracted from the ProgressBar definition?

Aceba1: With v4, styled-components no longer supports interpolating keyframe declarations in untagged strings. 

Should I instead embed the css tag within the ProgressBar object? ex:
```
const ProgressBar = styled.div`
  background-color: ${colors.white};
  height: ${height}px;
  width: ${width}px;
  ${props => css`
    animation: ${progressBarFrames} ${props.time}s linear 0s;
    animation-fill-mode: forwards;
    animation-play-state: ${props.isPaused ? `paused` : `playing`};
  `};
`;
```

https://styled-components.com/docs/api#css

aaddaamm: no i think it's ok the way it is. Thanks for the reference!

aaddaamm (approved):

PR #212, opened 2021-04-06
fix: Support styled-components v4 for moon-phase

Aceba1:
This addresses the moon not using styled components for PR #211, breaking the frontend when rendering

img

### DIFF
--- a/app/javascript/components/widgets/weather/moon-phase.jsx
+++ b/app/javascript/components/widgets/weather/moon-phase.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { colors } from '@lib/theme';
+import styled from 'styled-components';
 
 const getRearPath = (radius, phase, radiusModifier1) => {
   const diameter = radius * 2;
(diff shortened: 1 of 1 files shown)


app/javascript/components/widgets/weather/moon-phase.jsx
```diff
@@ -155,7 +155,7 @@ MoonPhase.propTypes = {
   radius: PropTypes.string.isRequired,
   phase: PropTypes.string.isRequired,
   arcPath: PropTypes.string.isRequired,
-  arcAnimation: PropTypes.string.isRequired,
+  arcAnimation: PropTypes.objectOf(PropTypes.any).isRequired,
```
aaddaamm, line 158: if this is indeed an object now, we should define its shape here.

Aceba1: Mistype there, it is an instance of the [Keyframes](https://styled-components.com/docs/api#keyframes) model now. The type is not exposed by the styled-components package, however. 

That line should really be `object` or `any`, but the linter advises strongly against such.

Any recommendations?

app/javascript/components/widgets/weather/moon-phase.jsx
```diff
@@ -155,7 +155,10 @@ MoonPhase.propTypes = {
   radius: PropTypes.string.isRequired,
   phase: PropTypes.string.isRequired,
   arcPath: PropTypes.string.isRequired,
-  arcAnimation: PropTypes.string.isRequired,
+  arcAnimation: PropTypes.shape({
```
Aceba1, line 158: Changed Prop-Types entry for arcAnimation to hold two properties unique to the Keyframes model

aaddaamm (approved):

PR #219, opened 2021-05-20
[fix] update docker node

chewnoill:
This gets the docker builds building at least. 

This is a way simpler setup then running things natively, install docker and run `docker-compose up`. Your web service and all dependent services should load and start with a single command.

The trade off is going to be with developer experience when you start making changes. If you have to rebuild the image every time you make a change its going to slow things down a lot.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM ruby:2.5.7
 ARG USER_ID=1000
 ARG GROUP_ID=1000
 
-RUN curl -sL https://deb.nodesource.com/setup_8.x | bash &&\
+RUN curl -sL https://deb.nodesource.com/setup_14.x | bash &&\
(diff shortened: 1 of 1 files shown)


cbarber:
Nice job!

Couple other issues with the developer expierence in docker:
* `debug` breaks down since we're still using foreman inside of a single container. We might be able to get this to work by splitting the foreman script into three separate container, and `docker attach` onto the stdin/tty of the rails container
* Entering the rails console was also tricky. Maybe another container that can be started separately?
* Running rails operations like migrate/generate/seed. Same as above.

cbarber (approved):

cbarber:
Sorry I wasn't clear. I meant there is broader work to do to make docker a viable dev env:
https://www.pivotaltracker.com/story/show/171465217

This PR is solid though :+1:

PR #223, opened 2021-05-21
feat: update README to include ubuntu redis install

bbazyk:
Adds instructions for installing redis, just for ubuntu linux

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,11 +34,36 @@ Prequisites:
 - `cp .env.local.sample .env` -- fill in env vars where applicable
 - Configure Redis:
 
+OSX:
   ```shell
   brew install redis
   redis-cli CONFIG SET dir /tmp/
   redis-cli CONFIG SET dbfilename temp.rdb
   redis-server &
+  redis-cli ping  # Should respond with PONG
+  rake db:setup
+  ```
+Ubuntu:
+- Update system
+  ```shell
+  sudo apt update
+  ```
+- Install redis server
+  ```shell
+  sudo apt install redis-server
+  ```
+- Update configuration
+  ```shell
+  sudo sh -c 'echo "supervised systemd" >> /etc/redis/redis.config'
+  ```
+- Restart running service
+  ```shell
+  sudo systemctl restart redis.service
+  ```
+- Configure and verify redis
+  ```shell
+  redis-cli CONFIG SET dir /tmp/
+  redis-cli CONFIG SET dbfilename temp.rdb
   redis-cli ping  # Should respond with PONG
   rake db:setup
   ```


README.md
```diff
@@ -42,6 +43,35 @@ Prequisites:
...
+  sudo apt install redis-server
+  ```
+- Update configuration
+  ```shell
+  sudo nano /etc/redis/redis.conf
+  ```
+- Append to configuration file:
+  ```shell
+  supervised systemd
+  ```
```
chewnoill, line 63: better to avoid starting any editors, could really trigger @cbarber :trollface: 

If you need to append a file do it like this: `echo 'supervised systemd' >> /etc/redis/redis.config`

Although all of these other steps seems super complicated also. If you install docker you can have a configured running redis server running on any OS by running:

```shell
docker-compose up redis
```

and you can still run the rest of the rails instance natively

cbarber: `sudoedit ... ` is generally preferred over `sudo some-editor ...`

You'd need to wrap the command with `sudo sh -c 'echo "supervised systemd" >> /etc/redis/redis.confg'` to execute the whole command as sudo. Otherwise the append write to /etc/redis/redis.config will fail with permission denied

The docker-compose.yml doesn't expose a port for redis. There's a ticket for that:
https://www.pivotaltracker.com/story/show/171465217

bbazyk: I've updated the suggested commands in the README, and removed the editor step. I tried running redis through docker, the server refused to connect. I think these separate instructions for Linux suffice until Craig's mentioned ticket is completed. @cbarber @chewnoill

README.md
```diff
@@ -42,6 +43,30 @@ Prequisites:
...
+  ```shell
+  sudo apt update
+  ```
+- Install redis server
+  ```shell
+  sudo apt install redis-server
+  ```
+- Update configuration
+  ```shell
+  sudo sh -c 'echo "supervised systemd" >> /etc/redis/redis.config'
```
fallynlogan, line 57: What is this line `sudo sh -c 'echo "supervised systemd" >> /etc/redis/redis.config'` doing/why would someone need to run this?

README.md
```diff
@@ -42,6 +43,30 @@ Prequisites:
...
+- Restart running service
+  ```shell
+  sudo systemctl restart redis.service
+  ```
+- Configure and verify redis
+  ```shell
+  redis-cli CONFIG SET dir /tmp/
+  redis-cli CONFIG SET dbfilename temp.rdb
+  redis-cli ping  # Should respond with PONG
+  rake db:setup
```
tmsavage, line 68: Do you need to have the command `redis-server` included in the configuration for Ubuntu, similarly to the instructions in OSX?

tucker-weed: I second this question - I believe you would still need to include running `redis-server` in the instructions?

sandipa12: what is the purpose of running redis-cli pong ? Is it to check whether the server is alive?

chewnoill (approved):
🚀

README.md
jennyckaplan: This command appends the line `supervised systemd` to the redis configuration file (`/etc/redis/redis.config`), which creates a [systemd](https://en.wikipedia.org/wiki/Systemd) service that runs the redis server instance. It allows you to start, stop, check the status of, restart the redis-server using `systemctl` commands like `systemctl start redis-server`.

README.md
jennyckaplan: After we update the redis configuration file to specify using `systemd` to manage redis, we run the command:
`sudo systemctl restart redis.service`, which restarts the redis systemd service, which starts running the redis-server. So for these instructions, we don't need an additional command to run `redis-server` in the background in the same manner as the OSX instructions, since `systemd` is managing redis and since we start the redis server with that `systemctl` command. And @sandipa12, yes the pong is just to ensure that the server is running.

tucker-weed (approved):
Merge the latest changes of master and you should be good.

cbarber:
> Merge the latest changes of master and you should be good.

Nit: rebase'ing is generally preferred over merging.

cbarber (approved):

PR #224, opened 2021-05-24
feat: runtime correction of whitespaces in ENV via proxy AppEnv

tucker-weed:
Strips whitespace off of rails environment variables before the rails app initializes.

### DIFF
--- a/app/lib/app_env.rb
+++ b/app/lib/app_env.rb
@@ -1,0 +1,5 @@
+class AppEnv
+  def self.[](key)
+    ENV.key?(key) && ENV[key].strip
+  end
+end
--- a/app/controllers/web_hooks/github_controller.rb
+++ b/app/controllers/web_hooks/github_controller.rb
@@ -33,6 +33,6 @@ class WebHooks::GithubController < ApplicationController
   end
 
   private def webhook_secret(_payload)
-    ENV['GITHUB_WEBHOOK_SECRET']
+    AppEnv['GITHUB_WEBHOOK_SECRET']
   end
 end
--- a/app/controllers/web_hooks/slack_auth_controller.rb
+++ b/app/controllers/web_hooks/slack_auth_controller.rb
@@ -3,10 +3,10 @@ require 'slack-ruby-client'
 class WebHooks::SlackAuthController < ApplicationController
   layout false
   SLACK_CONFIG = {
-    slack_client_id: ENV['SLACK_CLIENT_ID'],
-    slack_api_secret: ENV['SLACK_API_SECRET'],
-    slack_redirect_uri: ENV['SLACK_REDIRECT_URI'],
-    slack_verification_token: ENV['SLACK_VERIFICATION_TOKEN']
+    slack_client_id: AppEnv['SLACK_CLIENT_ID'],
+    slack_api_secret: AppEnv['SLACK_API_SECRET'],
+    slack_redirect_uri: AppEnv['SLACK_REDIRECT_URI'],
+    slack_verification_token: AppEnv['SLACK_VERIFICATION_TOKEN']
   }.freeze
 
   missing_params = SLACK_CONFIG.select { |_key, value| value.nil? }
--- a/app/controllers/web_hooks/slack_controller.rb
+++ b/app/controllers/web_hooks/slack_controller.rb
@@ -2,7 +2,8 @@ class WebHooks::SlackController < ApplicationController
   protect_from_forgery with: :null_session
 
   def create
-    halt 403, "Invalid Slack verification token: #{params[:token]}" if ENV['SLACK_VERIFICATION_TOKEN'] != params[:token]
+    is_invalid_token = AppEnv['SLACK_VERIFICATION_TOKEN'] != params[:token]
+    halt 403, "Invalid Slack verification token: #{params[:token]}" if is_invalid_token
     render plain: build_create_response
   end
 
@@ -72,7 +73,7 @@ class WebHooks::SlackController < ApplicationController
     require 'uri'
     require 'json'
 
-    uri = URI.parse(ENV['SLACK_WEBHOOK_URL'])
+    uri = URI.parse(AppEnv['SLACK_WEBHOOK_URL'])
 
     header = { 'Content-Type': 'application/json' }
     text = { 'text': "<@#{user}> #{message}", 'channel': channel.to_s }
--- a/app/lib/clients/twitter_client.rb
+++ b/app/lib/clients/twitter_client.rb
@@ -2,10 +2,10 @@ require 'twitter'
 
 class Clients::TwitterClient
   @client = Twitter::REST::Client.new do |config|
-    config.consumer_key = ENV['TWITTER_CONSUMER_KEY']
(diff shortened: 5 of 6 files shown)


config/environment.rb
```diff
@@ -1,3 +1,6 @@
+# Strip whitespace from environment values for proper API to app handshake.
+ENV.each { |k, v| ENV[k] = v.strip }
```
cbarber, line 2: The risk here is we'll mutate all environment variables rather than just the ones our application cares about. This could have a surprise side effect if some rails variable needs leading or trailing whitespace.

What do you think of an AppEnv singleton class that removes the white space for us?
`AppEnv['GITHUB_WEBHOOK_SECRET']` returns `ENV['GITHUB_WEBHOOK_SECRET'].strip`

tucker-weed: I like that idea - looks like it allows for more granular control over specific environment variables (beyond just .strip as well). Will work on that now.

config/environment.rb
```diff
@@ -1,3 +1,19 @@
...
+end
+
+AppEnv['WEATHER_API_KEY']
+AppEnv['SLACK_CLIENT_ID']
+AppEnv['SLACK_API_SECRET']
+AppEnv['SLACK_VERIFICATION_TOKEN']
+AppEnv['TWITTER_CONSUMER_KEY']
+AppEnv['TWITTER_CONSUMER_SECRET']
+AppEnv['TWITTER_ACCESS_TOKEN']
+AppEnv['TWITTER_ACCESS_SECRET']
```
cbarber, line 15: Sorry I wasn't clear. I was thinking we'd avoid mutating the ENV variables directly:
```
class AppEnv
  def self.[](key)
    ENV[key].strip
  end
end
```
and then find instances of `ENV[...]` in the `app` directory to replace them with `AppEnv[...]`

I think we should also put AppEnv in its own file in `lib/app_env.rb`

config/environment.rb
```diff
@@ -1,3 +1,19 @@
+# Strip whitespace from environment values for proper API to app handshake.
+class AppEnv
+  def self.[](key)
+    ENV.key?(key) && ENV[key] = ENV[key].strip
```
fallynlogan, line 4: For this part `ENV.key?(key)` are we checking that the key is not null? Also, why would we want to include `&&` instead of just separating these onto different lines?

config/environment.rb
fallynlogan: @cbarber is there an example in this project where an environment variable has trailing whitespaces that are needed? If so, why is that the case?

tmsavage:
@tucker-weed if you need help 'condensing' your 5 commits into 1, I can help you with that.

config/environment.rb
tmsavage: What instances would leading spaces be necessary for some rails variables? If we know these instances, would it be less costly to overwrite specific variables in the ENV file once so the application doesn't have to keep parsing the white spaces every time we start the app?

sandipa12:
what does  ENV.key?(key) evaluate?

config/environment.rb
tucker-weed: .strip is O(n) where n is the length of the string. I think that because the length of the strings that need to be stripped are not long (around 25) and not variable in length, it is probably not a bottleneck in the runtime of the application. What are everyone's thoughts on this?

config/environment.rb
tucker-weed: Yes, it is a check that a key is not null (kind of functions as a defensive check at this point). The && is included so that if `ENV.key?(key)` is equal to null, the second part of the statement, `ENV[key] = ENV[key].strip` will not be evaluated. If this is not the proper way to do this, let me know.

tucker-weed:
> what does ENV.key?(key) evaluate?

This evaluates whether a key exists in the ENV hash.

config/environment.rb
joealba: Looks like `n` is the length of the leading and trailing whitespace (https://ruby-doc.org/core-3.0.1/String.html#method-i-strip / https://bugs.ruby-lang.org/issues/12788), but either way isn't a problem here.  If you're concerned about `String#strip` performance, you're already using the wrong language.  ;)

app/lib/app_env.rb
```diff
@@ -0,0 +1,6 @@
+# A class which overrides the default indexing behavior of ENV.
```
joealba, line 1: I'm not sure the comment adds much value here due to the simplicity of the code.

joealba:
I recommend correcting the commit string before merging.  Whitespace is being stripped at runtime during each access to a value now, not during the init phase.  Otherwise, 👍 💯 !

config/environment.rb
chewnoill: I don't think there are examples in **any** project. Wanting to keep leading/trailing whitespace is most likely a non-issue. Just impolite, maybe, to edit env variables for the whole system. Adding an abstraction on top of ENV to strip whitespace adds extra complexity to a solution that is generally just "update your env variables to the correct value". 

The real problem here, I think, is that rails fails to load, or fails to provide some critical function, because env variables are not set correctly and doesn't provide helpful information to the operator that an action is required. Stripping whitespace that was accidentally copy and pasted in solves only a symptom of this issue.

config/environment.rb
cbarber: I don't think we should mutate the existing ENV hash since it would be difficult to discover and troubleshoot *if* it is the ever the wrong operation.

I would instead opt for adding a new class:
```
class AppEnv
  def self.[](key)
    ENV[key].strip
  end
end
```
and changing existing usages of `ENV` found in our `app/` folder. An example:
`app/controllers/web_hooks/github_controller.rb`
```
 35   private def webhook_secret(_payload)
 36     ENV['GITHUB_WEBHOOK_SECRET']
 37   end
```
would become
```
 35   private def webhook_secret(_payload)
 36     AppEnv['GITHUB_WEBHOOK_SECRET']
 37   end
 ```

Why this change instead of the original?
* Set up other developers to do the right thing. Some time in the future when another developer wants to add a new ENV variable for some new service, they're probably going to look in the application for other services that do the same thing and copy it. With `AppEnv` used in our app for all of our env vars, they will likely copy that. They might miss the allow/disallow list of env variables to mutate in ENV.
* Discoverability. If someone needs to find out why whitespace is being stripped out, `AppEnv` in their call provides a clear callstack to follow. Trying to find a mutation in `config/environment.rb` is harder.

I think these are good coding practices to follow, even if the likelihood of us or rails ever needing whitespace in our ENV vars is pretty low.

app/lib/app_env.rb
tucker-weed: Removed the comment - I completely agree that it is an unnecessary annotation of the class.

config/environment.rb
tucker-weed: This code snippet is slightly outdated though - `ENV[key] = ENV[key].strip` is actually just `ENV[key].strip` so that the stripped whitespace ENV value is returned from the method call rather than mutating ENV permanently.

chewnoill (approved):
🚀

PR #225, opened 2021-05-24
Added San Diego to seeds.rb file

tmsavage:
Added San Diego compatibility for Helios2.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -31,6 +31,14 @@ Location.find_or_initialize_by(
   r.time_zone = 'America/New_York'
 end.save!
 
+Location.find_or_initialize_by(
+  city_name: 'SanDiego'
+) do |r|
+  r.latitude = 32.7157
+  r.longitude = -117.1611
+  r.time_zone = 'America/Los_Angeles'
+end.save!
+
 primary_location = Location.find_by(city_name: ENV['PRIMARY_CITY_NAME'])
 Widget.find_or_initialize_by(
   name: "Guests"


db/seeds.rb
```diff
@@ -31,6 +31,14 @@
   r.time_zone = 'America/New_York'
 end.save!
 
+Location.find_or_initialize_by(
+  city_name: 'SD'
```
cbarber, line 35: We have a change coming where we will access these locations from the URL. What do you think of `SanDiego` for the name to match Boulder and Providence?

tmsavage: I wrote 'SD' to match the abbreviation format similar to 'NYC' and 'DC'. If it would be more clear and consistent with 'Boulder' and 'Providence' to change 'SD' to `SanDiego`, then I agree with the change.

db/seeds.rb
```diff
@@ -31,6 +31,14 @@
   r.time_zone = 'America/New_York'
 end.save!
 
+Location.find_or_initialize_by(
+  city_name: 'SanDiego'
+) do |r|
+  r.latitude = 32.7157
+  r.longitude = -117.1611
+  r.time_zone = 'America/Los_Angeles'
```
fallynlogan, line 39: What is the '|r|' and why do we need it?

tmsavage: So, from reading up on this, it's a form of saving values to each block of code. In this case, we're creating a new Location object for San Diego with multiple attributes. `do |r|`goes through each variable and saves their values to the new object. Mentors, is my statement correct? I may be on the right track but I'm not 100% positive on my response.

cbarber (approved):
:+1:

chewnoill (approved):
🚀

PR #226, opened 2021-05-24
fix: add missing environment variables to sample env file

bbazyk:
This PR adds 4 missing used environment variables to the .env file.


.env.local.sample
```diff
@@ -8,3 +8,8 @@ SLACK_API_SECRET=
 SLACK_REDIRECT_URI=https://mojotech-helios2.herokuapp.com/web_hooks/create_slack_auth
 SLACK_VERIFICATION_TOKEN=
 SLACK_WEBHOOK_URL=
+GITHUB_WEBHOOK_SECRET=
```
fallynlogan, line 11: Do you know where we can find this on GitHub or how we add this to the project?

tmsavage (approved):
Since we use the command 'cp .env.local.sample .env', which copies these lines from the .env.local.sample file to a new .env file, then this is a good update.

.env.local.sample
```diff
@@ -8,3 +8,8 @@ SLACK_API_SECRET=
...
+TWITTER_CONSUMER_KEY=
+TWITTER_CONSUMER_SECRET=
+TWITTER_ACCESS_TOKEN=
+TWITTER_ACCESS_SECRET=
```
tucker-weed, line 15: I believe that this change you made fixes an issue I was running into earlier where I tried to strip whitespace from ENV variables that didn't exist. I may drop the check ENV.key?(key) from my PR if you merge this PR, so I will be following this (or I may not drop it because the check doesn't hinder proper functioning of helios, not sure yet).

.env.local.sample
sandipa12: Is the term 'WEBHOOK' similar to http callback?

sandipa12:
Is the term ' webhook' similar to http callback?

.env.local.sample
cbarber: In github project settings (that you own or have admin access):
Screenshot from 2021-05-25 07-46-25

cbarber:
> Is the term ' webhook' similar to http callback?

Yes. You provide a webhook to trigger some functionality in your application. Other services can use that webhook to make web requests to your servers with events/data.

chewnoill (approved):
🚀

cbarber (approved):

PR #228, opened 2021-05-25
 fix : broken traffic cam

sandipa12:
The url for Tobey street was broken so it has been replaced with the new url.  https://www.pivotaltracker.com/story/show/178115637

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -138,8 +138,12 @@ TrafficCam.find_or_initialize_by(
 end.save!
 
 TrafficCam.find_or_initialize_by(
-  title: "Tobey Street"
+  title: "Dean Street"
 ) do |r|
-  r.url = "http://www.dot.ri.gov/img/travel/camimages/6-2%20Rt%206%20E%20%20%2010%20N%20@%20Tobey%20St.jpg"
+  r.url = "https://www.dot.ri.gov/img/travel/camimages/6-1%20Rt%206%20E%20%20%2010%20N%20@%20Dean%20St.jpg"
   r.location = providence
 end.save!
+
+TrafficCam.where(
+  title: "Tobey Street"
+).destroy_all


db/seeds.rb
```diff
@@ -127,6 +127,6 @@
 TrafficCam.find_or_initialize_by(
   title: "Tobey Street"
```
cbarber, line 128: I think the title should probably change as well.

tucker-weed:
A part of the URL changed from Tobey to Dean - is the new URL for a different street then?

cbarber (commented):
This is a great start. I think we should probably add a migration to manipulate existing data in databases:
https://guides.rubyonrails.org/v5.2/active_record_migrations.html

sandipa12:
@tucker-weed  Yes, the new url is for a different state (Dean Street)

db/seeds.rb
```diff
@@ -125,8 +125,8 @@
 end.save!
 
 TrafficCam.find_or_initialize_by(
-  title: "Tobey Street"
+  title: "Dean Street"
 ) do |r|
-  r.url = "http://www.dot.ri.gov/img/travel/camimages/6-2%20Rt%206%20E%20%20%2010%20N%20@%20Tobey%20St.jpg"
+  r.url = "https://www.dot.ri.gov/img/travel/camimages/6-1%20Rt%206%20E%20%20%2010%20N%20@%20Dean%20St.jpg"
```
tmsavage, line 130: What would be the purpose of changing http to https for the link?

sandipa12: It does not really matter. So far I know, https is more secure than http but either of them works to open the link.

chewnoill:
We can squash these commits together, but other wise LGTM

db/seeds.rb
```diff
@@ -133,8 +133,8 @@
...
```
fallynlogan, line 138: Would the logic & where you retrieved the street cam info be similar for adding traffic cams for other locations?

sandipa12: @fallynlogan Yeah, it  should be the same.

db/seeds.rb
bencates: You should basically always prefer `https` to `http` when you have the choice, even if you don't directly care about any of the security features that TLS gives you.

Pages served to the user via HTTP will happily download HTTPS assets. However, pages served over HTTPS will refuse to fetch HTTP assets as doing so could compromise the security guarantees HTTPS makes for the page as a whole.

cbarber (changes requested):
I think we should use `rake db:seed` to update production until we have an admin interface to change these values. In this case, we'll need our seeds.rb to delete the old `Tobey Street` camera in addition to adding a new `Dead Street`.

Any objections to adding?
`TrafficCam.delete_by(title: "Tobey Street")`

chewnoill (dismissed):
Adding a delete seems like a good idea for the old row. As long as it can be run multiple times without breaking anything. I'm also okay with dropping the database and reseeding if we run into any problems.

chewnoill (approved):

cbarber (approved):
Looks like the API is `destroy_all` or `delete_all`. `destroy_all` triggers ActiveRecord callbacks for the destroy event, and `delete_all` sends the SQL directly without any callbacks. Nice work. :+1:

PR #229, opened 2021-05-25
feat: added placeholder icons for wifi and password

bbazyk:
Pivotal Tracker wifi icons : id 166386395
This PR adds placeholders for the icons. To be replaced in a later PR with updated designs @fallynlogan . 

### Before
Screenshot from 2021-05-25 14-36-48

### After
Screenshot from 2021-05-25 14-36-08


app/assets/images/wifi_icon.svg
```diff
@@ -0,0 +1,85 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
```
chewnoill, line 5: svgs should be minimized before we add them to the project. Check out a tool like this one to turn it into a regular jsx component; I think it minimizes also

https://svg2jsx.com/

app/javascript/components/wifi.jsx
```diff
@@ -5,6 +5,8 @@ import gql from 'graphql-tag';
 import { WhiteSubTitle, WhiteTitleLarge } from '@components/typography';
 import { Row } from '@components/row';
 import { LoadingMessage, ErrorMessage } from '@messages/default-messages';
+import WifiIcon from '@assets/images/wifi_icon.svg';
```
jennyckaplan, line 8: maybe we name this `wifi_name.svg` and `wifi_password.svg` since that's the name of the text for these icons?

app/javascript/components/wifi.jsx
```diff
@@ -44,8 +46,14 @@ export const Wifi = () => (
           <Row>
             <Column>
               <WhiteSubTitle>WIFI</WhiteSubTitle>
-              <WhiteTitleLarge>{wifiName}</WhiteTitleLarge>
-              <WhiteTitleLarge>{wifiPassword}</WhiteTitleLarge>
+              <Row>
+                <img src={WifiIcon} alt="wifi" width={35} height={49.5} />
```
jennyckaplan, line 50: Let's make the alt text here "wifi-name" and "wifi-password"

app/javascript/components/wifi.jsx
jennyckaplan: @fallynlogan, once you're finished with the updated designs, all you need to do is replace these svgs with your new files (and possibly update the width and height to make sure they look good).

app/assets/images/wifi_icon.svg
```diff
@@ -0,0 +1,85 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
fallynlogan, line 1: Why would we want to use xml here?

app/javascript/components/wifi.jsx
fallynlogan: @jennyckaplan sounds good I will work on that today!

app/assets/images/wifi_icon.svg
bencates: SVG is a subset of XML. All SVG images are stored as XML documents and it's not uncommon to hand-edit the SVG document and/or render it dynamically via JSX just like we do with HTML documents.

This particular line is roughly equivalent to the `<!doctype html>` line that starts HTML documents. Version 1.0 is the default value, so in XML documents it can be (and usually is) omitted. Inkscape apparently includes it, though. ¯\\_(ツ)_/¯

cbarber (approved):

PR #230, opened 2021-05-25
Added new Javascript component for city name

tmsavage:
### Files Modified
- app/javascript/components/side-panel.jsx
- app/javascript/components/city-name.jsx
- db/seeds.rb

### Primary Update
- This pull request is to display the city name on the application. This component updates each time .env `PRIMARY_CITY_NAME` is changed.

### Secondary Update 
- Replaced 'SanDiego' with 'San Diego'

### Example
Screen Shot 2021-05-25 at 12 36 52 PM

### DIFF
--- a/app/javascript/components/city-name.jsx
+++ b/app/javascript/components/city-name.jsx
@@ -1,0 +1,32 @@
+import React from 'react';
+import { Query } from 'react-apollo';
+import gql from 'graphql-tag';
+import { WhiteTitleLarge } from '@components/typography';
+import { LoadingMessage, ErrorMessage } from '@messages/default-messages';
+
+const getCityName = gql`
+  {
+    primaryLocation {
+      cityName
+    }
+  }
+`;
+
+export const City = () => (
+  <Query query={getCityName}>
+    {({ loading, error, data }) => {
+      if (loading) {
+        return <LoadingMessage />;
+      }
+
+      if (error) {
+        return <ErrorMessage />;
+      }
+
+      const { cityName } = data.primaryLocation;
+      return cityName && <WhiteTitleLarge>{cityName}</WhiteTitleLarge>;
+    }}
+  </Query>
+);
+
+export default City;
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -32,7 +32,7 @@ Location.find_or_initialize_by(
 end.save!
 
 Location.find_or_initialize_by(
-  city_name: 'SanDiego'
+  city_name: 'San Diego'
 ) do |r|
   r.latitude = 32.7157
   r.longitude = -117.1611
--- a/app/javascript/components/side-panel.jsx
+++ b/app/javascript/components/side-panel.jsx
@@ -4,6 +4,7 @@ import styled from 'styled-components';
 import { spacing, colors, sidePanelWidth } from '@lib/theme';
 import Wifi from '@components/wifi';
 import Bathroom from '@components/bathroom';
+import CityName from '@components/city-name';
 import { Row } from '@components/row';
 import TopCorner from '@components/top-corner-controller';
 import Widgets from '@components/widgets';
@@ -34,6 +35,7 @@ export const SidePanel = ({
 }) => (
   <Wrapper>
     <TopCorner showWeather={showWeather} />
+    <CityName />
     <Row>
       <Wifi />
       <Bathroom />


app/javascript/components/city-name.jsx
```diff
@@ -0,0 +1,54 @@
...
+  flex-direction: column;
+  margin-top: 100px;
+`;
+// const Column = styled.div`
+//     font-size: ${fontSizes.xlarge};
+//     font-family: ${fonts.extended};
+//     font-weight: ${weights.regular};
+//     position: relative;
+//     top: 8px;
+// `;
```
cbarber, line 27: Do we need these commented lines? Lets remove them if not.

app/javascript/components/city-name.jsx
```diff
@@ -0,0 +1,54 @@
...
+
+export const City = () => (
+    <Query query={getCityName}>
+        {({ loading, error, data }) => {
+            if (loading) {
+            return <LoadingMessage />;
```
cbarber, line 33: Prettier should clean up these indentation issues:
https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode

db/seeds.rb
```diff
@@ -32,7 +32,7 @@
 end.save!
 
 Location.find_or_initialize_by(
-  city_name: 'SanDiego'
+  city_name: 'San Diego'
```
cbarber, line 35: This probably should be its own commit fixing the data.

app/javascript/components/city-name.jsx
```diff
@@ -0,0 +1,54 @@
...
+
+            const { cityName } = data.primaryLocation;
+            if (!cityName) {
+            return null;
+            }
+
+            return (
+            //<Column>
+                <WhiteTitleLarge>{cityName}</WhiteTitleLarge>
+            //</Column>
```
cbarber, line 48: And these comments

app/javascript/components/city-name.jsx
```diff
@@ -0,0 +1,54 @@
...
+// const Column = styled.div`
+//     font-size: ${fontSizes.xlarge};
+//     font-family: ${fonts.extended};
+//     font-weight: ${weights.regular};
+//     position: relative;
+//     top: 8px;
+// `;
+
+export const City = () => (
+    <Query query={getCityName}>
```
cbarber, line 30: Ah! I forgot side panel wasn't using the newer graphql fragment pattern. Out of scope here so I'll make a new ticket.

cbarber (commented):
That's a pretty nice addition. Lets rope in @jssee and @fallynlogan for any feedback on position/margin.

I think you can clear up most of these issues with prettier.

app/javascript/components/city-name.jsx
tmsavage: No we do not. I can remove them.

db/seeds.rb
tmsavage: A new commit within this pull request or a commit for a brand new PR?

chewnoill (approved):
🚀

app/javascript/components/city-name.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { Query } from 'react-apollo';
+import gql from 'graphql-tag';
+import { WhiteTitleLarge } from '@components/typography';
+import { LoadingMessage, ErrorMessage } from '@messages/default-messages';
+
+const getCityName = gql`
+  {
+    primaryLocation {
```
fallynlogan, line 9: How is this being set? (ie how would I be able to change it to Boulder)

fallynlogan (commented):
I think it looks good! Maybe increasing the font size, using a bold font, and evening out the spacing between the time and wifi sections could look good, or displaying it above the hourly forecast on the left side of the page could look nice. @jssee what do you think about those suggestions?

app/javascript/components/city-name.jsx
tmsavage: In the .env file, you can change the value set for `PRIMARY_CITY_NAME` based on the available cities in the data base. graphQL queries the database to get the set city name.

sandipa12: Will I be able to display two cities at the same time? Can I set two different values for PRIMARY_CITY_NAME in the .env file if I want to do so?

tmsavage: No this only works for one city.

bencates: ...so far :wink:

app/javascript/components/city-name.jsx
```diff
@@ -0,0 +1,36 @@
...
+      cityName
+    }
+  }
+`;
+
+export const City = () => (
+  <Query query={getCityName}>
+    {({ loading, error, data }) => {
```
bencates, line 17:
```suggestion
    {({ loading, error, data: { primaryLocation: { cityName } }) => {
```

Nitpick: Since the only thing we're using from `data` is the city name, I would consider destructuring to it directly over the assignment on line 26.

app/javascript/components/city-name.jsx
```diff
@@ -0,0 +1,36 @@
...
+      if (error) {
+        return <ErrorMessage />;
+      }
+
+      const { cityName } = data.primaryLocation;
+      if (!cityName) {
+        return null;
+      }
+
+      return <WhiteTitleLarge>{cityName}</WhiteTitleLarge>;
```
bencates, line 31:
```suggestion
      return (cityName && <WhiteTitleLarge>{cityName}</WhiteTitleLarge>);
```

Nitpick: There is nothing wrong with the early return as written. I do, however, see `&&` and `||` used heavily for control flow in JSX. In my opinion the one-liner version is slightly more idiomatic.

bencates (approved):
Looks great from an engineering perspective. Nice work!

I added a couple of opinionated nitpick suggestions, but they're just that: nitpicks. You have every right to ignore them and it won't hurt my feelings one bit.

app/javascript/components/city-name.jsx
tmsavage: Interesting, what do you mean by destructing in this case? would I have to change the variable name on line 31?

app/javascript/components/city-name.jsx
tmsavage: Ah yes, makes perfect sense! I agree with you, its much neater too.

app/javascript/components/city-name.jsx
bencates: Okay, so the `({ loading, error, data })` syntax is called "destructuring". So is the `{ cityName }` in `const { cityName } = data.primaryLocation`. It's a relatively new syntax in JS, the old equivalent would look like

tmsavage: Okay cool. Yeah its not a syntax I am familiar with. What would the old equivalent look like?

tmsavage: Hm, I seem to be getting type errors where the "right side of assignment cannot be destructured." Even if I change the line `return cityName && <WhiteTitleLarge>{cityName}</WhiteTitleLarge>;` to `return data && <WhiteTitleLarge>{data}</WhiteTitleLarge>;`, I get the same issue.

jssee:
Code looks good to me 👍 

One suggestion for the future; consider adding screenshots to the pull request (here on github) when it involves visual changes so reviewing the results are easier for everyone.

PR #231, opened 2021-05-25
 fix: add unique key for twitter images

fallynlogan:
**Ticket Description: **
A console error is displayed when the twitter widget is loaded:
Warning: Each child in a list should have a unique "key" prop.

**Changes: **
Add index & unique key (image path + index) to image wrapper component.

### DIFF
--- a/app/javascript/components/widgets/twitter/panel.jsx
+++ b/app/javascript/components/widgets/twitter/panel.jsx
@@ -19,7 +19,10 @@ const getMojoTweets = gql`
         retweetCount
       }
       media {
-        images
+        images {
+          id
+          mediaUrl
+        }
         link
       }
       user {
@@ -87,7 +90,12 @@ Twitter.propTypes = {
         retweetCount: PropTypes.number.isRequired,
       }).isRequired,
       media: PropTypes.shape({
-        images: PropTypes.arrayOf(PropTypes.string),
+        images: PropTypes.arrayOf(
+          PropTypes.shape({
+            id: PropTypes.number.isRequired,
+            mediaUrl: PropTypes.string.isRequired,
+          }),
+        ),
         link: PropTypes.string,
       }),
       user: PropTypes.shape({
--- a/app/javascript/components/widgets/twitter/tweet_content.jsx
+++ b/app/javascript/components/widgets/twitter/tweet_content.jsx
@@ -89,8 +89,8 @@ const imageLayout = images => {
   return images.map(image => {
     const isMain = image === images[0] && images.length === 3;
     return (
-      <ImageWrapper main={isMain}>
-        <TwitterImage isMain={isMain} src={image} alt="twitterMedia" />
+      <ImageWrapper key={image.id} main={isMain}>
+        <TwitterImage isMain={isMain} src={image.mediaUrl} alt="twitterMedia" />
       </ImageWrapper>
     );
   });
@@ -117,6 +117,11 @@ TweetMedia.defaultProps = {
 };
 
 TweetMedia.propTypes = {
-  images: PropTypes.arrayOf(PropTypes.string),
+  images: PropTypes.arrayOf(
+    PropTypes.shape({
+      id: PropTypes.number.isRequired,
+      media_url: PropTypes.string.isRequired,
+    }),
+  ),
   linkUrl: PropTypes.string,
 };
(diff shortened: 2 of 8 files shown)


app/javascript/components/widgets/twitter/tweet_content.jsx
```diff
@@ -86,10 +86,10 @@ TweetText.propTypes = {
 };
 
 const imageLayout = images => {
-  return images.map(image => {
+  return images.map((image, index) => {
```
tmsavage, line 89: What does adding the index, or unique key, do in this situation to eliminate the error?

app/javascript/components/widgets/twitter/tweet_content.jsx
```diff
@@ -86,10 +86,10 @@ TweetText.propTypes = {
...
     const isMain = image === images[0] && images.length === 3;
     return (
-      <ImageWrapper main={isMain}>
+      <ImageWrapper key={`${image}-${index}`} main={isMain}>
```
chewnoill, line 92: this is failing a linter check
> /home/circleci/repo/app/javascript/components/widgets/twitter/tweet_content.jsx
>  92:26  error  Do not use Array index in keys  react/no-array-index-key

Im okay if you disable this check, I don't think its valuable.

or `key={image}` would probably also work fine

fallynlogan: Okay, how would I disable the check?

app/javascript/components/widgets/twitter/tweet_content.jsx
bbazyk: https://reactjs.org/docs/lists-and-keys.html - error 404 on the page this link should lead to

bbazyk: nevermind, it just works now... idk

app/javascript/components/widgets/twitter/tweet_content.jsx
bencates: I wouldn't disable the check, it's there for a reason.

React uses the `key` to be able to tell which child is which in the event that a list is reordered or has an item inserted or removed. That allows it to reorder the existing children in the DOM rather than removing them all and recreating them in the correct order. However, changing the list will also change the indices of the items in the list, which will in turn change their keys. That means React won't be able to pair the old and new keys and will still have to drop and recreate the child.

A key with an array index in it is logically equivalent to omitting the key. If you're really okay omitting the key then silence _that_ error, otherwise pick a key that doesn't include the index.

schema.graphql
```diff
@@ -152,7 +152,7 @@ type TweetInteraction {
 }
 
 type TweetMedia {
-  images: [String!]
+  images: [Shape!]
```
bencates, line 155: What is a `Shape` and where is it's `type` defined? I don't see it in the schema and, naively, it seems like a surprising choice of model for an asset URL.

fallynlogan: I thought that was the type for an array or nested data, I tried that to get rid of the linter error but it did not so I was probably misunderstanding that. Trying out more fixes now

cbarber: FYI - `schema.graphql` and `schema.json` are generated files from `rake graphql:schema:dump`. This would be a good thing to include in the README.md in a separate PR: https://www.pivotaltracker.com/story/show/178305541

cbarber: To elaborate a little more, these files represent the graphql schema that is defined in the rails server. The `schema.json` file is used by eslint-graphql to check the graphql queries in the javascript files.

app/javascript/components/widgets/twitter/panel.jsx
```diff
@@ -87,7 +90,12 @@ Twitter.propTypes = {
         retweetCount: PropTypes.number.isRequired,
       }).isRequired,
       media: PropTypes.shape({
-        images: PropTypes.arrayOf(PropTypes.string),
+        images: PropTypes.arrayOf(
+          PropTypes.shape({
+            id: PropTypes.number.isRequired,
+            mediaUrl: PropTypes.string.isRequired,
+          }),
+        ),
```
tucker-weed, line 98: Does PropTypes.shape enforce what each element of the array is supposed to look like?

cbarber (approved):
Nice work :+1:

chewnoill (approved):
🚀

PR #232, opened 2021-05-26
Fix: add icons

fallynlogan:
Fixup icons and spacing for placeholder icons for ticket id 166386395

**Before: **
119551710-eec37480-bd67-11eb-8cf0-c5b8fcd2694d

**After**
Screen Shot 2021-05-26 at 2 59 03 PM


cbarber (commented):
I think `fixup styling` could mention what you're fixing e.g. `fix: add margins to wifi icons`

A before/after screenshot would be useful in the PR description. You can just click/drag files into a github description text field.

chewnoill (approved):
🚀

PR #234, opened 2021-05-27
feat: support displaying location based on the url

tucker-weed:
**Description**

1. Adds support for displaying different location _weather_ data based on /<city_name> url query option. Primary Location is still utilized in some components on the frontend for data unrelated to _weather_.
2. Removed obsolete locations from the database.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,4 +1,9 @@
 class PagesController < ApplicationController
   def index
+    if !params[:city_name]
+      redirect_to "/#{ENV['PRIMARY_CITY_NAME']}"
+    elsif Location.where(city_name: params[:city_name]).empty?
+      raise RoutingError, 'Not Found'
+    end
   end
 end
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -3,9 +3,21 @@ class Types::QueryType < Types::BaseObject
     description "MojoTech office locations"
   end
 
+  field :location, Types::LocationType do
+    description "Target location to receive updates"
+    argument :city_name, String, "City name of location", required: true
+  end
+
   def locations
     Rails.cache.fetch('locations-response', expires_in: 15.minutes) do
       Location.all
+    end
+  end
+
+  def location(options = {})
+    city_name = options[:city_name]
+    Rails.cache.fetch('location-response', expires_in: 15.minutes) do
+      Location.where(city_name: city_name).first
     end
   end
 
--- a/app/javascript/components/app.jsx
+++ b/app/javascript/components/app.jsx
@@ -1,5 +1,6 @@
 import 'regenerator-runtime/runtime';
 import React from 'react';
+import { BrowserRouter as Router, Switch, Route } from 'react-router-dom';
 import { ApolloClient } from 'apollo-client';
 import { ApolloProvider } from 'react-apollo';
 import { ApolloLink } from 'apollo-link';
@@ -75,7 +76,19 @@ const client = new ApolloClient({
 const App = () => (
   <ApolloProvider client={client}>
     <GlobalStyle />
-    <WidgetController client={client} />
+    <Router>
+      <Switch>
+        <Route
+          path="/:city_name"
+          component={routeProps => (
+            <WidgetController
+              client={client}
+              cityName={routeProps.match.params.city_name}
+            />
+          )}
+        />
+      </Switch>
+    </Router>
   </ApolloProvider>
 );
 export default App;
--- a/app/javascript/components/get-weather-query.jsx
+++ b/app/javascript/components/get-weather-query.jsx
@@ -6,21 +6,22 @@ import { assocPath } from 'ramda';
 import SunriseSunset from '@weather/sunrise-sunset';
 
 /* eslint-disable graphql/template-strings */
-const getPrimaryLocationWeather = (WeatherFrag, queryName) => gql`
-  {
-    primaryLocation {
-      latitude
-      longitude
-      ...SunriseSunsetLocation
-      weather {
-        ...${queryName}
-      }
+const getLocationWeather = (WeatherFrag, queryName) => gql`
+query getLocationWeather($cityName: String!) {
+  location(cityName: $cityName) {
+    latitude
+    longitude
+    ...SunriseSunsetLocation
+    weather {
+      ...${queryName}
     }
   }
+}
 
-  ${SunriseSunset.fragments.location}
-  ${WeatherFrag}
+${SunriseSunset.fragments.location}
+${WeatherFrag}
 `;
+
 /* eslint-enable graphql/template-strings */
 
 /* eslint-disable graphql/template-strings */
@@ -41,8 +42,12 @@ export const WeatherQuery = ({
   DisconnectedMessage,
   WeatherFrag,
   queryName,
+  cityName,
 }) => (
-  <Query query={getPrimaryLocationWeather(WeatherFrag, queryName)}>
+  <Query
+    query={getLocationWeather(WeatherFrag, queryName)}
+    variables={{ cityName }}
+  >
     {({ loading, error, data, subscribeToMore }) => {
       if (loading) {
         return <LoadingMessage />;
@@ -53,13 +58,13 @@ export const WeatherQuery = ({
       }
       return (
         <Subscription
-          primaryLocation={data.primaryLocation}
+          primaryLocation={data.location}
           subscribeToPublishedEvents={() =>
             subscribeToMore({
               document: subscribeWeatherPublished(WeatherFrag, queryName),
               variables: {
-                latitude: data.primaryLocation.latitude,
-                longitude: data.primaryLocation.longitude,
+                latitude: data.location.latitude,
+                longitude: data.location.longitude,
               },
               updateQuery: (prev, { subscriptionData }) => {
                 if (!subscriptionData.data) {
@@ -90,6 +95,7 @@ WeatherQuery.propTypes = {
     definitions: PropTypes.array.isRequired,
   }).isRequired,
   queryName: PropTypes.string.isRequired,
+  cityName: PropTypes.string.isRequired,
 };
 
 export default WeatherQuery;
--- a/app/javascript/components/top-corner-controller.jsx
+++ b/app/javascript/components/top-corner-controller.jsx
@@ -26,7 +26,7 @@ const Row = styled.div`
   margin-right: 100px;
 `;
 
-const TopCorner = ({ showWeather }) => {
+const TopCorner = ({ showWeather, cityName }) => {
   return (
     <div>
       <DateLogoRow>
@@ -36,8 +36,8 @@ const TopCorner = ({ showWeather }) => {
         <Logo />
       </IconWrapper>
       <Row>
-        <TimeHero />
-        {showWeather && <WeatherCorner />}
+        <TimeHero cityName={cityName} />
+        {showWeather && <WeatherCorner cityName={cityName} />}
       </Row>
     </div>
   );
@@ -45,5 +45,6 @@ const TopCorner = ({ showWeather }) => {
 
 TopCorner.propTypes = {
   showWeather: PropTypes.bool.isRequired,
+  cityName: PropTypes.string.isRequired,
 };
 export default TopCorner;
--- a/app/javascript/components/widget-controller.jsx
+++ b/app/javascript/components/widget-controller.jsx
@@ -13,8 +13,8 @@ import Weather from '@widgets/weather';
 import Traffic from '@widgets/traffic';
 
 const getWidgets = gql`
-  query getWidgets($id: Int!) {
-    primaryLocation {
+  query getWidgets($id: Int!, $cityName: String!) {
+    location(cityName: $cityName) {
       widgets {
         enabled {
           id
@@ -75,10 +75,14 @@ export class WidgetController extends React.Component {
         this.lastPrefetchTime < now - prefetchDelay
       ) {
         const { nextWidgetId } = this.state;
+        const { cityName } = this.props;
         this.lastPrefetchTime = now;
         client.query({
           query: getWidgets,
-          variables: { id: nextWidgetId },
+          variables: {
+            id: nextWidgetId,
+            cityName,
+          },
         });
       }
     }, 500);
@@ -160,12 +164,16 @@ export class WidgetController extends React.Component {
 
     const fetchPolicy = currentWidgetId === 0 ? 'network-only' : 'cache-first';
 
+    const { cityName } = this.props;
+
     return (
       <Query
         query={getWidgets}
-        variables={{ id: currentWidgetId }}
+        variables={{ id: currentWidgetId, cityName }}
         fetchPolicy={fetchPolicy}
-        onCompleted={({ primaryLocation: { widgets } }) => {
+        onCompleted={response => {
+          const { location } = response;
+          const { widgets } = location;
           const { enabled: enabledWidgets, byIdOrFirst: current } = widgets;
 
           const index = enabledWidgets.findIndex(w => w.id === current.id);
@@ -196,6 +204,7 @@ export class WidgetController extends React.Component {
                   showWeather={false}
                   totalTime={0}
                   isPaused={false}
+                  cityName={cityName}
                 />
               </Wrapper>
             );
@@ -215,14 +224,14 @@ export class WidgetController extends React.Component {
                   showWeather={false}
                   totalTime={0}
                   isPaused={false}
+                  cityName={cityName}
                 />
               </Wrapper>
             );
           }
 
-          const {
-            primaryLocation: { widgets },
-          } = data;
+          const { location } = data;
+          const { widgets } = location;
           const { enabled: enabledWidgets, byIdOrFirst: current } = widgets;
 
           const WidgetElement = widgetElements[current.name].Panel;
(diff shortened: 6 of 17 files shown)


app/graphql/types/query_type.rb
```diff
@@ -1,11 +1,16 @@
...
-  def locations
-    Rails.cache.fetch('locations-response', expires_in: 15.minutes) do
-      Location.all
+  def locations(city_name: nil)
+    if city_name
+      Location.all.select { |l| l.city_name == city_name }
+    else
+      Rails.cache.fetch('locations-response', expires_in: 15.minutes) do
+        Location.all
+      end
```
tucker-weed, line 13: Should I be wrapping `Location.all.select { |l| l.city_name == city_name }`with the Rails.cache.fetch as well?

app/controllers/pages_controller.rb
```diff
@@ -1,4 +1,6 @@
 class PagesController < ApplicationController
   def index
+    return if !params[:city_name] || Location.all.any? { |a| a.city_name == params[:city_name] }
+    raise RoutingError, 'Not Found'
```
cbarber, line 4: Stay positive. Generally speaking, prefer positive conditions over inverting conditions or reversing method return short circuits e.g.
`raise RoutingError, 'Not Found' if params[:city_name].empty? || ...`

In this case, I think we should probably hack in a `redirect_to ...` on empty param to the city_name of our PRIMARY_LOCATION to remove the need to handle primary location in our javascript code.

app/controllers/pages_controller.rb
```diff
@@ -1,4 +1,6 @@
 class PagesController < ApplicationController
   def index
+    return if !params[:city_name] || Location.all.any? { |a| a.city_name == params[:city_name] }
```
cbarber, line 3: With rails ActiveRecord, you have to be mindful when you potentially evaluate SQL queries with Enumerable methods: https://ruby-doc.org/core-2.5.1/Enumerable.html

If you're not sure, you can test it out in rails console first:
```
$ rails c 
irb(main):001:0> Location.all.any? { |a| a.city_name == "Providence" }
  Location Load (0.4ms)  SELECT "locations".* FROM "locations"
=> true
```

You can see above that we query all records from the locations table which means they are all loaded into memory, and then enumerated for the `city_name ==` condition. What want to do is only query what we need and nothing more with the ActiveRecord query methods:
https://api.rubyonrails.org/v5.2.3/classes/ActiveRecord/Relation.html#method-i-any-3F
https://api.rubyonrails.org/v5.2.3/classes/ActiveRecord/Relation.html#method-i-empty-3F
```
irb(main):002:0> Location.where(city_name: "Providence").any?
  Location Exists (0.7ms)  SELECT  1 AS one FROM "locations" WHERE "locations"."city_name" = ? LIMIT ?  [["city_name", "Providence"], ["LIMIT", 1]]
=> true
irb(main):003:0> Location.where(city_name: "Providence").empty?
  Location Exists (0.5ms)  SELECT  1 AS one FROM "locations" WHERE "locations"."city_name" = ? LIMIT ?  [["city_name", "Providence"], ["LIMIT", 1]]
=> false
```
Now we're only returning a single scalar from the database if there are rows and ActiveRecord checks the row count to give us the appropriate boolean result.

app/graphql/types/query_type.rb
```diff
@@ -1,11 +1,16 @@
...
     description "MojoTech office locations"
+    argument :city_name, String, "City name of location", required: false
   end
 
-  def locations
-    Rails.cache.fetch('locations-response', expires_in: 15.minutes) do
-      Location.all
+  def locations(city_name: nil)
+    if city_name
+      Location.all.select { |l| l.city_name == city_name }
```
cbarber, line 9: Same ActiveRecord performance improvement can made here with `find_by(...)`

app/graphql/types/query_type.rb
cbarber: Yes, but a probably a different key

app/graphql/types/query_type.rb
```diff
@@ -1,11 +1,16 @@
...
   field :locations, [Types::LocationType] do
     description "MojoTech office locations"
+    argument :city_name, String, "City name of location", required: false
   end
 
-  def locations
-    Rails.cache.fetch('locations-response', expires_in: 15.minutes) do
-      Location.all
+  def locations(city_name: nil)
+    if city_name
```
cbarber, line 8: I think this would be more accurately represented as a separate property in the schema `def location(city_name)` where city_name was not optional and we only ever return on location. Ideally, methods do just one thing.

app/javascript/components/app.jsx
```diff
@@ -71,10 +72,19 @@ const client = new ApolloClient({
...
+  return <WidgetController client={client} />;
+};
+
 const App = () => (
   <ApolloProvider client={client}>
     <GlobalStyle />
-    <WidgetController client={client} />
+    <Router>
+      <Switch>
+        <Route path="/" component={WidgetWindow} />
```
cbarber, line 84: I think we can omit this path, no? A missing city name isn't particularly useful to us since the plan is to do away with `PRIMARY_LOCATION` entirely.

app/javascript/components/get-weather-query.jsx
```diff
@@ -41,46 +59,61 @@ export const WeatherQuery = ({
...
   WeatherFrag,
   queryName,
-}) => (
-  <Query query={getPrimaryLocationWeather(WeatherFrag, queryName)}>
-    {({ loading, error, data, subscribeToMore }) => {
-      if (loading) {
-        return <LoadingMessage />;
+  location,
+}) => {
+  const cityName = location.pathname.substring(1);
```
cbarber, line 64: react-router passes props to its target component:
https://reactrouter.com/web/api/Route/route-props
https://reactrouter.com/web/api/match

You should be able to retrieve the matched `city_name` from the match prop. This will probably require passing this from WidgetWindow to WidgetController. Doing so would remove the need for the `withRouter` higher order component.

I think passing the `cityName` property directly will keep our queries independent of the url logic.

app/javascript/components/get-weather-query.jsx
```diff
@@ -41,46 +59,61 @@ export const WeatherQuery = ({
...
-              document: subscribeWeatherPublished(WeatherFrag, queryName),
-              variables: {
-                latitude: data.primaryLocation.latitude,
-                longitude: data.primaryLocation.longitude,
-              },
-              updateQuery: (prev, { subscriptionData }) => {
-                if (!subscriptionData.data) {
-                  return prev;
-                }
+      variables={isPrimary ? {} : { location: cityName }}
```
cbarber, line 73: Lets remove primary entirely.

db/seeds.rb
```diff
@@ -39,63 +39,64 @@
...
-  name: "Guests"
-) do |r|
-  r.enabled = false
-  r.duration_seconds = 20
-  r.position = 0
-  r.sidebar_text = 'Guests'
-  r.show_weather = false
-  r.location = primary_location
-end.save!
+Location.all.each do |location|
```
cbarber, line 42: Maybe we only do this for `Providence`, `Boulder`, and `San Diego`?

The other gotcha here is we can't `db:seed` the production database since it data already like events. We could use a migration to populate but that also doesn't feel right. I'm going to mull this over for a bit. I'm open to suggestions.

app/javascript/components/get-weather-query.jsx
tucker-weed: Are you referring to the useRouteMatch hook? I had trouble with that because I couldn't use it unless it nested in a functional component. Are you saying to reference match via props in the same way that location is referenced via props? Would I not need withRouter in that case?

app/javascript/components/app.jsx
tucker-weed: Should the "/" path return a 404 error then, and only URLs with a city attached to the slug are valid?

app/javascript/components/get-weather-query.jsx
cbarber: Nope. I mean `<Route ....` passes props to `component`, one of them is `match` which is based on the pattern(s) in the path. Try out:
```
const WidgetWindow = (props) => {
  console.log(props); // or you could: debugger;
  return <WidgetController client={client} />;
};
```

app/javascript/components/app.jsx
cbarber: It depends (TM).

I think the rails router should redirect `/` to `"/#{AppEnv['PRIMARY_LOCATION']}"`. This is to avoid `/` on the client side

I think the client side should skip handling `/` for now. Now the client side is agnostic of PRIMARY_LOCATION.

I'll add new tickets for:
* Client side route `/` should display some kind of prompt to select a location.
* Remove the server side redirect for `/`
* Remove PRIMARY_LOCATION
* :tada: the death of PRIMARY_LOCATION

db/seeds.rb
cbarber: It seems like seeds.rb is the only way to maintain the production database until we have an admin interface. I do think we can skip NYC and DC here since they're artifacts from an old design.

db/seeds.rb
```diff
@@ -39,63 +23,64 @@
...
+  Widget.find_or_initialize_by(
+    name: "Guests",
+    location: location
+  ) do |r|
+    r.enabled = false
+    r.duration_seconds = 20
+    r.position = 0
+    r.sidebar_text = 'Guests'
+    r.show_weather = false
+  end.save!
```
tmsavage, line 36: What does line 29 do and the following values you changed from the original version?

db/seeds.rb
chewnoill: If they are invalid we should drop them from the database. That doesn't need to happen here though, keeping this general case makes sense.

db/seeds.rb
chewnoill: Widgets are being created for each location instead of just the primary location. There is a whitespace change happening also that makes the diff a little more confusing.

---
Protip:

> Append ?w=1 to the github diff url.
>
> This can be used on pull requests, compare pages etc

https://coderwall.com/p/iuvlpg/remove-whitespace-changes-from-diff-in-github

chewnoill (commented):
There are a lot of things happing in this PR. Let's try to use our commits to help break up the changes so its easier to review.

* Be concise, changes you make in early commits should not be undone in the next commit
* Big diffs are hard to look at. Try to break them up so that one thing is happening in each commit.
* Each commits can build on top of previous commits, but each commit should still make sense on its own.

config/routes.rb
```diff
@@ -7,6 +7,8 @@
 
   post "/graphql", to: "graphql#execute"
 
+  get "/:city_name", controller: "PagesController", to: 'pages#index'
```
chewnoill, line 10: do we need to worry about this catch all route intercepting web_hooks traffic?

schema.graphql
```diff
@@ -103,6 +103,12 @@ type Query {
   # MojoTech slack/github events
   events(after: String, type: EventSource): EventCollection!
 
+  # Target location to receive updates
+  location(
+    # City name of location
+    cityName: String!
+  ): [Location!]!
```
chewnoill, line 110: I think this should return a single location, not a list of locations

```gql
  location(
    # City name of location
    cityName: String!
  ): Location!
```

chewnoill:
> :robot: :beep: :boop: needs a rebase

schema.graphql
```diff
@@ -107,7 +107,7 @@ type Query {
   location(
     # City name of location
     cityName: String!
-  ): [Location!]!
+  ): Location!
```
chewnoill, line 110: looks like this got squashed into the wrong commit. This should be on e80c119f6b474c8f67d943de41bfaa0b91138e10

tucker-weed: Just pushed the correct commits

app/graphql/types/query_type.rb
```diff
@@ -3,12 +3,23 @@ class Types::QueryType < Types::BaseObject
...
+
   def locations
     Rails.cache.fetch('locations-response', expires_in: 15.minutes) do
       Location.all
     end
   end
 
+  def location(city_name)
+    Rails.cache.fetch('location-response', expires_in: 15.minutes) do
+      Location.where(city_name: city_name[:city_name]).first
```
cbarber, line 19: Nit: `city_name` here is actually a hash argument. It would be idiomatic to use the ruby keyword arguments instead:
```
def locatoin(city_name: nil)
```
this is equivalent to:
```
def location(options = {})
  city_name = options[:city_name]
```

app/javascript/components/app.jsx
```diff
@@ -75,7 +76,16 @@ const client = new ApolloClient({
 const App = () => (
   <ApolloProvider client={client}>
     <GlobalStyle />
-    <WidgetController client={client} />
+    <Router>
+      <Switch>
+        <Route
+          path="/:city_name"
+          component={routeProps => (
+            <WidgetController client={client} location={routeProps.location} />
```
cbarber, line 84: `routeProps` should contain `match` generated from the pattern(s) in `path` above:
https://reactrouter.com/web/api/match 

This would skip the querystring logic from leaking into our component down below with `location.pathname.substring(1)`

app/javascript/components/get-weather-query.jsx
```diff
@@ -6,21 +6,22 @@ import { assocPath } from 'ramda';
...
-  {
-    primaryLocation {
-      latitude
-      longitude
-      ...SunriseSunsetLocation
-      weather {
-        ...${queryName}
-      }
+const getLocationWeather = (WeatherFrag, queryName) => gql`
+query getLocationWeather($loc: String!) {
```
cbarber, line 10: Thank you for naming this :+1:

app/javascript/components/top-corner-controller.jsx
```diff
@@ -26,7 +26,7 @@ const Row = styled.div`
   margin-right: 100px;
 `;
 
-const TopCorner = ({ showWeather }) => {
+const TopCorner = props => {
```
cbarber, line 29: Its idiomatic to destructure in the argument list like: `({ showWeather, location })`

app/javascript/components/widget-controller.jsx
```diff
@@ -160,12 +163,16 @@ export class WidgetController extends React.Component {
 
     const fetchPolicy = currentWidgetId === 0 ? 'network-only' : 'cache-first';
 
+    const cityName = this.props.location.pathname.substring(1);
+
     return (
       <Query
         query={getWidgets}
-        variables={{ id: currentWidgetId }}
+        variables={{ id: currentWidgetId, loc: cityName }}
```
cbarber, line 171: Would it be clearer for us to match `cityName` or `location` instead of `loc` in our graphql API?

db/seeds.rb
```diff
@@ -21,22 +21,6 @@
...
-end.save!
-
-Location.find_or_initialize_by(
-  city_name: 'NYC'
-) do |r|
-  r.latitude = 40.712775
-  r.longitude = -74.005973
-  r.time_zone = 'America/New_York'
-end.save!
-
```
cbarber, line 39: Nice! Thank you.

cbarber (commented):
Some minor comments on idomatic ruby/javascript and consistent naming between props/graphql variables.

app/javascript/components/widget-controller.jsx
```diff
@@ -160,12 +163,16 @@ export class WidgetController extends React.Component {
...
         fetchPolicy={fetchPolicy}
-        onCompleted={({ primaryLocation: { widgets } }) => {
+        onCompleted={response => {
+          const { location } = response;
```
chewnoill, line 174: this is getting confusing because we have a `props.location` and now a `response.location` with two different shapes. Can we rename one of them?

app/javascript/components/app.jsx
chewnoill: I agree. I would suggest a hooks solution so we don't need to put it in props at all. Lets also rename this since `location` is getting to overloaded.

```jsx
import { useRouteMatch } from "react-router-dom";

function useCurrentCityName(){
  const { params: {city_name} } =  useRouteMatch("/:city_name");
  return city_name;
}
```

tucker-weed: I see what you are saying, will make those changes

app/javascript/components/widget-controller.jsx
tucker-weed: Will go for cityName

app/javascript/components/widget-controller.jsx
tucker-weed: Will rename to props.cityName to match the gql query parameter name change to cityName. Seems clearer that way for sure.

app/javascript/components/app.jsx
cbarber: I think route props provides transparency over hooks.

app/javascript/components/widget-controller.jsx
```diff
@@ -44,223 +44,205 @@ const widgetElements = {
...
+            </Wrapper>
+          );
+        }
 
-      if (!transitionTime || pausedTime) {
-        return;
-      }
+        const { location } = data;
+        const { widgets } = location;
+        const { enabled: enabledWidgets, byIdOrFirst: current } = widgets;
```
chewnoill, line 127: unless you need to keep these intermediate variables around for some reason you can destructure all at once like this:

```js
const { 
  location: { 
    widgets: {
      enabled: enabledWidgets, 
      byIdOrFirst: current } } } = data;
```

tucker-weed: I'll address that in the refactor PR

chewnoill (approved):
Nice work @tucker-weed!

PR #235, opened 2021-05-27
feat: add Boulder traffic cams

bbazyk:
boulder_traffic_cams

Ticket to add support for Boulder Traffic Cams, id : #178271931
Boulder cams are livestreams, not images as with the Providence cams. This PR adds react hls player library to enable displaying these streams. Also adds a field to the database which determines how to display the feed.
Cams are visible after 4pm and can be viewed by setting the env field for CITY_NAME to "Boulder" - as was the case in viewing cams before.

### DIFF
--- a/app/javascript/components/widgets/traffic/panel.jsx
+++ b/app/javascript/components/widgets/traffic/panel.jsx
@@ -4,6 +4,7 @@ import { splitAt, take } from 'ramda';
 import { Query } from 'react-apollo';
 import gql from 'graphql-tag';
 import styled from 'styled-components';
+import ReactHlsPlayer from 'react-hls-player';
 import { colors, weights, fonts, fontSizes } from '@lib/theme';
 import { GreySubText } from '@components/typography';
 import { LoadingMessage, DisconnectedMessage } from '@messages/message';
@@ -15,6 +16,7 @@ const getTrafficCams = gql`
         id
         title
         url
+        feedFormat
       }
     }
   }
@@ -56,16 +58,21 @@ const Notice = styled(GreySubText)`
   margin-right: 750px;
 `;
 
-const TrafficCam = ({ title, url }) => (
+const TrafficCam = ({ title, url, feedFormat }) => (
   <div>
     <Title>{title}</Title>
-    <TrafficCamImage src={url} />
+    {feedFormat === 'video' ? (
+      <ReactHlsPlayer src={url} autoPlay controls width="450px" height="auto" />
+    ) : (
+      <TrafficCamImage src={url} />
+    )}
   </div>
 );
 
 TrafficCam.propTypes = {
   title: PropTypes.string.isRequired,
   url: PropTypes.string.isRequired,
+  feedFormat: PropTypes.string.isRequired,
 };
 
 const Traffic = () => (
@@ -84,18 +91,19 @@ const Traffic = () => (
         return null;
       }
 
-      const [firstColumn, secondColumn] = splitAt(3, trafficCams);
+      const columnSize = Math.floor(trafficCams.length / 2);
+      const [firstColumn, secondColumn] = splitAt(columnSize, trafficCams);
 
       return (
         <>
           <Row>
             <Column>
-              {take(3, firstColumn).map(({ id, ...cam }) => (
+              {take(columnSize, firstColumn).map(({ id, ...cam }) => (
                 <TrafficCam key={id} {...cam} />
               ))}
             </Column>
             <Column>
-              {take(3, secondColumn).map(({ id, ...cam }) => (
+              {take(columnSize, secondColumn).map(({ id, ...cam }) => (
                 <TrafficCam key={id} {...cam} />
               ))}
             </Column>
--- a/app/javascript/components/widgets/traffic/panel.jsx
+++ b/app/javascript/components/widgets/traffic/panel.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { splitAt, take } from 'ramda';
+import { splitAt, take, uniq, map, compose } from 'ramda';
 import { Query } from 'react-apollo';
 import gql from 'graphql-tag';
 import styled from 'styled-components';
@@ -90,9 +90,12 @@ const Traffic = () => (
       if (!trafficCams) {
         return null;
       }
-
       const columnSize = Math.floor(trafficCams.length / 2);
       const [firstColumn, secondColumn] = splitAt(columnSize, trafficCams);
+      const getUniqUrls = compose(
+        uniq,
+        map(cam => new URL(cam.url).origin),
+      );
 
       return (
         <>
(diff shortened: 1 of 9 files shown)


db/seeds.rb
```diff
@@ -102,39 +103,77 @@
...
 
 TrafficCam.find_or_initialize_by(
   title: "Tobey Street"
 ) do |r|
   r.url = "http://www.dot.ri.gov/img/travel/camimages/6-2%20Rt%206%20E%20%20%2010%20N%20@%20Tobey%20St.jpg"
   r.location = providence
+  r.feed_format = "image"
+end.save!
+
+TrafficCam.find_or_initialize_by(
```
jennyckaplan, line 149: Let's make adding the Boulder cams a separate commit

package.json
```diff
@@ -31,6 +31,7 @@
     "react-apollo": "^2.1.4",
     "react-dom": "^16.8.6",
     "react-highlight-words": "^0.16.0",
+    "react-hls-player": "^3.0.1",
```
jennyckaplan, line 34: This should also be part of the second commit, since the `ReactHlsPlayer` is only used in the next commit with the front-end changes

yarn.lock
```diff
@@ -4551,7 +4551,7 @@ etag@~1.8.1:
   resolved "https://registry.yarnpkg.com/etag/-/etag-1.8.1.tgz#41ae2eeb65efa62268aebfea83ac7d79299b0887"
```
jennyckaplan, line 4551: The `yarn.lock` file change should also be in the next commit with the `package.json` change.

tmsavage:
I recognize all of those camera locations haha!

db/seeds.rb
```diff
@@ -7,13 +7,14 @@
 end
 providence.save!
 
-Location.find_or_initialize_by(
+boulder = Location.find_or_initialize_by(
```
tmsavage, line 10: Does this initialize to a boulder variable? In general, should this be done for the other towns?

bbazyk: The variable is only referenced by initialized "TrafficCam"  objects - just as a means of differentiating between which should be displayed to the widget.

chewnoill (dismissed):
Nice work @bbazyk!

jennyckaplan:
LGTM, except for the `Notice` that still only says `dot.ri.gov` even when displaying Boulder - we should add this in as a field in the database as well and dynamically display the correct source @bbazyk

app/javascript/components/widgets/traffic/panel.jsx
```diff
@@ -84,18 +91,19 @@ const Traffic = () => (
...
-      const [firstColumn, secondColumn] = splitAt(3, trafficCams);
+      const columnSize = Math.floor(trafficCams.length / 2);
+      const [firstColumn, secondColumn] = splitAt(columnSize, trafficCams);
 
       return (
         <>
           <Row>
             <Column>
-              {take(3, firstColumn).map(({ id, ...cam }) => (
+              {take(columnSize, firstColumn).map(({ id, ...cam }) => (
```
chewnoill, line 101: does take make sense if we're using the whole list?
can this just be `firstColumn.map(...`?

chewnoill: Oh I see, we want both columns to be the same size

db/seeds.rb
```diff
@@ -109,6 +109,7 @@
   r.url = "http://www.dot.ri.gov/img/travel/camimages/95-21%20I-95%20S%20@%20Broad%20St%20.jpg"
   r.location = providence
   r.feed_format = "image"
+  r.source_url = "http://www.dot.ri.gov"
```
chewnoill, line 112: I don't think we need an extra column for this, if it's just the base url for the url we are already keeping. You can parse this from the js side like this:

```js
const url = "http://www.dot.ri.gov/img/travel/camimages/95-21%20I-95%20S%20@%20Broad%20St%20.jpg"
const source_url = new URL(url).origin
```

app/javascript/components/widgets/traffic/panel.jsx
```diff
@@ -93,6 +94,7 @@ const Traffic = () => (
 
       const columnSize = Math.floor(trafficCams.length / 2);
       const [firstColumn, secondColumn] = splitAt(columnSize, trafficCams);
+      const sourceUrls = uniq(trafficCams.map(cam => cam.sourceUrl));
```
chewnoill, line 97:
```js
const sourceUrls = uniq(trafficCams.map(cam => new URL(cam.url).origin));
```

cbarber: Bonus points for functional composition:
```
const getUniqueUrls = compose(uniq, map(cam => new URL(cam.url).origin);
...
          {getUniqUrls(trafficCams).map(url => {
            <Notice>{url}</Notice>
          ))}
```

db/migrate/20210527193617_add_feed_format_to_traffic_cams.rb
```diff
@@ -0,0 +1,5 @@
+class AddFeedFormatToTrafficCams < ActiveRecord::Migration[5.2]
+  def change
+    add_column :traffic_cams, :feed_format, :string
```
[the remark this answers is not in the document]
chewnoill: does `default` need to be added before `null: false` or can it be same time? A little confusing about how rails runs migrations, is there a way to see what SQL ends up running here?

chewnoill (approved):
🚀

db/migrate/20210527193617_add_feed_format_to_traffic_cams.rb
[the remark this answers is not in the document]
cbarber: `null: false` requires a `default: <value>` or existing data. The latter can be accomplished with:
```
add_column :traffic_cams, :feed_format, :string
execute "UPATE traffic_cams SET feed_format = '...'"
change_column :traffic_cams, :feed_format, :string, null: false
```
If you want to use a modal to build the update statement, you create a temporary modal class inside the migration to prevent breaking the migration with future modal changes.

In this particular case, I think `null: false, default: 'image'` is sufficient.

app/javascript/components/widgets/traffic/panel.jsx
bbazyk: it just makes it easier for anyone to export the function later, if they need its output

db/migrate/20210527193617_add_feed_format_to_traffic_cams.rb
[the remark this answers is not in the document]
bbazyk: This is ready for re-review @cbarber @chewnoill

app/javascript/components/widgets/traffic/panel.jsx
```diff
@@ -90,9 +90,14 @@ const Traffic = () => (
...
       }
-
       const columnSize = Math.floor(trafficCams.length / 2);
       const [firstColumn, secondColumn] = splitAt(columnSize, trafficCams);
+      const getUniqUrls = compose(
+        uniq,
+        map(cam => new URL(cam.url).origin),
+      );
+
+      console.log(getUniqUrls(trafficCams));
```
chewnoill, line 100: 🦅

chewnoill (approved):
there is an extra console log, but otherwise LGTM

Nice commit history also, very clean :atom:

PR #236, opened 2021-05-27
Sp/move env variables into database

sandipa12:
Moved the following env variables into the location record in the database.
* WIFI_NAME
* WIFI_PASSWORD
* BATHROOM_CODE
* Also updated seeds.rb to initialize a new database.
https://www.pivotaltracker.com/story/show/178116607


sandipa12:
branched off of the feature branch

PR #237, opened 2021-05-27
feat:move env variables into the location record in the database

sandipa12:
Moved the following env variables into the location record in the database.

*WIFI_NAME
*WIFI_PASSWORD
*BATHROOM_CODE
Also updated seeds.rb to initialize a new database.
https://www.pivotaltracker.com/story/show/178116607

### DIFF
--- a/app/graphql/types/location_type.rb
+++ b/app/graphql/types/location_type.rb
@@ -13,8 +13,8 @@ class Types::LocationType < Types::BaseObject
   field "moon_phase", Float
   field "solar_cycles", [Types::SolarcycleType]
 
-  field "wifi_name", String
-  field "wifi_password", String
-  field "bathroom_code", String
+  field "wifi_name", String, null: true
+  field "wifi_password", String, null: true
+  field "bathroom_code", String, null: true
   field "traffic_cams", [Types::TrafficCamType]
 end
--- a/app/javascript/components/wifi.jsx
+++ b/app/javascript/components/wifi.jsx
@@ -44,7 +44,7 @@ export const Wifi = () => (
         }
 
         const { wifiName, wifiPassword } = data.primaryLocation;
-        if (!wifiName) {
+        if (!wifiName || !wifiPassword) {
           return null;
         }
 
--- a/schema.graphql
+++ b/schema.graphql
@@ -56,7 +56,7 @@ scalar IconIdToName
 scalar KelvinToFarenheit
 
 type Location {
-  bathroomCode: String!
+  bathroomCode: String
   cityName: String!
   dayAnnouncements: [Announcement!]!
   id: ID!
@@ -69,8 +69,8 @@ type Location {
   trafficCams: [TrafficCam!]!
   weather: Weather!
   widgets: WidgetCollection!
-  wifiName: String!
-  wifiPassword: String!
+  wifiName: String
+  wifiPassword: String
 }
 
 type MojoTweet {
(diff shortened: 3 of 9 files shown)


cbarber (commented):
This is an excellent start.

Some tips on how I approach these kinds of find/replace tasks is with git grep:
```
$ git grep WIFI
.env.local.sample:WIFI_NAME=mojotech-guest
.env.local.sample:WIFI_PASSWORD=p@ssw0rd
app/javascript/components/wifi.jsx:              <WhiteSubTitle>WIFI</WhiteSubTitle>
app/models/location.rb:    ENV['WIFI_NAME']
app/models/location.rb:    ENV['WIFI_PASSWORD']
```

The relevant methods in location.rb from above are:
```
   61   def wifi_name
   62     ENV['WIFI_NAME']
   63   end
   64 
   65   def wifi_password
   66     ENV['WIFI_PASSWORD']
   67   end
```

which should now come from the now database. This gives us a new search for how to find the graphql types:
```
$ git grep wifi_name
app/graphql/types/location_type.rb:  field "wifi_name", String
app/javascript/components/wifi.jsx:import WifiIcon from '@assets/images/wifi_name.svg';
app/models/location.rb:  def wifi_name
```

There is `app/graphql/types/location_type.rb` which maps the `wifi_name` method to a String type `wifiName` (the graphql gem automatically converts snake case to camel case). 

You can also

sandipa12:
"git grep" seems to be really cool !!!

app/graphql/types/location_type.rb
```diff
@@ -13,8 +13,8 @@ class Types::LocationType < Types::BaseObject
   field "moon_phase", Float
   field "solar_cycles", [Types::SolarcycleType]
 
-  field "wifi_name", String
-  field "wifi_password", String
-  field "bathroom_code", String
+  field "wifi_name", String, null: true
+  field "wifi_password", String, null: true
+  field "bathroom_code", String, null: true
```
fallynlogan, line 18: Why would we want to add this (null: true)?

schema.graphql
```diff
@@ -69,8 +69,8 @@ type Location {
   trafficCams: [TrafficCam!]!
   weather: Weather!
   widgets: WidgetCollection!
-  wifiName: String!
-  wifiPassword: String!
+  wifiName: String
+  wifiPassword: String
```
fallynlogan, line 73: What does removing '!' do/mean?

app/javascript/components/wifi.jsx
```diff
@@ -38,7 +38,7 @@ export const Wifi = () => (
         }
 
         const { wifiName, wifiPassword } = data.primaryLocation;
-        if (!wifiName) {
+        if (!wifiName || !wifiPassword) {
```
tmsavage, line 41: Nice! This is a good logical expression when both wifiName and wifiPassword should have a non-NULL value.

schema.graphql
cbarber: This is the graphql schema Interface Definition Language, IDL. `!` means a field is not null:
https://graphql.org/learn/schema/#lists-and-non-null

app/graphql/types/location_type.rb
chewnoill: I imagine because not all of the locations have wifi or secured bathroom doors

schema.graphql
chewnoill: caused by the `null: true` that got added over here https://github.com/mojotech/helios2/pull/237/files#r641690789

chewnoill (approved):
🚀

app/graphql/types/location_type.rb
sandipa12: @fallynlogan the field  may return nil when it is set to null:true.

PR #240, opened 2021-06-03
feat: Implemented query and comparison of previous/new SHA value

tmsavage:
### Files Edited
- app/graphql/types/query_type.rb
- app/javascript/components/app.jsx
- app/javascript/components/get-version.jsx

### Description
Implemented a new query to attain a current (dummy) SHA value and conditions to compare with a new (dummy) SHA value. If the two do not match, then there is an update to the application; the page will refresh and the recorded current SHA is overwritten by the new SHA value.

The query gets the value stored in a new variable in the .env file. Ex: `HEROKU_SLUG_COMMIT=foosomething`, which serves as the current dummy SHA value.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,6 +57,12 @@ jobs:
       - run: bundle exec rake db:create
       - run: bundle exec rake db:schema:load
 
+      - run:
+          name: run diff on schema.json
+          command: |
+            bundle exec rake graphql:schema:dump
+            git diff --quiet schema.graphql
+
       # run tests!
       - run:
           name: run tests
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -39,4 +39,12 @@ class Types::QueryType < Types::BaseObject
       Location.primary
     end
   end
+
+  field :schema_sha, String do
+    description "Get current sha"
+  end
+
+  def schema_sha
+    AppEnv['HEROKU_SLUG_COMMIT']
+  end
 end
--- a/app/graphql/types/subscription_type.rb
+++ b/app/graphql/types/subscription_type.rb
@@ -14,4 +14,8 @@ class Types::SubscriptionType < Types::BaseObject
   field :announcement_published, Types::AnnouncementType, description: "An announcement was published"
 
   def announcement_published; end
+
+  field :schema_sha, String, description: "Updated SHA value"
+
+  def schema_sha; end
 end
--- a/app/javascript/components/get-version.jsx
+++ b/app/javascript/components/get-version.jsx
@@ -1,0 +1,35 @@
+import React from 'react';
+import { Subscription } from 'react-apollo';
+import gql from 'graphql-tag';
+
+const getVersion = gql`
+  {
+    schemaSha
+  }
+`;
+
+let lastValue = null;
+
+export const Version = () => (
+  <Subscription subscription={getVersion}>
+    {({ loading, error, data }) => {
+      if (loading || error) {
+        return null;
+      }
+
+      const newValue = data.schemaSha;
+      if (lastValue == null) {
+        lastValue = newValue;
+        console.log(lastValue);
+        console.log(newValue);
+      } else if (newValue !== lastValue) {
+        lastValue = newValue;
+        window.location.reload();
+      }
+
+      return newValue;
+    }}
+  </Subscription>
+);
+
+export default Version;
--- a/app/javascript/components/app.jsx
+++ b/app/javascript/components/app.jsx
@@ -14,6 +14,7 @@ import ActionCableLink from 'graphql-ruby-client/subscriptions/ActionCableLink';
 import { persistCache } from 'apollo-cache-persist';
 import WidgetController from '@components/widget-controller';
 import helioSchema from '@javascript/schema.json';
+import GetVersion from '@components/get-version';
 import GlobalStyle from '../styles';
 
 const cable = ActionCable.createConsumer('/cable');
@@ -76,6 +77,7 @@ const App = () => (
   <ApolloProvider client={client}>
     <GlobalStyle />
     <WidgetController client={client} />
+    <GetVersion />
   </ApolloProvider>
 );
 export default App;
--- a/app/javascript/schema.json
+++ b/app/javascript/schema.json
@@ -118,6 +118,24 @@
               "deprecationReason": null
             },
             {
+              "name": "schemaSha",
+              "description": "Get current sha",
+              "args": [
+
+              ],
+              "type": {
+                "kind": "NON_NULL",
+                "name": null,
+                "ofType": {
+                  "kind": "SCALAR",
+                  "name": "String",
+                  "ofType": null
+                }
+              },
+              "isDeprecated": false,
+              "deprecationReason": null
+            },
+            {
               "name": "tweets",
               "description": "Tweets from MojoTech",
               "args": [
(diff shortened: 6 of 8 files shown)


app/graphql/types/query_type.rb
```diff
@@ -39,4 +39,14 @@ def primary_location
...
     end
   end
+  
+  #Define a field
+  field :schema_sha, String do
+    description "Get current sha"
+  end
+
+  def schema_sha
+    ENV['HEROKU_SLUG_COMMIT']
```
chewnoill, line 49: I think we've switched this to `AppEnv`

see https://github.com/mojotech/helios2/pull/224

app/javascript/components/get-version.jsx
```diff
@@ -0,0 +1,38 @@
...
+      if (error) {
+        return <ErrorMessage />;
+      }
+
+      var newValue = data.schemaSha;
+      if (lastValue == null) {
+        lastValue = newValue;
+      } else if (newValue !== lastValue) {
+        lastValue = newValue;
+        object.location.reload();
```
chewnoill, line 30: does this work?
I think you want `window.location.reload()`

chewnoill (commented):
needs a squash and a rebase

app/javascript/components/get-version.jsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import { Query } from 'react-apollo';
+import gql from 'graphql-tag';
+import { LoadingMessage, ErrorMessage } from '@messages/default-messages';
+
+var lastValue = null;
+
+const getVersion = gql`
+  {
```
cbarber, line 9: Nit: I think specifying a name for the graphql query will provide a name in the apollo devs tools.

app/javascript/components/get-version.jsx
```diff
@@ -0,0 +1,38 @@
...
+    schemaSha
+  }
+`;
+
+export const Version = () => (
+  <Query query={getVersion}>
+    {({ loading, error, data }) => {
+      if (loading) {
+        return <LoadingMessage />;
```
cbarber, line 18: I think we should probably return `null` for both loading and error. No reason for this component to create dom elements.

app/javascript/components/get-version.jsx
```diff
@@ -0,0 +1,38 @@
...
+      }
+
+      if (error) {
+        return <ErrorMessage />;
+      }
+
+      var newValue = data.schemaSha;
```
cbarber, line 25: Nit: this could be a const

app/javascript/components/get-version.jsx
```diff
@@ -0,0 +1,38 @@
...
+var lastValue = null;
+
+const getVersion = gql`
+  {
+    schemaSha
+  }
+`;
+
+export const Version = () => (
+  <Query query={getVersion}>
```
cbarber, line 15: Does this `Query` refresh the page on a stop server/edit HEROKU_SLUG_COMMIT/start server?

app/javascript/components/get-version.jsx
tmsavage: Yeah it was meant to be `window.location.reload()`. object is not correct.

fallynlogan (commented):
Why would we want to compare the new and old SHA's?

tmsavage:
> Why would we want to compare the new and old SHA's?

The purpose of comparing the current SHA on a client with a new SHA sent from the server is to identify if the application version on a client is out of date. Making this comparison would trigger a refresh on the client with the new received SHA, showing the most up to date version of helios.

app/javascript/components/get-version.jsx
chewnoill: This query will only trigger a refresh after mounting

app/javascript/components/get-version.jsx
```diff
@@ -0,0 +1,33 @@
...
+const getVersion = gql`
+  {
+    schemaSha
+  }
+`;
+
+let lastValue = null;
+
+export const Version = () => (
+  <Query query={getVersion}>
```
tucker-weed, line 14: Is there a reason lastValue is defined outside of the Version component? Does it make it such that new renders of the Version component don't reset lastValue?

tmsavage: Yes, exactly. This is a global variable, which is outside the scope of the Version component. The reason `lastValue` is a global variable is to have easy access to the current SHA value on the client. It is initialized as `null`, but is reset to the current SHA after the first instance of Version.

app/javascript/components/get-version.jsx
```diff
@@ -0,0 +1,35 @@
...
+    {({ loading, error, data }) => {
+      if (loading || error) {
+        return null;
+      }
+ 
+      const newValue = data.schemaSha;
+      if (lastValue == null) {
+        lastValue = newValue;
+        console.log(lastValue);
+        console.log(newValue);
```
tmsavage, line 24: `console.log()` for testing purposes.

tmsavage (commented):
fixed formatting

app/javascript/components/get-version.jsx
chewnoill: +1 this is best practice

https://www.apollographql.com/docs/react/data/operation-best-practices/#name-all-operations

app/javascript/components/get-version.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+const getVersion = gql`
+  {
+    schemaSha
+  }
+`;
+
+let lastValue = null;
+
+export const Version = () => (
```
cbarber, line 13: I'm curious if we could call the ApolloClient directly with `subscribe` to skip React entirely for our reload logic:
https://www.apollographql.com/docs/react/api/core/ApolloClient/#ApolloClient.subscribe

app/graphql/types/subscription_type.rb
```diff
@@ -14,4 +14,8 @@ def weather_published(latitude:, longitude:); end
   field :announcement_published, Types::AnnouncementType, description: "An announcement was published"
 
   def announcement_published; end
+
+  field :schema_sha, String, description: "Updated SHA value"
+
+  def schema_sha; end
```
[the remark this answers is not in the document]
tmsavage: Good point, I'll change that.

.circleci/config.yml
```diff
@@ -57,6 +57,12 @@ jobs:
       - run: bundle exec rake db:create
       - run: bundle exec rake db:schema:load
 
+      - run:
+          name: run diff on schema.json
+          command: |
+            bundle exec rake graphql:schema:dump
+            git diff --quiet schema.graphql
+
```
chewnoill, line 65: be careful with rebasing, looks like this got pulled into your branch by mistake.  Do you know what happened here?

app/javascript/components/get-version.jsx
chewnoill: can you use this same query as a query AND a subscription? I thought you needed to name subscriptions specifically.

app/javascript/components/get-version.jsx
chewnoill: lets make sure not to merge these

.circleci/config.yml
tmsavage: Hm, I was trying to fix my branch which diverged into two separate ones. My terminal stated it had '2 and 17 different commits each'. This is probably what was causing some errors while rebasing.

PR #241, opened 2021-06-07
check diff on schema.json

fallynlogan:
**Ticket description: **

We sometimes forget to dump the graphql schema from rails leaving schema.json in a broken state until future javascript changes. We should catch this automatically with a CircleCI step that runs the schema dump and asserts the schema files were not changed

**Changes: **

Add shell script to check if schema.json is included in git diff

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,6 +57,11 @@ jobs:
       - run: bundle exec rake db:create
       - run: bundle exec rake db:schema:load
 
+      - run:
+          name: run diff on schema.json
+          command: |
+            git diff --quiet schema.graphql
+
       # run tests!
       - run:
           name: run tests


.circleci/config.yml
```diff
@@ -57,6 +57,11 @@ jobs:
       - run: bundle exec rake db:create
       - run: bundle exec rake db:schema:load
 
+      - run:
+          name: run diff on schema.json
+          command: |
+            test/schema_diff.sh
```
chewnoill, line 63: I wonder if we could fit this onto one line, instead of adding a script file? What do you think about this?
```sh
command: git diff --quiet schema.json || exit 1
```

chewnoill: or maybe this? https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---exit-code

test/schema_diff.sh
```diff
@@ -0,0 +1,2 @@
+diff=`git diff`
+grep -q 'schema.json' <<< $diff && echo "changes" exit 1 || echo "no changes" exit 0
```
cbarber, line 2: If we're going to use a shell script, we should specify the executable and (if bash) should pipefail: https://sipb.mit.edu/doc/safe-shell/

.circleci/config.yml
fallynlogan: Sounds good, this is a lot simpler!

chewnoill (approved):
🚀

cbarber:
@chewnoill @fallynlogan I'm confused. How does this actually work?

PR #243, opened 2021-06-08
feat: active admin frontend to manage locations

bbazyk:
helios active admin

ticket id: 178270028
accessible through http://localhost:5000/admin
user email: admin@example.com
password: password
Use the above tabs to navigate models and edit their parameters. Can be used to alter the displayed wifi and bathroom codes, rearrange and adjust widget screen time, or even delete a traffic cam from view.


config/initializers/active_admin.rb
```diff
@@ -0,0 +1,335 @@
+ActiveAdmin.setup do |config|
+  # == Site Title
+  #
+  # Set the title that is displayed on the main layout
+  # for each of the active admin pages.
+  #
+  config.site_title = "Helios2"
+
+  # Set the link url for the title. For example, to take
```
chewnoill, line 9: I think we can remove all of these comments, these files are all pretty simple

db/seeds.rb
```diff
@@ -186,3 +186,5 @@
   r.location = boulder
   r.feed_format = "video"
 end.save!
+
+AdminUser.create!(email: 'admin@example.com', password: 'password', password_confirmation: 'password') if Rails.env.development?
```
chewnoill: Let's make sure that the seed script is **idempotent**. It should be able to be run more then once without breaking.

app/admin/events.rb
```diff
@@ -5,7 +5,7 @@
   #
   # Uncomment all parameters which should be permitted for assignment
   #
-  # permit_params :source, :external_id
+  permit_params :source, :external_id
```
chewnoill, line 8: lets squash this commit with cb3e566b8df8df287ab1890bf8855749a46c4391 and remove all of these extra comments

db/seeds.rb
```diff
@@ -187,4 +187,4 @@
   r.feed_format = "video"
 end.save!
 
-AdminUser.create!(email: 'admin@example.com', password: 'password', password_confirmation: 'password') if Rails.env.development?
+AdminUser.create!(email: 'helios@mojotech.com', password: 'mojojojo', password_confirmation: 'mojojojo') if Rails.env.development?
```
chewnoill, line 190: if this is only for development, its nice to keep the username/password simpler: admin@example.com/password is pretty standard for development

chewnoill: also, lets make sure the readme gets updated

db/seeds.rb
```diff
@@ -186,3 +186,11 @@
...
   r.feed_format = "video"
 end.save!
+
+if Rails.env.development?
+  AdminUser.create!(
+    email: 'helios@mojotech.com',
+    password: 'mojojojo',
+    password_confirmation: 'mojojojo'
+  )
+end
```
tucker-weed, line 196: Should there be an else for this conditional to create an admin user for production?

bbazyk: When rails boots up, it would automatically start with any file named development (if in development mode). It doesn't mean that the development stage isn't executed only once, and the difference between the environments has more to do with caching. Either way, the internet doesn't recommend any code of the format: 
if Rails.env.production?
  do_additional_work
end
so I'll try looking into something else...

cbarber (commented):
Looking good. I agree with Will that some of these feel like sub-atomic commits. I could see the case for adding deps being separate from adding functionality or squashing them all.

I noticed some usability issues where Widgets are showing either the Location#id or the Location#object_id where we could opt into showing the city name instead:
Screenshot from 2021-06-10 17-53-05
Screenshot from 2021-06-10 17-52-17

db/seeds.rb
```diff
@@ -186,3 +186,11 @@
   r.location = boulder
   r.feed_format = "video"
 end.save!
+
+if Rails.env.development?
+  AdminUser.create!(
+    email: 'admin@example.com',
+    password: 'password',
+    password_confirmation: 'password'
+  )
```
tmsavage, line 195: So this creates an admin user and their credentials are stored into the database?

config/initializers/devise.rb
```diff
@@ -0,0 +1,313 @@
...
+  #
+  # ActiveSupport.on_load(:devise_failure_app) do
+  #   include Turbolinks::Controller
+  # end
+
+  # ==> Configuration for :registerable
+
+  # When set to false, does not sign a user in automatically after their password is
+  # changed. Defaults to true, so a user is signed in automatically after changing a password.
+  # config.sign_in_after_change_password = true
```
chewnoill, line 312: I think we can remove all these comments also

db/seeds.rb
```diff
@@ -186,3 +186,11 @@
   r.location = boulder
   r.feed_format = "video"
 end.save!
+
+if Rails.env.development?
```
chewnoill, line 190: I would like to create an admin user in production also. Maybe something like this:

  AdminUser.create!(
    email: ENV['ADMIN_EMAIL']
    password: ENV['ADMIN_PASSWORD'],
    password_confirmation: ENV['ADMIN_PASSWORD']
  )

config/routes.rb
```diff
@@ -1,4 +1,6 @@
 Rails.application.routes.draw do
+  devise_for :admin_users, ActiveAdmin::Devise.config
+  ActiveAdmin.routes(self)
```
tucker-weed, line 3: What does this do?

db/migrate/20210607180232_devise_create_admin_users.rb
```diff
@@ -0,0 +1,43 @@
...
+      ## Confirmable
+      # t.string   :confirmation_token
+      # t.datetime :confirmed_at
+      # t.datetime :confirmation_sent_at
+      # t.string   :unconfirmed_email # Only if using reconfirmable
+
+      ## Lockable
+      # t.integer  :failed_attempts, default: 0, null: false # Only if lock strategy is :failed_attempts
+      # t.string   :unlock_token # Only if unlock strategy is :email or :both
+      # t.datetime :locked_at
```
fallynlogan, line 33: Would we want to keep all of these in the code (for reference or to be used later) or should these comments be deleted?

spec/models/admin_user_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe AdminUser, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
fallynlogan, line 4: What is this doing?

config/routes.rb
sandipa12: Is it to use devise to log in regular users with same user model?

db/seeds.rb
```diff
@@ -186,3 +186,9 @@
   r.location = boulder
   r.feed_format = "video"
 end.save!
+
+AdminUser.create!(
```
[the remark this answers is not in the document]
chewnoill: Maybe `find_or_create_by` is enough?

spec/models/admin_user_spec.rb
bbazyk: This file was added by active admin, I've found other examples online and they all seem to contain code relevant for creating a more accurate login page (making sure the username contains a valid email address, rejecting those that are invalid, rejecting two accounts with the same email address, etc.)

db/migrate/20210607180232_devise_create_admin_users.rb
bbazyk: I just asked about it, and most blocks of comments and documentation that come with added files can be deleted for simplicity's sake. They're not that difficult to find again from the source (if say someone else on the project wanted some direction for the code they're tasked with).

bbazyk:
Removed comments from code, added 'if statement' to seeds.rb that enables reseeding without breaking, admin user for rails production mode, and display city names instead of location IDs in the widgets model for active admin.
Ready for one final review? :) @chewnoill @cbarber @jennyckaplan

chewnoill (approved):
Nice work @bbazyk

PR #247, opened 2021-06-10
husky formatter

chewnoill:
* formats js, jsx, json files

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,6 +48,7 @@ jobs:
             - ~/.yarn-cache
           key: yarn-cache-{{ checksum "yarn.lock" }}
 
+      - run: yarn format --check
       - run: yarn run lint
       - run: yarn run test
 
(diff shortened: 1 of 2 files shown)


schema.graphql
```diff
@@ -306,4 +306,4 @@ type WidgetCollection {
   byIdOrFirst(id: Int): Widget!
   enabled: [Widget!]!
   next(currentWidgetId: Int): Widget!
-}
```
cbarber, line 309: This will be problematic since our `graphql:schema:dump` CI step will likely fail on this

chewnoill:
> reasons mentioned in slack

The only reason I saw was "You'll have unrelated prettier changes on your last commit if you have more than one", 
which is solved with [git-format-staged](https://github.com/hallettj/git-format-staged), because it only formats staged hunks. 

Are there other reasons as well?

cbarber:
Yes, don't impose workflow choices on other developers

schema.graphql
chewnoill: dropped gql files from the glob pattern

PR #248, opened 2021-06-11
feat: server broadcasting to subscribed clients with updated SHA value

tmsavage:
The file included in this PR create a subscription type which updates the GraphQL database, as well as sends a broadcast to subscribed clients with an updated SHA value. The code in this PR is server side only.

### DIFF
--- a/app/graphql/types/subscription_type.rb
+++ b/app/graphql/types/subscription_type.rb
@@ -14,4 +14,8 @@ class Types::SubscriptionType < Types::BaseObject
   field :announcement_published, Types::AnnouncementType, description: "An announcement was published"
 
   def announcement_published; end
+
+  field :deployment_sha, String, description: "Updated SHA value"
+
+  def deployment_sha; end
 end
--- a/app/workers/sha_poller_worker.rb
+++ b/app/workers/sha_poller_worker.rb
@@ -1,0 +1,17 @@
+# Polls deployment sha for subscribed clients
+class ShaPollerWorker
+  include Sidekiq::Worker
+  include Sidekiq::Repeat::Repeatable
+
+  repeat { hourly.minute_of_hour(0, 5, 10, 15, 20, 25, 30, 35, 40, 45, 50, 55) }
+
+  def perform
+    logger.info "Running sha poll"
+
+    Helios2Schema.subscriptions.trigger(
+      "deploymentSha",
+      {},
+      AppEnv['HEROKU_SLUG_COMMIT'] || "N/A"
+    )
+  end
+end
--- a/spec/workers/sha_poller_worker_spec.rb
+++ b/spec/workers/sha_poller_worker_spec.rb
@@ -1,0 +1,22 @@
+require 'rails_helper'
+
+RSpec.describe ShaPollerWorker, type: :worker do
+  before {
+    allow(AppEnv).to receive(:[]).with('HEROKU_SLUG_COMMIT') { nil }
+  }
+
+  describe "SHA Poll Worker" do
+    context "should publish the deployed SHA value" do
+      it "will publish deployed SHA when triggered" do
+        expect(Helios2Schema.subscriptions).to(
+          receive(:trigger).with(
+            "deploymentSha",
+            {},
+            "N/A"
+          )
+        )
+        ShaPollerWorker.new.perform
+      end
+    end
+  end
+end
--- a/app/javascript/schema.json
+++ b/app/javascript/schema.json
@@ -3072,6 +3072,24 @@
               "deprecationReason": null
             },
             {
+              "name": "deploymentSha",
+              "description": "Updated SHA value",
+              "args": [
+
+              ],
+              "type": {
+                "kind": "NON_NULL",
+                "name": null,
+                "ofType": {
+                  "kind": "SCALAR",
+                  "name": "String",
+                  "ofType": null
+                }
+              },
+              "isDeprecated": false,
+              "deprecationReason": null
+            },
+            {
               "name": "eventPublished",
               "description": "An event was published to the API",
               "args": [
--- a/schema.graphql
+++ b/schema.graphql
@@ -128,6 +128,9 @@ type Subscription {
   # An announcement was published
   announcementPublished: Announcement!
 
+  # Updated SHA value
+  deploymentSha: String!
+
   # An event was published to the API
   eventPublished: Event!
 


app/workers/sha_poller_worker.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  repeat { hourly.minute_of_hour(0, 5, 10, 15, 20, 25, 30, 35, 40, 45, 50, 55) }
+
+  def perform
+    logger.info "Running sha poll"
+
+    Helios2Schema.subscriptions.trigger(
+      "schemaShaPublished",
+      {},
+      AppEnv['HEROKU_SLUG_COMMIT']
```
chewnoill, line 14: Let's make sure to update the `env.sample` file since we're depending on a new env variable now.

app/graphql/types/subscription_type.rb
```diff
@@ -14,4 +14,8 @@ def weather_published(latitude:, longitude:); end
   field :announcement_published, Types::AnnouncementType, description: "An announcement was published"
 
   def announcement_published; end
+
+  field :schema_sha, String, description: "Updated SHA value"
+
+  def schema_sha; end
```
chewnoill, line 20: there is something happening I don't understand. How is this subscription field connected to the poll worker?

app/graphql/types/subscription_type.rb
```diff
@@ -14,4 +14,8 @@ def weather_published(latitude:, longitude:); end
   field :announcement_published, Types::AnnouncementType, description: "An announcement was published"
 
   def announcement_published; end
+
+  field :schema_sha, String, description: "Updated SHA value"
```
cbarber, line 18: `schema` is misleading since this is not the database version. Its the deployment sha or application sha.

app/workers/sha_poller_worker.rb
```diff
@@ -0,0 +1,17 @@
...
+  include Sidekiq::Worker
+  include Sidekiq::Repeat::Repeatable
+
+  repeat { hourly.minute_of_hour(0, 5, 10, 15, 20, 25, 30, 35, 40, 45, 50, 55) }
+
+  def perform
+    logger.info "Running sha poll"
+
+    Helios2Schema.subscriptions.trigger(
+      "schemaShaPublished",
```
chewnoill, line 12: Oh I see, the trigger function must do a lookup into the schema subscriptions. Why do we add `Published` onto the end here? I would expect it to map more directly to the schema field name `schemaSha`.

chewnoill (commented):
Since we're not delivering a full vertical slice, can we include testing notes in the description?How was this tested?

app/workers/sha_poller_worker.rb
```diff
@@ -0,0 +1,17 @@
...
+      {},
+      AppEnv['HEROKU_SLUG_COMMIT']
+    )
```
tucker-weed, line 15: Does this create a connection with ActionCable? How often is this triggered/how does the trigger work?

tmsavage: No, this is a sidekiq worker on the server side. The connection is continuous through a web socket between the server and its clients. The data value in the variable `HEROKU_SLUG_COMMIT` is broadcasted to the clients every 5 minutes.

app/graphql/types/subscription_type.rb
tmsavage: Yes, but doesn't this field statement produce a subscription type in the GraphQL database? I agree it is the application sha on the server side, so renaming it to something like `deployment_sha` makes sense.

app/workers/sha_poller_worker.rb
cbarber: I'm on the fence with this. `HEROKU_SLUG_COMMIT` should only exist in the heroku instance. I guess we'd make a blank example in case a developer wanted to test? I don't think its worthwhile for us to integrate git into rails to look up just for local sha updates.

cbarber:
> Since we're not delivering a full vertical slice, can we include testing notes in the description?How was this tested?

Testing is a great idea. I think a smoke test (no exceptions thrown) should suffice for a blank/nil value and a non-empty value. Have a look at `spec/workers/weather_poller_worker_spec.rb` for an example of how to test a worker that publishes an event to a subscription.

tmsavage:
> > Since we're not delivering a full vertical slice, can we include testing notes in the description?How was this tested?
> 
> Testing is a great idea. I think a smoke test (no exceptions thrown) should suffice for a blank/nil value and a non-empty value. Have a look at `spec/workers/weather_poller_worker_spec.rb` for an example of how to test a worker that publishes an event to a subscription.

Yes, I agree. @bencates also mentioned creating unit tests last Friday. I will look into this!

app/workers/sha_poller_worker.rb
tmsavage: I initially wrote `published` at the end to keep consistency with the other subscription types and thought it was a proper formatting when working off `weather_poller_worker.rb`. I changed the field name from `schemaSha` to `deployment_sha` since it is the "deployment" sha, per Craig's note above.

chewnoill: I found the documentation for this here: https://graphql-ruby.org/subscriptions/triggers.html

> Events are triggered by name, and the name must match fields on your Subscription Type

fallynlogan (commented):
What is the SHA and why do we need to update it?

tmsavage:
> What is the SHA and why do we need to update it?

A SHA is the hash value that represents each commit. The SHA value is updated when a new commit, or update, has been made to the application.

cbarber (approved):
Nice work :+1: Just needs a rebase.

spec/workers/sha_poller_worker_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rails_helper'
+
+RSpec.describe ShaPollerWorker, type: :worker do
+  before {
+    allow(AppEnv).to receive(:[]).with('HEROKU_SLUG_COMMIT') { nil }
+  }
+
+  describe "a poller" do
+    context "when deployed sha is different from a known sha" do
+      it "will publish the response when performing" do
```
chewnoill, line 10: nit: test description is a little confusing.

Maybe something like this would be more accurate?:
 
SHA Poll worker
should publish the current SHA
when triggered

chewnoill (approved):
Small nit about some words, not blocking

spec/workers/sha_poller_worker_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rails_helper'
+
+RSpec.describe ShaPollerWorker, type: :worker do
+  before {
+    allow(AppEnv).to receive(:[]).with('HEROKU_SLUG_COMMIT') { nil }
+  }
+
```
sandipa12, line 7: what does this do? why have you passed 'HEROKU_SLUG_COMMIT' as a parameter ?

tmsavage: For the unit test, we are assuming AppEnv has the function `[]` with the variable name `'HEROKU_SLUG_COMMIT'`, as you would call `AppEnv['HEROKU_SLUG_COMMIT']`. It then expects a `nil` (null) value. We're basically telling the AppEnv object to return nil for the 'HEROKU_SLUG_COMMIT' variable.

PR #249, opened 2021-06-14
Feat: add bambooHR API integration

fallynlogan:
Proof of concept for BambooHR API to make sure that we can access all of the BambooHR fields required to display who's out of the office , who has birthdays, and who has anniversaries. 

Run `gem install bamboozled` to test

**Ticket information: **
Investigate the bamboo API calls and types to display employee birthdays
Investigate the bamboo API calls and types to display employee anniversaries (start of employment)
Investigate the bamboo API calls and types to display employee PTO
Investigate any existing ruby gems to support calling the API and/or serializing to ruby objects 

All of these were investigated in order to accomplish a design similar to this one: 
Screen Shot 2021-06-14 at 1 49 20 PM


app/lib/clients/bambooHR_client.rb
```diff
@@ -0,0 +1,45 @@
+require 'bamboozled'
+require 'date'
+
+
+class Clients::BambooHRClient
+    client = Bamboozled.client(subdomain: AppEnv['BAMBOOHR_SUBDOMAIN'], api_key: AppEnv['BAMBOOHR_API_KEY'])
+    # birthday = client.employee.all("birthday")
+    # anniversary = client.employee.all("hireDate")
+    day_before_yesterday = (Date.today - 2).to_formatted_s(:iso8601)
```
cbarber, line 9: Nit: we should probably start calculating things like today/tomorrow/yesterday based on a Location's timezone. In that case, we should probably pass a timezone to a method that uses that to calculate these values. Rails has some builtin methods for that: https://api.rubyonrails.org/v5.1.7/classes/ActiveSupport/TimeZone.html#method-i-yesterday

app/lib/clients/bambooHR_client.rb
```diff
@@ -0,0 +1,45 @@
...
+    # puts anniversary
+    puts "time off: "+time_off.to_s
+
+
+    employees_info = client.report.custom(["id", "birthday", "hireDate", "isPhotoUploaded"], format = "JSON")
+    employees = []
+
+    for employee in employees_info
+        #and employee['birthday'] != "" included because 2 of the employees in the test company do not have birthdays listed
+        if day_before_yesterday.include? employee['birthday'] or yesterday.include? employee['birthday'] or today.include? employee['birthday'] or tomorrow.include? employee['birthday'] or day_afer_tomorrow.include? employee['birthday'] and employee['birthday'] != ""
```
cbarber, line 26: Things like this could made a little more functional with Enumerable methods:
```
if employee['birthday'].present? &&
     [day_before_yesterday, yesterday, today, tomorrow].any? { |d| 
       d.include? employee['birthday']
     }
```
https://ruby-doc.org/core-2.7.2/Enumerable.html#method-i-any-3F

app/lib/clients/bambooHR_client.rb
```diff
@@ -0,0 +1,45 @@
...
+            employees.append(employee)
+        end
+        #employee['hireDate'].include? '8-13' was included because no employees were hired during the current dates so I added a date where I knew employees were hired on to test that this was working correctly
+        if day_before_yesterday.include? employee['hireDate'] or yesterday.include? employee['hireDate'] or today.include? employee['hireDate'] or tomorrow.include? employee['hireDate'] or day_afer_tomorrow.include? employee['hireDate'] or employee['hireDate'].include? '8-13' 
+            puts "hireDate: "+employee['hireDate'] 
+            employees.append(employee)
+        end
+        time_off.each{ |i| if i['employeeId'].to_s.include? employee['id']
+            puts "employees out: "+employee['id']
+            employees.append(employee)
```
cbarber, line 37: iterating/appending could also be done with map/compact:
https://ruby-doc.org/core-2.5.7/Enumerable.html#method-i-map
https://ruby-doc.org/core-2.5.7/Array.html#method-i-compact

app/lib/clients/bambooHR_client.rb
```diff
@@ -0,0 +1,45 @@
+require 'bamboozled'
+require 'date'
+
+
+class Clients::BambooHRClient
```
cbarber, line 5: Should have a method in here rather than class parse time code.

app/lib/clients/bambooHR_client.rb
```diff
@@ -0,0 +1,45 @@
+require 'bamboozled'
```
cbarber, line 1: :thinking: This should be required automatically as part of bundler/rails

app/lib/clients/bambooHR_client.rb
```diff
@@ -0,0 +1,36 @@
+def find_employees
+  client = Bamboozled.client(subdomain: AppEnv['BAMBOOHR_SUBDOMAIN'], api_key: AppEnv['BAMBOOHR_API_KEY'])
+
+  day_before_yesterday = (Time.zone.yesterday - 1).to_formatted_s(:iso8601)
+  yesterday = Time.zone.yesterday.to_formatted_s(:iso8601)
+  today = Time.zone.now.to_formatted_s(:iso8601)
+  tomorrow = Time.zone.tomorrow.to_formatted_s(:iso8601)
+  day_afer_tomorrow = (Time.zone.tomorrow + 1).to_formatted_s(:iso8601)
+
```
sandipa12, line 9: Is the information about birthdays, anniversaries & so on displayed only for those days or can you also access for the future/past days?

app/lib/clients/bambooHR_client.rb
```diff
@@ -0,0 +1,36 @@
...
+    if employee['hireDate'].present? &&
+          [day_before_yesterday, yesterday, today, tomorrow, day_afer_tomorrow].any? { |d| 
+          if d.include? employee['hireDate']
+            employees.append(employee)
+          end
+          }
+    end
+    time_off.map { |e|
+      if e['employeeId'].to_s.include? employee['id']
+        employees.append(employee)
```
tmsavage, line 31: So you are appending employees with recent/upcoming hire/birthdays. What happens to this array once you get all the employees? Does this reset each day?

app/lib/clients/bambooHR_client.rb
fallynlogan: You can access future/past days by changing the date fields (ie today, tomorrow...) I do not believe there is a way to search for the birthday and anniversaries by inputing the date that you want to match on using the API (for example something like this `client.employee.all("hireDate,05-17") ` wouldn't work) but you can do that for the time off field like this `client.time_off.whos_out(Time.now, "2014-12-31")`

app/lib/clients/bamboohr_client.rb
```diff
@@ -0,0 +1,8 @@
+class Clients::BamboohrClient
+  def self.find_employees
+    client = Bamboozled.client(subdomain: AppEnv['BAMBOOHR_SUBDOMAIN'], api_key: AppEnv['BAMBOOHR_API_KEY'])
```
chewnoill, line 3: make sure to update the env.sample and readme

app/lib/clients/create_employees.rb
```diff
@@ -0,0 +1,54 @@
...
+  def create_dates
+    day_before_yesterday = (Time.zone.yesterday - 1).to_formatted_s(:iso8601)
+    yesterday = Time.zone.yesterday.to_formatted_s(:iso8601)
+    today = Time.zone.now.to_formatted_s(:iso8601)
+    tomorrow = Time.zone.tomorrow.to_formatted_s(:iso8601)
+    day_afer_tomorrow = (Time.zone.tomorrow + 1).to_formatted_s(:iso8601)
+    [day_before_yesterday, yesterday, today, tomorrow, day_afer_tomorrow]
+  end
+
+  def find_birthday_employees
```
chewnoill, line 17: can we connect these functions to the graphql resolvers?

app/lib/clients/create_employees.rb
```diff
@@ -0,0 +1,54 @@
...
+    birthdays = []
+    @employees.each do |employee|
```
cbarber, line 19: This `.each` / `.if` / `.append` looks like a `select`: https://ruby-doc.org/core-2.5.7/Enumerable.html#method-i-select

`[1,2,3,4,5].select { |num|  num.even?  }   #=> [2, 4]`

app/lib/clients/create_employees.rb
```diff
@@ -0,0 +1,54 @@
...
+    birthdays
+  end
+
+  def find_anniversary_employees
+    anniversaries = []
+    @employees.each do |employee|
+      if employee['hireDate'].present? &&
+         @dates.any? { |d|
+           d.include? employee['hireDate']
+         }
```
cbarber, line 36: We repeat this `@dates.any?` block. Could we extract and name it something like `dates_contains(...)` or `this_week_contains(...)`?

app/lib/clients/create_employees.rb
fallynlogan: How would I go about doing that?

app/lib/clients/bamboohr_client.rb
fallynlogan: Currently we are working with free trial information and API keys that expire a week from setup. We are going to set up a new company that has an API key and has the information that we need for Mojo employees.  Should we wait until that is done before updating the readme? Otherwise I think we would need to update it every time we start a new trial. Would putting the information that I am currently using (API key and company domain) in this PR description be better?

chewnoill: Adding placeholders to `.env.sample` is fine

```
BAMBOOHR_SUBDOMAIN=<ask a team mate>
```
or similar is fine. Sign posting requirements, and their purpose, in the readme is nice also.

app/lib/clients/create_employees.rb
chewnoill: Something like this https://github.com/mojotech/helios2/commit/a1e79cdab283dc8afa485a65b4af1bcf255682f0?

app/lib/clients/bamboohr_client.rb
fallynlogan: Okay thanks!

app/lib/clients/create_employees.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+    anniversaries
+  end
+
+  def find_time_off_employees
+    whos_out = []
+    @employees.each do |employee|
+      if @time_off.any? { |e|
+        e['employeeId'].to_s.include? employee['id']
+      }
```
tucker-weed, line 48: What is the difference between the employee object and the e object? Would it be possible to only iterate over @time_off here?

bbazyk:
@fallynlogan  are you also going to code your design for the widget in javascript as part of this ticket? Or maybe we could add one on pivotal tracker as a next step for this project (I just don't see it listed yet)

chewnoill (commented):
LGTM let's start thinking about how to connect this to our graph. Doesn't really seem like employee data really belongs with the other events, but we still want to limit responses to between dates. What do you think about something like this?

```gql
type Query {
  dayEvents(startDate: Date, endDate: Date): [DayEvent!]!
}

type DayEvent {
  date: Date!
  birthdays: [EmployeeEvent!]!
  anniversaries: [EmployeeEvent!]!
  time_off: [EmployeeEvent!]!
}

type EmployeeEvent {
  id: ID!
  avatar: String!
}

scalar Date
```

fallynlogan:
@bbazyk I will not be doing that as part of this ticket, so it could definitely be made into its own! I can add it!

fallynlogan:
@chewnoill I think that looks good! Just to make sure I would want to add these under the types folder correct?

app/lib/clients/create_employees.rb
fallynlogan: The e object refers to the employees that are in the time off array, and the employee object is one of all of the employees. I think we would need to iterate through all of the employees in the company as well as the ones in the time off array in order to find a match between the two groups. Do you have an example of what you mean?

chewnoill:
looks like this needs a rebase and a rubocop fix

spec/graphql/bamboohr_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+]
+
+
+describe Helios2Schema do
+  let(:context) { {} }
+  let(:variables) { {} }
+
+  context 'from anonymous' do 
+
+    it 'will return valid employee data' do
```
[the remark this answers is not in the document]
chewnoill: Do we need to include tests for the other functions as well? `find_time_off_employees`, `find_birthday_employees` would look pretty similar I think.

app/lib/clients/create_employees.rb
tucker-weed: I was thinking perhaps performance could be improved by putting @employees data into a separate Map data structure for constant time lookup of @employees, and then you would only have to iterate over @time_off. I'm realizing now though that it seems like `employeeId` and `id` are not the same value because you use .include? to compare them?

spec/graphql/bamboohr_spec.rb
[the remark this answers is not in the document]
fallynlogan: Sounds good, that makes sense. I am currently writing tests for all of the functions, do you think that would be redundant?

app/lib/clients/create_employees.rb
```diff
@@ -0,0 +1,43 @@
...
+  def initialize(employees, time_off)
+    @employees = employees
+    @time_off = time_off
+    @dates = create_dates
+  end
+
+  def create_dates
+    dates = []
+    (-2..2).each do |n|
+      dates.append((Time.zone.today + n).to_formatted_s(:iso8601))
```
chewnoill, line 11: 😎

chewnoill (approved):

spec/graphql/bamboohr_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+TIMEOFFDATA = [
+  { "id" => 1501, "type" => "timeOff", "employeeId" => 5, "name" => "Ashley Adams",
+    "start" => Time.zone.today.to_formatted_s(:iso8601), "end" => Time.zone.tomorrow.to_formatted_s(:iso8601) }
+].freeze
+
+describe Helios2Schema do
+  let(:context) { {} }
+  let(:variables) { {} }
+
+  context 'from anonymous' do
```
cbarber, line 25: Lets name this.

spec/graphql/bamboohr_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    "birthday" => Time.zone.now.strftime("%m-%d"), "isPhotoUploaded" => "true",
+    :photoUrl => "http://blahbambooapitestblah.bamboohr.com/employees/photos/?h=206a1f7e6807f2dac63a7611dfce8453" }
+].freeze
+
+TIMEOFFDATA = [
+  { "id" => 1501, "type" => "timeOff", "employeeId" => 5, "name" => "Ashley Adams",
+    "start" => Time.zone.today.to_formatted_s(:iso8601), "end" => Time.zone.tomorrow.to_formatted_s(:iso8601) }
+].freeze
+
+describe Helios2Schema do
```
cbarber, line 21: Nit: we are describing `Clients::CreateEmployees` in our spec.

cbarber (changes requested):
Forgot to mark as requested changes

spec/graphql/bamboohr_spec.rb
fallynlogan: This should be fixed now!

spec/graphql/bamboohr_spec.rb
fallynlogan: This should be fixed now!

app/lib/clients/bamboohr_client.rb
```diff
@@ -0,0 +1,18 @@
...
+    Clients::CreateEmployees.new(create_time_off, create_employees)
+  end
+
+  def self.create_client
+    Bamboozled.client(subdomain: AppEnv['BAMBOOHR_SUBDOMAIN'], api_key: AppEnv['BAMBOOHR_API_KEY'])
+  end
+
+  def self.create_employees
+    employees = create_client.employee.all(%w[id displayName hireDate birthday isPhotoUploaded])
+    employees.each { |employee| employee.merge!(photoUrl: create_client.employee.photo_url(employee['id'])) }
```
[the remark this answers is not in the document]
fallynlogan: This should be fixed now!

cbarber (approved):
Nice job! :+1:

PR #250, opened 2021-06-14
[fix] location response rails caching error fix

tucker-weed:
Description: Fixes an issue where only primary location location data was being served to the client by providing a proper rails cache key for queried location data.


chewnoill (approved):
🚀

PR #252, opened 2021-06-15
chore: WidgetController refactor

tucker-weed:
## Description

1. Adds display components for the WidgetController
2. Refactors WidgetController to follow functional component pattern
3. Adds snapshot tests to assert the WidgetDisplay renders properly for correct input query data
4. Adds tests to assert that the widget transition control state selectors output the correct values given test states

### DIFF
--- a/app/javascript/components/__snapshots__/widget-display.test.jsx.snap
+++ b/app/javascript/components/__snapshots__/widget-display.test.jsx.snap
@@ -1,0 +1,73 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`DisconnectedDisplay component renders correctly 1`] = `
+<styled.div>
+  <FullPanel>
+    <DisconnectedMessage />
+  </FullPanel>
+  <SidePanel
+    cityName="Providence"
+    isPaused={false}
+    selectedWidgetId={0}
+    showWeather={false}
+    totalTime={0}
+    widgets={Array []}
+  />
+</styled.div>
+`;
+
+exports[`LoadingDisplay component renders correctly 1`] = `
+<styled.div>
+  <FullPanel>
+    <LoadingMessage />
+  </FullPanel>
+  <SidePanel
+    cityName="Providence"
+    isPaused={false}
+    selectedWidgetId={0}
+    showWeather={false}
+    totalTime={0}
+    widgets={Array []}
+  />
+</styled.div>
+`;
+
+exports[`WidgetDisplay component renders correctly 1`] = `
+<styled.div
+  onClick={[Function]}
+  onKeyDown={[Function]}
+  tabIndex="0"
+>
+  <FullPanel>
+    <getWeatherQuery
+      cityName="Providence"
+    />
+  </FullPanel>
+  <SidePanel
+    cityName="Providence"
+    isPaused={false}
+    selectedWidgetId={2}
+    showWeather={false}
+    totalTime={20000}
+    widgets={
+      Array [
+        Object {
+          "id": 2,
+          "name": "Weather",
+          "sidebarText": "Weather",
+        },
+        Object {
+          "id": 3,
+          "name": "Twitter",
+          "sidebarText": "@MojoTech",
+        },
+        Object {
+          "id": 4,
+          "name": "Numbers",
+          "sidebarText": "MojoTech by the Numbers",
+        },
+      ]
+    }
+  />
+</styled.div>
+`;
--- a/app/javascript/components/widget-controller.jsx
+++ b/app/javascript/components/widget-controller.jsx
@@ -1,13 +1,13 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import gql from 'graphql-tag';
-import { mathMod } from 'ramda';
 import { Query } from 'react-apollo';
 import {
   WidgetDisplay,
   LoadingDisplay,
   DisconnectedDisplay,
 } from '@components/widget-display';
+import WidgetTransitionControls from '@components/widget-transition-controls';
 
 const getWidgets = gql`
   query getWidgets($id: Int!, $cityName: String!) {
@@ -30,178 +30,59 @@ const getWidgets = gql`
   }
 `;
 
-export class WidgetController extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      currentWidgetId: 0,
-      transitionTime: null,
-    };
-  }
+export const WidgetController = ({ client, cityName }) => {
+  const [newWidgetId, setNewWidgetId] = useState(0);
 
-  componentDidMount() {
-    const prefetchDelay = 1500;
+  const fetchPolicy = newWidgetId === 0 ? 'network-only' : 'cache-first';
 
-    this.lastPrefetchTime = 0;
-    this.timerId = setInterval(() => {
-      const { transitionTime, pausedTime } = this.state;
-
-      if (!transitionTime || pausedTime) {
-        return;
-      }
-
-      const { client } = this.props;
-      const prefetchTime = transitionTime - prefetchDelay;
-      const now = Date.now();
-
-      if (transitionTime < now) {
-        this.nextWidget();
-      } else if (
-        prefetchTime < now &&
-        this.lastPrefetchTime < now - prefetchDelay
-      ) {
-        const { nextWidgetId } = this.state;
-        const { cityName } = this.props;
-        this.lastPrefetchTime = now;
-        client.query({
-          query: getWidgets,
-          variables: {
-            id: nextWidgetId,
-            cityName,
-          },
-        });
-      }
-    }, 500);
-  }
-
-  componentWillUnmount() {
-    if (this.timerId) {
-      clearInterval(this.timerId);
-    }
-  }
-
-  keyToWidgetAction = ({ keyCode }) => {
-    const upKey = 38;
-    const downKey = 40;
-    const SPACEBAR = 32;
-    if (keyCode === SPACEBAR) {
-      this.setState(({ pausedTime, transitionTime }) =>
-        pausedTime
-          ? {
-              pausedTime: null,
-              transitionTime: transitionTime - pausedTime + Date.now(),
-            }
-          : { pausedTime: Date.now() },
-      );
-    }
-    if (keyCode === downKey) {
-      this.nextWidget();
-    }
-    if (keyCode === upKey) {
-      this.prevWidget();
-    }
-  };
-
-  nextWidget = () => {
-    this.setState(({ nextWidgetId }) =>
-      nextWidgetId
-        ? {
-            currentWidgetId: nextWidgetId,
-            nextWidgetId: null,
-            prevWidgetId: null,
-            transitionTime: null,
-          }
-        : {},
-    );
-  };
-
-  prevWidget = () => {
-    this.setState(({ prevWidgetId }) =>
-      prevWidgetId
-        ? {
-            currentWidgetId: prevWidgetId,
-            nextWidgetId: null,
-            prevWidgetId: null,
-            transitionTime: null,
-          }
-        : {},
-    );
-  };
-
-  switchToPage = id => {
-    this.setState({
-      currentWidgetId: id,
-      nextWidgetId: null,
-      prevWidgetId: null,
-      transitionTime: null,
+  const queryCallback = nextWidgetId => {
+    client.query({
+      query: getWidgets,
+      variables: {
+        id: nextWidgetId,
+        cityName,
+      },
     });
   };
 
-  clickToWidgetAction = ({ target }) => {
-    const targetId = target.id;
-    if (targetId.startsWith('widget_')) {
-      const widgetId = parseInt(targetId.replace('widget_', ''), 10);
-      this.switchToPage(widgetId);
-    }
-  };
+  return (
+    <Query
+      query={getWidgets}
+      variables={{ id: newWidgetId, cityName }}
+      fetchPolicy={fetchPolicy}
+      onCompleted={response => {
+        const {
+          location: {
+            widgets: { byIdOrFirst: current },
+          },
+        } = response;
+        setNewWidgetId(current.id);
+      }}
+    >
+      {({ loading, error, data }) => {
+        if (loading) {
+          return <LoadingDisplay cityName={cityName} />;
+        }
 
-  render() {
-    const { currentWidgetId, pausedTime } = this.state;
+        if (error) {
+          // eslint-disable-next-line
+          console.error(error);
+          return <DisconnectedDisplay cityName={cityName} />;
+        }
 
-    const fetchPolicy = currentWidgetId === 0 ? 'network-only' : 'cache-first';
-
-    const { cityName } = this.props;
-
-    return (
-      <Query
-        query={getWidgets}
-        variables={{ id: currentWidgetId, cityName }}
-        fetchPolicy={fetchPolicy}
-        onCompleted={response => {
-          const { location } = response;
-          const { widgets } = location;
-          const { enabled: enabledWidgets, byIdOrFirst: current } = widgets;
-
-          const index = enabledWidgets.findIndex(w => w.id === current.id);
-
-          const prevWidgetId =
-            enabledWidgets[mathMod(index - 1, enabledWidgets.length)].id;
-          const nextWidgetId =
-            enabledWidgets[mathMod(index + 1, enabledWidgets.length)].id;
-
-          this.setState({
-            currentWidgetId: current.id,
-            nextWidgetId,
-            prevWidgetId,
-            transitionTime: Date.now() + current.durationSeconds * 1000,
-          });
-        }}
-      >
-        {({ loading, error, data }) => {
-          if (loading) {
-            return <LoadingDisplay cityName={cityName} />;
-          }
-
-          if (error) {
-            // eslint-disable-next-line
-            console.error(error);
-            return <DisconnectedDisplay cityName={cityName} />;
-          }
-
-          return (
-            <WidgetDisplay
-              keyToWidgetAction={this.keyToWidgetAction}
-              clickToWidgetAction={this.clickToWidgetAction}
-              isPaused={!!pausedTime}
-              cityName={cityName}
-              widgets={data.location.widgets}
-            />
-          );
-        }}
-      </Query>
-    );
-  }
-}
+        return (
+          <WidgetTransitionControls
+            WrappedComponent={WidgetDisplay}
+            requestPrefetch={queryCallback}
+            requestWidget={setNewWidgetId}
+            cityName={cityName}
+            widgets={data.location.widgets}
+          />
+        );
+      }}
+    </Query>
+  );
+};
 
 WidgetController.propTypes = {
   client: PropTypes.shape({
--- a/app/javascript/components/widget-controller.jsx
+++ b/app/javascript/components/widget-controller.jsx
@@ -3,14 +3,11 @@ import PropTypes from 'prop-types';
 import gql from 'graphql-tag';
 import { mathMod } from 'ramda';
 import { Query } from 'react-apollo';
-import styled from 'styled-components';
-import FullPanel from '@components/full-panel';
-import SidePanel from '@components/side-panel';
-import { DisconnectedMessage, LoadingMessage } from '@messages/message';
-import Twitter from '@widgets/twitter';
-import Numbers from '@widgets/numbers';
-import Weather from '@widgets/weather';
-import Traffic from '@widgets/traffic';
+import {
+  WidgetDisplay,
+  LoadingDisplay,
+  DisconnectedDisplay,
+} from '@components/widget-display';
 
 const getWidgets = gql`
   query getWidgets($id: Int!, $cityName: String!) {
@@ -32,17 +29,6 @@ const getWidgets = gql`
     }
   }
 `;
-
-const Wrapper = styled.div`
-  outline: none;
-`;
-
-const widgetElements = {
-  Weather,
-  Twitter,
-  Numbers,
-  Traffic,
-};
 
 export class WidgetController extends React.Component {
   constructor(props) {
@@ -94,7 +80,7 @@ export class WidgetController extends React.Component {
     }
   }
 
-  switchPages = ({ keyCode }) => {
+  keyToWidgetAction = ({ keyCode }) => {
     const upKey = 38;
     const downKey = 40;
     const SPACEBAR = 32;
@@ -151,7 +137,7 @@ export class WidgetController extends React.Component {
     });
   };
 
-  handleClicks = ({ target }) => {
+  clickToWidgetAction = ({ target }) => {
     const targetId = target.id;
     if (targetId.startsWith('widget_')) {
       const widgetId = parseInt(targetId.replace('widget_', ''), 10);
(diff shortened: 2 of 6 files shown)


cbarber (changes requested):
Full disclaimer, I'm not convinced react hooks aren't an anti-pattern. I appreciate you've kept their use to a pretty small sub-section. The extraction and naming of the *Display components is great and I think deserves it own commit to help reviewers assert the hooks are an accurate representation of the class component callbacks.

This might be a good opportunity to have some snapshot tests to assert a refactor hasn't introduced unexpected changes to the DOM.

app/javascript/components/displays.jsx
```diff
@@ -0,0 +1,214 @@
...
+      const now = Date.now();
+
+      if (transitionTime < now) {
+        setNextWidgetId();
+      } else if (
+        prefetchTime < now &&
+        lastPrefetchTime.current < now - prefetchDelay
+      ) {
+        lastPrefetchTime.current = now;
+        setNextWidgetId();
```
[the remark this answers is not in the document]
tucker-weed: Is it lost because lastPrefetchTime is defined in the WidgetDisplay and not the WidgetController (so re renders of WidgetController reset lastPrefetchTime)?

app/javascript/components/displays.jsx
```diff
@@ -0,0 +1,214 @@
...
+      cityName={cityName}
+    />
+  </Wrapper>
+);
+
+LoadingDisplay.propTypes = {
+  cityName: PropTypes.string.isRequired,
+};
+
+export const DisconnectedDisplay = ({ cityName }) => (
```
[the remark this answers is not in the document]
tucker-weed: All of the displays also depend on Wrapper which is defined in this file - should I just copy and paste wrapper into all of the new files or create a separate Wrapper component which can be imported?

app/javascript/components/displays.jsx
```diff
@@ -0,0 +1,214 @@
...
+        setNextWidgetId();
+      }
+    }, 500);
+
+    return () => {
+      if (timerId.current) {
+        clearInterval(timerId.current);
+      }
+    };
+  }, [superState]);
```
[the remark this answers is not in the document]
tucker-weed: The WidgetDisplay has information about enabledWidgets from data that is necessary for setting the next Widget. This is why the componentDidMount was moved to the WidgetDisplay.

Regarding your question, without the dependency superState the WidgetDisplay could be paused (by spacebar)  but the timer would still be running (so the widgets would still switch even if paused).

The interval for timerId is set up with the superState which is _passed_ from WidgetController - the superState remains the same in WidgetDisplay even though WidgetController state may have changed.  I suspect this means the superState is outdated in setInterval, and thus the interval needs to be reset with a new superState on every re render of WidgetController.

Is this ok or should it be changed?

app/javascript/components/displays.jsx
```diff
@@ -0,0 +1,214 @@
...
+  Weather,
+  Twitter,
+  Numbers,
+  Traffic,
+};
+
+const prefetchDelay = 1500;
+
+export const WidgetDisplay = ({
+  stateCallback,
```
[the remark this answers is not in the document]
tucker-weed: Would passing in a callback Object as a single argument to WidgetDisplay (like const callbacks = { setPrevWidgetId, setNextWidgetId, . . . } ) be an appropriate solution?

app/javascript/components/displays.jsx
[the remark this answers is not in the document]
cbarber: Sorry, I wasn't clear. I don't think there is value to copy/pasting (or event extracting) wrapper. I think its fine to keep them in a file when they are tightly coupled to the actual component.

app/javascript/components/displays.jsx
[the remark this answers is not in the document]
cbarber: Maybe? An object of functions my have gotchas with changing by reference and causing re-renders so it may require a memoization or just pass them as direct properties.

app/javascript/components/displays.jsx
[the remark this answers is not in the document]
cbarber: Its lost because we stopped calling `client.query(...)`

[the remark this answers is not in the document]
tucker-weed: Oh I see, I'll revert that back

app/javascript/components/displays.jsx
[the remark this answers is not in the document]
cbarber: Ok, that makes more sense but it seems like we introduced a new edge case. Our setInterval is destroyed and created on updates to state, not just component mount. This could mean destroying in interval that has partially run (say 250ms from start) and shift all subsequent checks. We may need to rethink approach or increase the frequency of our checks to account for the new edge case.

app/javascript/components/displays.jsx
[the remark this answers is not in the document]
tucker-weed: I'll pass them in as direct properties

app/javascript/components/displays.jsx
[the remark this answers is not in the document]
tucker-weed: Ok, sounds good

app/javascript/components/displays.jsx
```diff
@@ -0,0 +1,214 @@
...
+    >
+      <FullPanel>
+        <WidgetElement cityName={cityName} />
+      </FullPanel>
+      <SidePanel
+        widgets={enabledWidgets}
+        selectedWidgetId={superState.currentWidgetId}
+        showWeather={current.showWeather}
+        totalTime={current.durationSeconds * 1000}
+        isPaused={!!superState.pausedTime}
```
sandipa12, line 145: Why do you have !!superState?

tucker-weed: The field isPaused checks whether the Widget timer is currently paused for rendering purposes. !!superState.pausedTime casts superState.pausedTime to a boolean value (from either null, <a_number> to false, true respectively)

app/javascript/components/widget-display.jsx
```diff
@@ -0,0 +1,225 @@
...
+        changeValue: Date.now() + current.durationSeconds * 1000,
+      });
+      requestWidget(nextWidgetId);
+    }
+    if (keyCode === upKey) {
+      dispatch({
+        type: 'resetTimerValue',
+        changeValue: Date.now() + current.durationSeconds * 1000,
+      });
+      requestWidget(prevWidgetId);
```
tmsavage, line 101: Some of the lines in these cases seem a bit redundant, such as lines 92 and 99. In the original `widget-controller.jsx` file, these cases called a `nextWidget()` function. Do you think lines like these could be implemented into a callable function to reduce redundancy?

app/javascript/components/widget-controller.jsx
```diff
@@ -33,234 +29,58 @@ const getWidgets = gql`
...
+      fetchPolicy={fetchPolicy}
+      onCompleted={response => {
+        const {
+          location: {
+            widgets: { byIdOrFirst: current },
+          },
+        } = response;
+        setNewWidgetId(current.id);
+      }}
+    >
```
chewnoill, line 60: hmmmmm, this seems confusing. Why are we setting the newWidgetId onComplete? Is it possible that the widgetId that is returned the one that was selected? Will this cause the query to be fired twice on the first load?

Maybe what you want is the [useLazyQuery](https://www.apollographql.com/docs/react/api/react/hooks/#uselazyquery) hook?

app/javascript/components/widget-display.jsx
```diff
@@ -0,0 +1,225 @@
...
+    if (keyCode === SPACEBAR) {
+      if (pausedTime) {
+        dispatch({
+          type: 'extendTimer',
+          changeValue: Date.now() - pausedTime,
+        });
+      } else {
+        dispatch({ type: 'setPausedTime', changeValue: Date.now() });
+      }
+    }
```
chewnoill, line 121: if we're dispatching differently based on whats in state, its a sign that that this calculation should be inside the reducer. I think the action you want here is `togglePause`

```js
if (keyCode === SPACEBAR) {
  dispatch({type: 'togglePause'})
}
```

app/javascript/components/widget-display.jsx
tucker-weed: I saw your point and made a change where the resetTimerValue event is only called one time in the display in a useEffect hook dependent on new `widgets`, data. Let me know what you think.

app/javascript/components/widget-controller.jsx
tucker-weed: It does fire twice on the first load which I thought was strange.  Will look into useLazyQuery.

cbarber: > Why are we setting the newWidgetId onComplete?

Because on first load our widgetId is 0. Also a potential edge case (with the future admin page) is the queried widget could have been disabled, causing us to jump back to the first enabled widget.

app/javascript/components/widget-display.jsx
```diff
@@ -0,0 +1,216 @@
...
+  const { byIdOrFirst: current, enabled: enabledWidgets } = widgets;
+
+  const WidgetElement = widgetElements[current.name].Panel;
+
+  const index = enabledWidgets.findIndex(w => w.id === current.id);
+
+  const prevWidgetId =
+    enabledWidgets[mathMod(index - 1, enabledWidgets.length)].id;
+  const nextWidgetId =
+    enabledWidgets[mathMod(index + 1, enabledWidgets.length)].id;
```
cbarber, line 65: Thanks for taking the time to unravel this. Sometimes when we're working through a refactor, we discover other improvements that can be made in other sections of the application. In this case, it looks like we can't extract a `switchToNextWidget` and `switchToPrevWidget` callbacks because prev/next depend on the data from current widget state passed to the graphql query. A potential improvement is to expose `nextWidget(currWidget: Int)` and `prevWidget(currWidget: Int)` in the API itself. I'll make a ticket for this.

app/javascript/components/widget-controller.jsx
cbarber: > It does fire twice on the first load which I thought was strange. Will look into useLazyQuery.

I would recommend reading and understanding the code before exploring alternatives. You can't assert the alternative is correct if you don't understand the problem.

app/javascript/components/widget-controller.jsx
```diff
@@ -33,234 +29,58 @@ const getWidgets = gql`
...
+      query={getWidgets}
+      variables={{ id: newWidgetId, cityName }}
+      fetchPolicy={fetchPolicy}
+      onCompleted={response => {
+        const {
+          location: {
+            widgets: { byIdOrFirst: current },
+          },
+        } = response;
+        setNewWidgetId(current.id);
```
chewnoill, line 58: I'm still confused by this I think. If the API returns the `current.id`, why do we need to keep it in state?

app/javascript/components/widget-display.jsx
```diff
@@ -0,0 +1,207 @@
...
+
+  useEffect(() => {
+    dispatch({
+      type: 'startTimer',
+      durationSeconds: widgetTransitionTime,
+    });
+  }, [current.id]);
+
+  useEffect(() => {
+    timerId.current = setInterval(() => {
```
chewnoill, line 81:
```js
const timerId = ...
```
Since we're not using this ref outside of the effect we don't need to create a ref at all. We can store it in a const and clean it up in the return-callback.

chewnoill (approved):
🚀

app/javascript/components/widget-controller.jsx
```diff
@@ -1,16 +1,12 @@
...
-import { mathMod } from 'ramda';
 import { Query } from 'react-apollo';
-import styled from 'styled-components';
-import FullPanel from '@components/full-panel';
-import SidePanel from '@components/side-panel';
-import { DisconnectedMessage, LoadingMessage } from '@messages/message';
-import Twitter from '@widgets/twitter';
-import Numbers from '@widgets/numbers';
-import Weather from '@widgets/weather';
-import Traffic from '@widgets/traffic';
```
fallynlogan, line 13: Why would we want to remove all of the widgets from the controller?

tucker-weed: They moved to the WidgetDisplay - my thinking was that it would be better to group these Widget components with display functionality rather than with the query logic in the controller.

bbazyk:
Just a clarifying question on how we name our PRs : would this still be considered a chore after all of the additions made? Or yes since the functionality is still relatively the same?

tucker-weed:
> Just a clarifying question on how we name our PRs : would this still be considered a chore after all of the additions made? Or yes since the functionality is still relatively the same?

I believe the answer to that is yes because the functionality of the app is unchanged.

app/javascript/components/widget-display.test.jsx
```diff
@@ -0,0 +1,186 @@
...
+import { getMainDefinition } from 'apollo-utilities';
+import ActionCableLink from 'graphql-ruby-client/subscriptions/ActionCableLink';
+import { persistCache } from 'apollo-cache-persist';
+import helioSchema from '@javascript/schema.json';
+import { WidgetDisplay } from '@components/widget-display';
+import { ApolloProvider } from 'react-apollo';
+
+const cable = ActionCable.createConsumer('/cable');
+
+const httpLink = new HttpLink({
```
cbarber, line 21: Sorry I think I sent you down the wrong path here. A snapshot test should just render a component without connecting to any external resources. Lets pair to work through this.

cbarber:
The PR title is a summary of the changes in the branch. We onl care about the commits having identifiers like: chore, refactor, feat, fix.

app/javascript/components/widget-display.jsx
```diff
@@ -20,18 +21,111 @@ const widgetElements = {
...
+
+  useEffect(() => {
+    const timerId = setInterval(() => {
+      const { durationRemaining, start } = state;
+      if (isPaused(state)) {
+        return;
+      }
+
+      const prefetchTime = durationRemaining - prefetchDelay;
+      const now = Date.now();
```
cbarber, line 90: Something is going wrong and I suspect it stems from here. I noticed the rails server logs were getting more query requests for `getWidgets` after this change. Adding some logging:
```
diff --git a/app/javascript/components/widget-controller.jsx b/app/javascript/components/widget-controller.jsx
index 7715697..9dd5df6 100644
--- a/app/javascript/components/widget-controller.jsx
+++ b/app/javascript/components/widget-controller.jsx
@@ -35,6 +35,7 @@ export const WidgetController = ({ client, cityName }) => {
   const fetchPolicy = newWidgetId === 0 ? 'network-only' : 'cache-first';
 
   const queryCallback = nextWidgetId => {
+    console.log('caling getWidgets');
     client.query({
       query: getWidgets,
       variables: {
```
Shows a console log about every second fetching the widgets. Looking at the changes between the previous logic and here it appears we have switched to ms durations in our state where before we had Date.now() values which is an epoch style numeric time.

Looking at the names `start` and `durationRemaining` is a little misleading. Start actually starts with the ms remaining for the transition in the logic.

We could keep the original state and corresponding logic for this refactor. Or we could delve a little deeper extracting functions from the original logic with some tests coverage to refactor prior to class component to hooks.

cbarber (changes requested):
I marked the outdated old comments as addressed to help clean up the noise in the PR.

I found a bug. Lets meet up tomorrow to go over options.

app/javascript/components/widget-display.jsx
tucker-weed: @cbarber good catch - I cleaned it up a bit and fixed the bug. It looks like I forgot to alter the prefetch logic in the interval to work with the changes made to the timer state. It works now, and lastPrefetchTime isn't necessary anymore which is a bonus.

Regarding the naming of `start`, I can see why it was confusing: I accidentally gave it the `widgetTransitionTime` in the initial state when it should have been `Date.now()`. I also renamed `start` to be `startTimestamp` for less confusion.

Let me know what you think of the changes.

app/javascript/components/hocs/with-widget-transition-controls.jsx
```diff
@@ -0,0 +1,144 @@
...
+    const timerId = setInterval(() => {
+      const { startTimestamp } = state;
+      if (isPaused(state)) {
+        return;
+      }
+
+      const now = Date.now();
+      const timeElapsedSinceStart = now - startTimestamp;
+
+      if (shouldRequestWidget(state, timeElapsedSinceStart)) {
```
cbarber, line 81: Nit: `timeElapsedSinceStart` is a calculation from now and state. `shouldRequestWidget` / `shouldRequestPrefecth` are like selectors so I'd argue we should be passing in `now` and letting it do the `timeElapsedSinceStart` calculation from state. You could even have another selector just for that calculation.

cbarber (commented):
Nit: I don't think there is a lot of value separating with-widget-transition-controls from the components since it is very unlikely to be reused in the project. On the other hand, the file was already getting large so I could be convinced either way. Non-blocking.

app/javascript/components/__snapshots__/widget-display.test.jsx.snap
```diff
@@ -0,0 +1,73 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`DisconnectedDisplay component renders correctly 1`] = `
```
fallynlogan, line 3: Why add the '1' to the end of the test description test?

app/javascript/components/hocs/with-widget-transition-controls.jsx
```diff
@@ -0,0 +1,140 @@
+import React, { useEffect, useReducer } from 'react';
+import PropTypes from 'prop-types';
+import { mathMod } from 'ramda';
+
+const prefetchDelay = 1500;
+
+export const shouldRequestWidget = state =>
+  state.msDurationRemaining < Date.now() - state.startTimestamp;
```
fallynlogan: Is there a reason we would want to use a 'Date' object over a 'Time' object here?

app/javascript/components/widget-controller.jsx
```diff
@@ -33,234 +30,59 @@ const getWidgets = gql`
...
+    >
+      {({ loading, error, data }) => {
+        if (loading) {
+          return <LoadingDisplay cityName={cityName} />;
+        }
 
-          const WidgetElement = widgetElements[current.name].Panel;
+        if (error) {
+          // eslint-disable-next-line
+          console.error(error);
```
fallynlogan, line 69: Is there a reason we would want to display the error on the console as opposed to on the screen itself?

app/javascript/components/widget-transition-controls.jsx
```diff
@@ -0,0 +1,136 @@
+import React, { useEffect, useReducer } from 'react';
+import PropTypes from 'prop-types';
+import { mathMod } from 'ramda';
+
+const prefetchDelay = 1500;
+
+export const shouldRequestWidget = state =>
+  state.msDurationRemaining < Date.now() - state.startTimestamp;
```
fallynlogan, line 8: Is the "<" operator being used to make a comparison between `state.msDurationRemaining` and `Date.now() - state.startTimestamp;` and if so why?

fallynlogan (commented):
Just had a few questions but LGTM!

app/javascript/components/widget-transition-controls.jsx
tucker-weed: This check is used in the interval which handles transitioning between widgets once the timer has run out. `state.msDurationRemaining` is initially the full 20000 millisecond transition time - the comparison here is checking whether the time elapsed since `state.startTimestamp` exceeds  the duration remaining since the last start. The start timestamp and duration are updated if the timer is paused then unpaused.

Writing this comment out made me realize that `state.msDurationRemaining` should be called `state.msRemainingSinceLastStart` for clarity. I'll make those changes now.

app/javascript/components/widget-controller.jsx
tucker-weed: There is a disconnected message on error to cleanly show that something went wrong - the console.error is there to show the verbose output of errors for debugging purposes

app/javascript/components/hocs/with-widget-transition-controls.jsx
tucker-weed: What is a Time object?

app/javascript/components/widget-controller.jsx
tmsavage, line 69: Last week we discussed that your refactoring may not cause much of an issue, but I think this is something you and I should go over again after you merge since side-panel is no longer imported, causing a difference in the functionality of error and loading.

app/javascript/components/__snapshots__/widget-display.test.jsx.snap
tucker-weed: The description (in the snapshot test implementation file) doesn't have a 1 at the end of it. This is an auto generated file, so there is stuff happening behind the scenes that I can't really speak to.

app/javascript/components/widget-controller.jsx
tucker-weed: Yeah the SidePanel was moved to the WidgetDisplay - we should definitely coordinate resolving any merge conflicts at some point.

app/javascript/components/widget-transition-controls.jsx
```diff
@@ -0,0 +1,136 @@
...
+const prefetchDelay = 1500;
+
+export const shouldRequestWidget = state =>
+  state.msRemainingSinceLastStart < Date.now() - state.startTimestamp;
+
+export const shouldRequestPrefetch = state =>
+  state.msRemainingSinceLastStart - (Date.now() - state.startTimestamp) <
+  prefetchDelay;
+
+export const isPaused = state => state.startTimestamp === null;
```
chewnoill, line 14: nit: This pattern of thing, where you take the current state and you pull out a certain part is called a **selector**. We should call this `selectIsPaused`

tucker-weed: @chewnoill just implemented that change, and I also prepended 'select' to `shouldRequestPrefetch` and `shouldRequestWidget` as well for consistency.

app/javascript/components/widget-controller.jsx
chewnoill: Technically, we do both. A specific error is reported on the console here, and on the next line the `DisconnectedDisplay` component is presented on the screen.

cbarber (commented):
The nits are just there for information.

It looks like we still have a regression where we request `getWidgets` 5 times right before transitioning.

tucker-weed:
> The nits are just there for information.
> 
> It looks like we still have a regression where we request `getWidgets` 5 times right before transitioning.

@cbarber Looks like shouldRequestPrefetch() was always true when the timer is below the `prefetchTime`. I  added a check for `noPrefetchRequestedYet` to make sure that the prefetch is only requested once per widget and the problem went away. Let me know what you think of the solution. I also changed the test descriptions for clarity.

app/javascript/components/widget-transition-controls.jsx
```diff
@@ -0,0 +1,139 @@
+import React, { useEffect, useReducer } from 'react';
+import PropTypes from 'prop-types';
+import { mathMod } from 'ramda';
+
+const prefetchDelay = 1500;
+let noPrefetchRequestedYet = true;
```
chewnoill, line 6: can we keep this in state? Would be nice to keep all of our side effects in one place.

app/javascript/components/widget-transition-controls.jsx
```diff
@@ -0,0 +1,139 @@
...
+    });
+  }, [current.id]);
+
+  useEffect(() => {
+    const timerId = setInterval(() => {
+      if (selectIsPaused(state)) {
+        return;
+      }
+      if (selectShouldRequestWidget(state)) {
+        requestWidget(nextWidgetId);
```
chewnoill, line 76: maybe if we dispatch something like this:
```js
dispatch({type:'preload', func: requestWidget })
```

Then we could setup the action in our reducer like this:
```js
...
case 'preload':
  if(!state.preloaded) {
    action.func(selectNextWidgetId(state))
  }
  return {
    preloaded: true,
   ...state,
  }
...
```

tucker-weed: @chewnoill I implemented a dispatch option for 'prefetch', and also moved `nextWidgetId` and `prevWidgetId` computations into a resolver function.

app/javascript/components/widget-transition-controls.jsx
```diff
@@ -0,0 +1,166 @@
...
+import PropTypes from 'prop-types';
+import { mathMod } from 'ramda';
+
+const prefetchDelay = 1500;
+
+export const selectShouldRequestWidget = state =>
+  state.msRemainingSinceLastStart < Date.now() - state.startTimestamp;
+
+export const selectShouldRequestPrefetch = state =>
+  state.noPrefetchRequestedYet &&
```
cbarber, line 11: Nit: "keep it positive" is generally a good mantra for boolean names. `needsPrefetch`.

app/javascript/components/widget-transition-controls.test.jsx
```diff
@@ -0,0 +1,119 @@
...
+    expect(
+      selectShouldRequestPrefetch({
+        msRemainingSinceLastStart: 20000,
+        startTimestamp: Date.now() - 19000,
+        noPrefetchRequestedYet: true,
+      }),
+    ).toEqual(true);
+  });
+
+  it('returns false if ready to request prefetch but prefetch has been made already', () => {
```
cbarber, line 38: :100:

cbarber (approved):
Excellent work. Just a nit pick on naming. Up to you to address or not. :shipit:

PR #254, opened 2021-06-15
feat: display wind speed

sandipa12:
Displays daily and hourly wind speed on the weather page.
Screen Shot 2021-06-15 at 5 22 56 PM


tucker-weed:
Looks good. Maybe the wind speed units could be displayed as kph or mph? Also, what do you think about adding a **wind** tag to the speeds?

tmsavage:
> Looks good. Maybe the wind speed units could be displayed as kph or mph? Also, what do you think about adding a **wind** tag to the speeds?

I think these are good suggestions.

chewnoill:
closed in favor of https://github.com/mojotech/helios2/pull/256

PR #255, opened 2021-06-17
feat: client receives deployed sha value via subscriptions and performs comparison with current sharing for refresh

tmsavage:
This branch consists of the client-side code that queries the GraphQL database for the deployed SHA value broadcasted by the server. The client then makes the comparison of its current SHA value with the updated queried value. This comparison occurs every five minutes per the server's SHA poller worker. If the SHA values are the same, nothing happens. If they are different, then the page reloads to an updated version of Helios 2.

### DIFF
--- a/app/javascript/components/app.jsx
+++ b/app/javascript/components/app.jsx
@@ -14,6 +14,7 @@ import { getMainDefinition } from 'apollo-utilities';
 import ActionCableLink from 'graphql-ruby-client/subscriptions/ActionCableLink';
 import { persistCache } from 'apollo-cache-persist';
 import WidgetController from '@components/widget-controller';
+import versionCompare from '@components/version-compare';
 import helioSchema from '@javascript/schema.json';
 import GlobalStyle from '../styles';
 
@@ -72,6 +73,8 @@ const client = new ApolloClient({
     },
   },
 });
+
+versionCompare(client);
 
 const App = () => (
   <ApolloProvider client={client}>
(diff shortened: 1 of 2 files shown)


tmsavage:
Forgot to remove my console.log statements. I'll remove them now.

chewnoill (approved):
🚀 Nice work @tmsavage!

app/javascript/components/deployedVersionCompare.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+      const newValue = data.deploymentSha;
+      if (lastValue == null) {
+        lastValue = newValue;
+      } else if (newValue !== lastValue) {
+        lastValue = newValue;
+        window.location.reload();
+      }
+
+      return newValue;
```
cbarber, line 28: This will render some content into the DOM. I don't think we need to render anything at all. In that case, could we skip using React entirely by calling the apollo client API directly with a subscription?

app/javascript/components/app.jsx
```diff
@@ -14,6 +14,7 @@ import { getMainDefinition } from 'apollo-utilities';
 import ActionCableLink from 'graphql-ruby-client/subscriptions/ActionCableLink';
 import { persistCache } from 'apollo-cache-persist';
 import WidgetController from '@components/widget-controller';
+import VersionCompare from '@components/deployedVersionCompare';
```
tucker-weed, line 17: Nit: this file name might be clearer if it follows the '-' pattern as opposed to camelCase

bencates (approved):
looks good! :ship:

PR #256, opened 2021-06-18
feat:Sp/vary cloud speed with wind intensity

sandipa12:
The cloud speed varies with the wind intensity. The wind speed is  queried from gql and passed it to the duration key inside  cloud which controls the speed of moving cloud.

### DIFF
--- a/app/javascript/components/widgets/weather/weather-effect.jsx
+++ b/app/javascript/components/widgets/weather/weather-effect.jsx
@@ -354,102 +354,102 @@ const DrizzleModerate = () => (
 const DrizzleHeavy = () => <Drizzle duration={1.2} variance={0.4} count={60} />;
 
 const weatherIdMap = {
-  200: (
+  200: () => (
     <>
       <RainModerate /> <Thunder />
     </>
   ), // Light rain
-  201: (
+  201: () => (
     <>
       <RainHeavy /> <Thunder />
     </>
   ), // Moderate rain
-  202: (
+  202: () => (
     <>
       <RainIntense /> <Thunder />
     </>
   ), // Heavy rain
-  210: <Thunder />, // Light thunderstorm
-  211: <Thunder />, // Moderate thunderstorm
-  212: <Thunder />, // Heavy thunderstorm
-  221: <Thunder />, // Ragged thunderstorm
-  230: (
+  210: () => <Thunder />, // Light thunderstorm
+  211: () => <Thunder />, // Moderate thunderstorm
+  212: () => <Thunder />, // Heavy thunderstorm
+  221: () => <Thunder />, // Ragged thunderstorm
+  230: () => (
     <>
       <DrizzleLight /> <Thunder />
     </>
   ), // Light drizzle
-  231: (
+  231: () => (
     <>
       <DrizzleModerate /> <Thunder />
     </>
   ), // Moderate drizzle
-  232: (
+  232: () => (
     <>
       <DrizzleHeavy /> <Thunder />
     </>
   ), // Heavy drizzle
-  300: <DrizzleLight />, // Light
-  301: <DrizzleModerate />, // Moderate
-  302: <DrizzleHeavy />, // Heavy
-  310: (
+  300: () => <DrizzleLight />, // Light
+  301: () => <DrizzleModerate />, // Moderate
+  302: () => <DrizzleHeavy />, // Heavy
+  310: () => (
     <>
       <DrizzleLight /> <RainLight />
     </>
   ), // Light rain
-  311: (
+  311: () => (
     <>
       <DrizzleLight /> <RainModerate />
     </>
   ), // Moderate rain
-  312: (
+  312: () => (
     <>
       <DrizzleModerate /> <RainModerate />
     </>
   ), // Heavy rain
-  313: (
+  313: () => (
     <>
       <DrizzleLight /> <RainModerate />
     </>
   ), // Shower rain
-  314: (
+  314: () => (
     <>
       <DrizzleModerate /> <RainModerate />
     </>
   ), // Heavy shower rain
-  321: <DrizzleModerate />, // Shower
-  500: <RainLight />, // Light
-  501: <RainModerate />, // Moderate
-  502: <RainHeavy />, // Heavy
-  503: <RainIntense />, // Very heavy
-  504: <RainExtreme />, // Extreme
-  511: <RainModerate />, // Freezing
-  520: <RainLight />, // Light shower
-  521: <RainModerate />, // Moderate shower
-  522: <RainHeavy />, // Heavy shower
-  531: <RainModerate />, // Ragged shower
-  600: <Snow duration={6} variance={2} count={20} />, // Light
-  601: <Snow duration={4} variance={2} count={50} />, // Normal
-  602: <Snow duration={2} variance={2} count={70} />, // Heavy
-  611: <HeavySnow duration={3} variance={2} count={20} />, // Sleet
-  612: <HeavySnow duration={2} variance={1} count={35} />, // Sleet
-  613: <HeavySnow duration={1.5} variance={1} count={45} />, // Sleet
-  615: (
+  321: () => <DrizzleModerate />, // Shower
+  500: () => <RainLight />, // Light
+  501: () => <RainModerate />, // Moderate
+  502: () => <RainHeavy />, // Heavy
+  503: () => <RainIntense />, // Very heavy
+  504: () => <RainExtreme />, // Extreme
+  511: () => <RainModerate />, // Freezing
+  520: () => <RainLight />, // Light shower
+  521: () => <RainModerate />, // Moderate shower
+  522: () => <RainHeavy />, // Heavy shower
+  531: () => <RainModerate />, // Ragged shower
+  600: () => <Snow duration={6} variance={2} count={20} />, // Light
+  601: () => <Snow duration={4} variance={2} count={50} />, // Normal
+  602: () => <Snow duration={2} variance={2} count={70} />, // Heavy
+  611: () => <HeavySnow duration={3} variance={2} count={20} />, // Sleet
+  612: () => <HeavySnow duration={2} variance={1} count={35} />, // Sleet
+  613: () => <HeavySnow duration={1.5} variance={1} count={45} />, // Sleet
+  615: () => (
     <>
       <Snow duration={4} variance={2} count={20} /> <RainLight />
     </>
   ), // Light rain and snow
-  616: (
+  616: () => (
     <>
       <Snow duration={2} variance={2} count={30} /> <RainModerate />
     </>
   ), // Rain and snow
-  620: <Snow duration={6} variance={1} count={20} />, // Light
-  621: <Snow duration={4} variance={1} count={50} />, // Normal
-  622: <Snow duration={2} variance={1} count={70} />, // Heavy
-  801: <Cloud duration={16} variance={1} count={2} />, // few cloud
-  802: <Cloud duration={16} variance={5} count={4} />, // scattered cloud
-  803: <Cloud duration={16} variance={10} count={8} />, // broken cloud
-  804: <Cloud duration={16} variance={15} count={16} />, // overcast cloud
+  620: () => <Snow duration={6} variance={1} count={20} />, // Light
+  621: () => <Snow duration={4} variance={1} count={50} />, // Normal
+  622: () => <Snow duration={2} variance={1} count={70} />, // Heavy
+  801: () => <Cloud duration={16} variance={1} count={2} />, // few cloud
+  802: () => <Cloud duration={16} variance={5} count={4} />, // scattered cloud
+  803: () => <Cloud duration={16} variance={10} count={8} />, // broken cloud
+  804: () => <Cloud duration={16} variance={15} count={16} />, // overcast cloud
 };
 /* eslint-enable prettier/prettier */
 
@@ -462,7 +462,11 @@ const WeatherEffect = ({ weather }) => {
 
   const Effect = weatherIdMap[id];
 
-  return Effect ? <EffectContainer>{Effect}</EffectContainer> : null;
+  return Effect ? (
+    <EffectContainer>
+      <Effect />
+    </EffectContainer>
+  ) : null;
 };
 
 WeatherEffect.propTypes = {
--- a/app/javascript/components/widgets/weather/weather-effect.jsx
+++ b/app/javascript/components/widgets/weather/weather-effect.jsx
@@ -10,6 +10,7 @@ const getWeatherEffect = gql`
       weather {
         id
       }
+      windSpeed
     }
   }
 `;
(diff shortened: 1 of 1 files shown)


app/javascript/components/widgets/weather/weather-effect.jsx
```diff
@@ -345,61 +346,62 @@ const DrizzleLight = () => <Drizzle duration={1.8} variance={0.5} count={30} />;
...
+    613: <HeavySnow duration={1.5} variance={1} count={45} />, // Sleet
+    615: <><Snow duration={4} variance={2} count={20} /> <RainLight /></>, // Light rain and snow
+    616: <><Snow duration={2} variance={2} count={30} /> <RainModerate /></>, // Rain and snow
+    620: <Snow duration={6} variance={1} count={20} />, // Light
+    621: <Snow duration={4} variance={1} count={50} />, // Normal
+    622: <Snow duration={2} variance={1} count={70} />, // Heavy
+    801: <Cloud duration={windSpeed * 100} variance={1} count={2} />, // few cloud
+    802: <Cloud duration={windSpeed * 100} variance={5} count={4} />, // scattered cloud
+    803: <Cloud duration={windSpeed * 100} variance={10} count={8} />, // broken cloud
+    804: <Cloud duration={windSpeed * 100} variance={15} count={16} />, // overcast cloud
```
tucker-weed, line 402: Accidentally removed my comment. Original post was: "How does duration work here? Would increasing the windSpeed make the animation take longer?"

sandipa12: Yes, greater the value inside duration, the longer it takes for animation and vice versa.

app/javascript/components/widgets/weather/weather-effect.jsx
```diff
@@ -345,61 +346,62 @@ const DrizzleLight = () => <Drizzle duration={1.8} variance={0.5} count={30} />;
...
+    602: <Snow duration={2} variance={2} count={70} />, // Heavy
+    611: <HeavySnow duration={3} variance={2} count={20} />, // Sleet
+    612: <HeavySnow duration={2} variance={1} count={35} />, // Sleet
+    613: <HeavySnow duration={1.5} variance={1} count={45} />, // Sleet
+    615: <><Snow duration={4} variance={2} count={20} /> <RainLight /></>, // Light rain and snow
+    616: <><Snow duration={2} variance={2} count={30} /> <RainModerate /></>, // Rain and snow
+    620: <Snow duration={6} variance={1} count={20} />, // Light
+    621: <Snow duration={4} variance={1} count={50} />, // Normal
+    622: <Snow duration={2} variance={1} count={70} />, // Heavy
+    801: <Cloud duration={windSpeed * 100} variance={1} count={2} />, // few cloud
```
cbarber, line 399: So it looks like you ran into a problem here trying to inject a property acquired from the graphql query into the constant component renders in `weatherIdMap`. This is probably the simplest solution by presents some problems:
* We've moved a lot of code that was not related to our feature
* `weatherIdMap` is rebuilt on every render removing the benefit of `weatherIdMap`
* The `WeatherEffect` render is very long now

Some other approaches:
* Leave `weatherIdMap` defined above and split out a condition for `801`, `802`, `803`, `804`. This is slightly simpler approach but breaks our original pattern. It is very likely we will want windSpeed (maybe other values) in our other weather effects so lets skip this.
* React provides an API `cloneElement` that can take an existing element and extend it with additional props: https://reactjs.org/docs/react-api.html#cloneelement This keeps our original pattern but likely does not do what we want for the weather effects in fragments: <><RainModerate /> <Thunder /></>
* Rethink our original pattern to suit our new desired functionality. We could switch all existing entries (refactor commit) to a new pattern of
```
{
  200: [{ type: RainModerate: props: {}}, { type: Thunder, props: {}}],
  ...
  600: [{ type: Snow, props: { duration: 6, variance: 2, count: 20 }} ],
}
```
and later our in `WeatherEffect`:
```
const Effects = weatherIdMap[id];
return Effects ? <EffectContainer>{Effects.map(({ type, props }) => <type {...props} />)}</EffectContainer> : null;
```

After the refactor, you're now free to inject `windSpeed` (and any other piece of weather data) into the rendered effects.

app/javascript/components/widgets/weather/weather-effect.jsx
```diff
@@ -345,61 +346,62 @@ const DrizzleLight = () => <Drizzle duration={1.8} variance={0.5} count={30} />;
...
   } = weather;
 
+  const weatherIdMap = {
+    200: <><RainModerate /> <Thunder /></>, // Light rain
+    201: <><RainHeavy /> <Thunder /></>, // Moderate rain
+    202: <><RainIntense /> <Thunder /></>, // Heavy rain
+    210: <Thunder />, // Light thunderstorm
+    211: <Thunder />, // Moderate thunderstorm
+    212: <Thunder />, // Heavy thunderstorm
+    221: <Thunder />, // Ragged thunderstorm
```
tmsavage, line 365: What do these numbers mean, such as `200:`, and how are they used?

sandipa12: These are the id's for each type of effect and it comes from open weather map API.
https://openweathermap.org/weather-conditions#Weather-Condition-Codes-

app/javascript/components/widgets/weather/weather-effect.jsx
```diff
@@ -345,61 +346,62 @@ const DrizzleLight = () => <Drizzle duration={1.8} variance={0.5} count={30} />;
...
-  615: <><Snow duration={4} variance={2} count={20} /> <RainLight /></>, // Light rain and snow
-  616: <><Snow duration={2} variance={2} count={30} /> <RainModerate /></>, // Rain and snow
-  620: <Snow duration={6} variance={1} count={20} />, // Light
-  621: <Snow duration={4} variance={1} count={50} />, // Normal
-  622: <Snow duration={2} variance={1} count={70} />, // Heavy
-  801: <Cloud duration={16} variance={1} count={2} />, // few cloud
-  802: <Cloud duration={16} variance={5} count={4} />, // scattered cloud
-  803: <Cloud duration={16} variance={10} count={8} />, // broken cloud
-  804: <Cloud duration={16} variance={15} count={16} />, // overcast cloud
-}
```
chewnoill, line 393: I'm not a fan of putting this weather map inside another component, because then it needs to be recreated on every render. 

I think a better patter for injecting props at runtime would be something like:

```jsx
const weatherIdMap = {
  ...
  804: ({windSpeed}) => <Cloud duration={16} variance={15} count={16} />, 
}

...
const WeatherEffect = ...
   const Component = weatherIdMap[id];
   return <Component windSpeed={windSpeed} />

File: app/javascript/components/widgets/weather/weather-effect.jsx (hunk 1 above)
Comment (chewnoill) [reply app/javascript/components/widgets/weather/weather-effect.jsx:402]: This is confusing. I expect increasing `windSpeed` to make the animation play faster, and increasing the duration to slow down the animation. This usually means that something is misnamed or we're doing it wrong.

File: app/javascript/components/widgets/weather/weather-effect.jsx (hunk 3 above)
Comment (tucker-weed) [reply app/javascript/components/widgets/weather/weather-effect.jsx:365]: Can any type of effect from the open weather API can be mapped to a component of the developer's choosing using this interface?

File: app/javascript/components/widgets/weather/weather-effect.jsx [hunk 5]
```diff
@@ -345,61 +346,62 @@ const DrizzleLight = () => <Drizzle duration={1.8} variance={0.5} count={30} />;
...
 
 const WeatherEffect = ({ weather }) => {
   const {
     current: {
       weather: { id },
+      windSpeed
     },
   } = weather;
 
+  const weatherIdMap = {
```
fallynlogan, line 358: How are these components being mapped in this function?

app/javascript/components/widgets/weather/weather-effect.jsx
sandipa12: It seems that the duration is controlling the windspeed as of now.  Increasing the duration slows down the animation and vice versa. How about writing a condition that would increase the duration if the windspeed is low and decrease it when the windspeed is high?

app/javascript/components/widgets/weather/weather-effect.jsx
sandipa12: They have their own id's that come from the Open Weather map API.

app/javascript/components/widgets/weather/weather-effect.jsx
```diff
@@ -345,64 +346,88 @@ const DrizzleLight = () => <Drizzle duration={1.8} variance={0.5} count={30} />;
...
   const Effect = weatherIdMap[id];
 
-  return Effect ? <EffectContainer>{Effect}</EffectContainer> : null;
+  return Effect ? (
+    <EffectContainer>
+      {Effect.map(({ Type, props, idx }) => {
+        if (weather.current.windSpeed <= 3.0) {
+          // eslint-disable-next-line no-param-reassign
+          props.duration += 15;
+          return <Type key={idx} {...props} />;
```
tucker-weed, line 414: Perhaps disabling lint rules can be avoided by passing duration like this: `<Type key={idx} {...props} duration={props.duration + 15} />`

I believe that passing the `duration` prop after the `{...props}` spread assignment will make it such that the child component has all the old props except for `duration` which was overwritten by `duration={props.duration + 15}`

app/javascript/components/widgets/weather/weather-effect.jsx
```diff
@@ -345,64 +346,88 @@ const DrizzleLight = () => <Drizzle duration={1.8} variance={0.5} count={30} />;
...
+          weather.current.windSpeed <= 4.5
+        ) {
+          return <Type key={idx} {...props} />;
+        }
+        if (weather.current.windSpeed > 4.5) {
+          // eslint-disable-next-line no-param-reassign
+          props.duration -= 5;
+          return <Type key={idx} {...props} />;
+        }
+        return null;
```
tucker-weed, line 427: I think that this last conditional can be dropped because the first two conditionals make it such that this statement: `if (weather.current.windSpeed > 4.5)` will always evaluate to true

sandipa12: Yeah. I could remove it.

app/javascript/components/widgets/weather/weather-effect.jsx
```diff
@@ -460,9 +474,28 @@ const WeatherEffect = ({ weather }) => {
...
+  const calcDuration = item => {
+    // eslint-disable-next-line no-shadow
+    const { Type, props, idx } = item;
+    if (idx > 53) {
+      if (weather.current.windSpeed <= 3.0) {
+        props.duration += 10;
+        return <Type key={idx} {...props} />;
+      }
+      if (weather.current.windSpeed > 4.5) {
+        props.duration -= 5;
```
cbarber, line 489: A general rule of props is you should not mutate them.

app/javascript/components/widgets/weather/weather-effect.jsx
```diff
@@ -460,9 +474,28 @@ const WeatherEffect = ({ weather }) => {
...
   } = weather;
 
+  /* eslint-enable prettier/prettier */
   const Effect = weatherIdMap[id];
 
-  return Effect ? <EffectContainer>{Effect}</EffectContainer> : null;
+  const calcDuration = item => {
+    // eslint-disable-next-line no-shadow
+    const { Type, props, idx } = item;
+    if (idx > 53) {
```
cbarber, line 483: Is this to only apply the logic to Clouds? If that is the case it would make more sense to move the logic into `Cloud`.

app/javascript/components/widgets/weather/weather-effect.jsx
```diff
@@ -354,104 +356,116 @@ const DrizzleModerate = () => (
...
-  620: <Snow duration={6} variance={1} count={20} />, // Light
-  621: <Snow duration={4} variance={1} count={50} />, // Normal
-  622: <Snow duration={2} variance={1} count={70} />, // Heavy
-  801: <Cloud duration={16} variance={1} count={2} />, // few cloud
-  802: <Cloud duration={16} variance={5} count={4} />, // scattered cloud
-  803: <Cloud duration={16} variance={10} count={8} />, // broken cloud
-  804: <Cloud duration={16} variance={15} count={16} />, // overcast cloud
+  200: [
+    { Type: RainModerate, props: {}, idx: 1 },
+    { Type: Thunder, props: {}, idx: 2 },
```
cbarber, line 361: There isn't much value to actually making unique idx here. Now that I see it I think Will's suggestion to make these nested components was better. 

I think we also have two commits here. The first being a refactor that adjusts `weatherIdMap` and `WeatherEffect` to support mapping and passing props. The second adding the new windSpeed functionality.

cbarber (changes requested):

app/javascript/components/widgets/weather/weather-effect.jsx
```diff
@@ -354,104 +356,116 @@ const DrizzleModerate = () => (
...
-    </>
-  ), // Rain and snow
-  620: <Snow duration={6} variance={1} count={20} />, // Light
-  621: <Snow duration={4} variance={1} count={50} />, // Normal
-  622: <Snow duration={2} variance={1} count={70} />, // Heavy
-  801: <Cloud duration={16} variance={1} count={2} />, // few cloud
-  802: <Cloud duration={16} variance={5} count={4} />, // scattered cloud
-  803: <Cloud duration={16} variance={10} count={8} />, // broken cloud
-  804: <Cloud duration={16} variance={15} count={16} />, // overcast cloud
+  200: [
```
fallynlogan, line 359: Why do we have these numbered ID's and is there any relation between the number and the weather types described in them?

app/javascript/components/widgets/weather/weather-effect.jsx
```diff
@@ -319,17 +320,36 @@ const MovingCloud = styled.div`
   ${CloudImage};
 `;
 
-const Cloud = ({ duration, variance, count }) => (
-  <HorizontalWeatherWrapper
-    {...{ Type: MovingCloud, duration, variance, count }}
-  />
-);
+const Cloud = ({ duration, variance, count, windSpeed }) => {
+  // eslint-disable-next-line no-shadow
```
cbarber, line 324: Why is there a no-shadow disable line here?

app/javascript/components/widgets/weather/weather-effect.jsx
```diff
@@ -446,13 +483,13 @@ const weatherIdMap = {
...
   621: <Snow duration={4} variance={1} count={50} />, // Normal
   622: <Snow duration={2} variance={1} count={70} />, // Heavy
-  801: <Cloud duration={16} variance={1} count={2} />, // few cloud
-  802: <Cloud duration={16} variance={5} count={4} />, // scattered cloud
-  803: <Cloud duration={16} variance={10} count={8} />, // broken cloud
-  804: <Cloud duration={16} variance={15} count={16} />, // overcast cloud
+  801: () => <CloudWrapper variance={1} count={2} />, // few cloud
+  802: () => <CloudWrapper variance={5} count={4} />, // scattered cloud
+  803: () => <CloudWrapper variance={10} count={8} />, // broken cloud
+  804: () => <CloudWrapper variance={15} count={16} />, // overcast cloud
```
cbarber, line 489: This doesn't look like it passes the `windSpeed` property to the `CloudWrapper`.

app/javascript/components/widgets/weather/weather-effect.jsx
```diff
@@ -319,17 +320,36 @@ const MovingCloud = styled.div`
...
+      <HorizontalWeatherWrapper
+        {...{ Type: MovingCloud, duration: duration - 5, variance, count }}
+      />
+    );
+  }
+  return (
+    <HorizontalWeatherWrapper
+      {...{ Type: MovingCloud, duration, variance, count }}
+    />
+  );
```
cbarber, line 343: Nit: this is a little odd that we have a default duration that is for winds > 10 and < 15. Defaulting to the slowest wind duration and decreasing the duration for faster wind speeds may be a little more intuitive.

cbarber (approved):
LGTM :+1:

PR #257, opened 2021-06-21
chore: fully synchronize components with url location

tucker-weed:
**Description**

1. Does away with primaryLocation references on the client side entirely.

### DIFF
--- a/app/javascript/components/city-name.jsx
+++ b/app/javascript/components/city-name.jsx
@@ -1,32 +1,13 @@
 import React from 'react';
-import { Query } from 'react-apollo';
-import gql from 'graphql-tag';
+import PropTypes from 'prop-types';
 import { WhiteTitleLarge } from '@components/typography';
-import { LoadingMessage, ErrorMessage } from '@messages/default-messages';
 
-const getCityName = gql`
-  {
-    primaryLocation {
-      cityName
-    }
-  }
-`;
-
-export const City = () => (
-  <Query query={getCityName}>
-    {({ loading, error, data }) => {
-      if (loading) {
-        return <LoadingMessage />;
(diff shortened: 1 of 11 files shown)


tucker-weed:
Only Providence has a bathroom code it seems - should the other locations display something like Bathroom Code: unavailable? Currently the bathroom code component is invisible when the query returns a null response.

app/javascript/components/city-name.jsx
```diff
@@ -1,32 +1,13 @@
...
-      if (error) {
-        return <ErrorMessage />;
-      }
-
-      const { cityName } = data.primaryLocation;
-      return cityName && <WhiteTitleLarge>{cityName}</WhiteTitleLarge>;
-    }}
-  </Query>
+export const City = ({ cityName }) => (
+  <WhiteTitleLarge>{cityName}</WhiteTitleLarge>
```
chewnoill, line 6: Be careful here. Just because cityName is a string does not mean its display ready. It's probably fine here though. Do we need to care about rendering every string that come in the URL?

Also, this will render right away, instead of having to manage a loading state, which is a better experience for everyone involved. 👍

chewnoill (approved):
This is a good example of a refactor that touches a lot of things, but is still easy to follow. Nice work @tucker-weed!!!!

Is the next step removing the primary location query from the gql schema?

cbarber:
> Currently the bathroom code component is invisible when the query returns a null response.

That sounds fine to me. No reason to display if it not relevant.

> Is the next step removing the primary location query from the gql schema?

* Remove primaryLocation
* Move sidebar queries to `withFragment` so we have a single graphql query to load our entire page.

PR #258, opened 2021-06-21
chore: remove unnecessary remounts of WidgetController on render

tucker-weed:
**Description**

Passing WidgetController to the `component` prop of Route via an inline function causes unmounting and remounting of the component rather than being updated. Use of the `render` property fixes this issue.

Source: https://reactrouter.com/web/api/Route/component


cbarber (approved):
Nice catch. This is a handy to have in the arsenal to catch unnecessary renders:
https://www.npmjs.com/package/@welldone-software/why-did-you-render

PR #259, opened 2021-06-21
chore: deprecated primaryLocation and unused Locations[] fields

tmsavage:
The query fields primaryLocation and Locations[] have been deprecated, meaning, removed from the GraphQL database. Locations[] was not being used, while primaryLocation was only used under the spec file `weather_spec.rb`. This file's test query was updated. All queries for a location given a name have been using the field location(cityName). The purpose of this PR is to eliminate redundancy on similar ways to query a location.


cbarber (commented):
Nit: technically not a bug. Can we split this into two commits, one for `locations` and one for `primaryLocation`?

tmsavage:
> Nit: technically not a bug. Can we split this into two commits, one for `locations` and one for `primaryLocation`?

@cbarber Would you call this a feature?

cbarber:
Probably a chore or feat.

bencates:
yup, `chore`

bencates (approved):
image
:heart: Always a nice thing to see on a PR.

chewnoill (approved):
🚀

cbarber:
You renamed the PR not the commit :(

PR #260, opened 2021-06-21
chore: remove redundant Switch component from routing logic

tucker-weed:


app/javascript/components/app.jsx
```diff
@@ -80,17 +80,15 @@ const App = () => (
   <ApolloProvider client={client}>
     <GlobalStyle />
     <Router>
-      <Switch>
-        <Route
-          path="/:city_name"
```
tmsavage, line 85: What is your reasoning for removing `Switch` instead of `Route`?

cbarber (approved):
lgtm :+1:

app/javascript/components/app.jsx
chewnoill: If there is only one `Route`, a `Switch` doesn't really do anything. Kind of like having a [`switch`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) with only one `case`

chewnoill (approved):
🚀

PR #263, opened 2021-06-23
feat: implemented graphql fragments to reduce side panel querying

tmsavage:
Fragments have been implemented for the side panel components in order to reduce the amount of querying from the GraphQL database using the `withFragment` HOC.


app/javascript/components/time.jsx
```diff
@@ -32,7 +32,7 @@ const TimeValue = styled(Row)`
 
 export class Time extends React.Component {
   static propTypes = {
-    timezone: PropTypes.string.isRequired,
+    timezone: PropTypes.string,
```
chewnoill, line 35: why is this component changing?

app/javascript/components/bathroom.jsx
```diff
@@ -20,34 +18,36 @@ const Column = styled.div`
...
+    bathroomCode: PropTypes.string,
+  }),
+  loading: PropTypes.bool,
+  error: PropTypes.bool,
 };
 
-export default Bathroom;
+Bathroom.fragments = {
+  location: getBathroomCode
+}
```
chewnoill, line 51: Might be worth while to investigate using the `relay` client here instead, since it seems like we're emulating that pattern  here. It might provide some nice features we can use, and help organize these fragment dependencies better. 

https://relay.dev/docs/guided-tour/rendering/fragments/

cc @cbarber

cbarber: @chewnoill 
* What are the other nice features it offers over apollo?
* How does it organize the fragment dependencies better?
* Does it support subscriptions in addition to queries?
* What's the difficulty of switching out apollo for relay in our entire frontend?

chewnoill: * [loading states with suspense](https://relay.dev/docs/guided-tour/rendering/loading-states/) seems pretty cool
* A main query component introspects named fragments, from its child component tree. Pretty interesting actually, basically the same concept as withFragments I think, but with extra hooks into the VDOM.
* Yes
* This is what I want investigated.

cbarber: @tmsavage don't worry about this discussion. Its out of scope of your work.

app/javascript/components/time.jsx
tucker-weed: I believe `.isRequired` can remain on the `timezone` property because the TimeHero component will return the LoadingMessage or ErrorMessage components if the location property doesn't contain data (so the Time component will never be rendered if location doesn't have the required data).

tmsavage: I removed `.isRequired` to avoid the error message: "The prop `timezone` is marked as required in `Time`, but its value is `undefined`" for all locations. Am I not declaring `timezone` properly?

app/javascript/components/side-panel.jsx
```diff
@@ -26,20 +28,33 @@ const FixedContent = styled.div`
...
 `;
 
+const getSidePanel = gql`
+  fragment SidePanel on Location {
+    ...Wifi
+    ...Bathroom
+  }
+
+  ${Wifi.fragments.location}
+  ${Bathroom.fragments.location}
```
bencates, line 38: I'm not a huge fan of how we're pulling the fragments in here. The `.fragments.location` feels like it's an implementation detail of the `withFragments` HOC.

Can we explicitly export these from their respective files and pull them in here as named imports? That way it makes it explicit to the reader that these are being referenced somewhere other than the file where they're defined.

app/javascript/components/top-corner-controller.jsx
```diff
@@ -26,7 +28,15 @@ const Row = styled.div`
   margin-right: 100px;
 `;
 
-const TopCorner = ({ showWeather, cityName }) => {
+const getTopCorner = gql`
+  fragment TopCorner on Location {
+    ...TimeHero
+  }
+
+  ${TimeHero.fragments.location}
```
bencates, line 36: Same comment as above

app/javascript/components/widget-controller.jsx
```diff
@@ -31,6 +32,9 @@ const getWidgets = gql`
       }
     }
   }
+
+
+  ${SidePanel.fragments.location}
```
bencates, line 37: Same comment as above

bencates (commented):
Looks pretty good!

There's a little bit of cleanup I'd like to see in how we're referencing the fragments. It's a small change, but it'll make it a bit more clear to someone reading the code for the first time what they can and can't safely change.

@cbarber and I had a chat this morning on Slack. There's some follow-on work we'd like you to do to build on this refactor once this ticket is merged. Please ping one of us before you pick up your next ticket.

app/javascript/components/time.jsx
tucker-weed: Is it possible that `loading` and `error` are both false at some point?

app/javascript/components/bathroom.jsx
```diff
@@ -20,34 +18,36 @@ const Column = styled.div`
...
-
-      if (error) {
-        return <ErrorMessage />;
-      }
-
-      const { bathroomCode } = data.location;
-      if (!bathroomCode) {
-        return null;
-      }
+export const Bathroom = ({ loading, error, location }) => {
```
fallynlogan, line 21: How does passing in these props instead of the previous one (cityName) reduce side panel queries?

app/javascript/components/top-corner-controller.jsx
```diff
@@ -47,4 +57,10 @@ TopCorner.propTypes = {
   showWeather: PropTypes.bool.isRequired,
   cityName: PropTypes.string.isRequired,
 };
-export default TopCorner;
+
+TopCorner.fragments = {
```
fallynlogan, line 61: How do fragments reduce the amount of times we are querying components in the side panel?

fallynlogan (commented):
Just had some questions to help with my understanding!

app/javascript/components/bathroom.jsx
```diff
@@ -1,16 +1,14 @@
...
+import withFragment from './hocs/with-fragment';
 
-const getBathroomCode = gql`
-  query getBathroomCode($cityName: String!) {
-    location(cityName: $cityName) {
-      bathroomCode
-    }
+export const getBathroomCode = gql`
+  fragment Bathroom on Location {
+    bathroomCode
```
tucker-weed, line 11: Why doesn't the fragment need parameters (i.e. $cityName: String!)?

cbarber: The only requirement for Bathroom (and the other components) is that this data come from a `Location` type. That could be composed in any Location type result but in our case is composed in the `getWidgets` query that specifies the city_name for a single location.

app/javascript/components/widget-controller.jsx
```diff
@@ -269,4 +276,9 @@ WidgetController.propTypes = {
   cityName: PropTypes.string.isRequired,
 };
 
+// WidgetController.fragments = {
+//   widgets: getWidgets,
+// };
+
 export default WidgetController;
+// export default withFragment(WidgetController);
```
bencates, line 284: Looks like some leftover comments snuck into the PR. Let's remove them.

bencates (approved):
This looks great! Just a couple of nitpicks then :ship:!

I do see a couple of follow-on tasks once this is merged in case someone wants to pick them up:

1. `withFragment()` currently expects the component it's passed to have a `.fragments` property. That's a bit more opaque than it needs to be and requires us to use the `hoist-non-react-statics` module inside the HOC. It would be cleaner to pass the fragments in as a second argument, so e.g.

   ```js
   Bathroom.fragments = {
     location: getBathroomCode,
   };

   export default withFragment(Bathroom);
   ```

   would become

   ```js
   export default withFragment(Bathroom, { location: getBathroomCode });
   ```

2. We're only making a single query in `widget-controller.jsx`, but we're passing the `loading` and `error` states down to the leaf components. That's leading to a lot of unnecessary duplication.
   I'd like to see how far we can move `loading` and especially `error` handling up the component heirarchy

app/javascript/components/bathroom.jsx
cbarber: The components originally had `<Query>` components, each running a separate query. This switches to fragments that are exported and composed within parent queries, eventually reaching the parent location query in widget-controller. These props are passed down from widget-controller through the children components to as a result of the fragment resolving in the main query.

app/javascript/components/widget-display.jsx
```diff
@@ -56,19 +61,28 @@ WidgetDisplay.propTypes = {
...
+      }).isRequired,
     }).isRequired,
   }).isRequired,
+  loading: PropTypes.bool,
+  error: PropTypes.bool,
 };
 
-export const LoadingDisplay = ({ cityName }) => (
+WidgetDisplay.defaultProps = {
+  loading: true,
```
tucker-weed, line 81: Why is the default value for loading true?

cbarber (approved):
Nice work :shipit: Next up, rewriting `withFragment` to remove the statics and adding in handy `renderWhileLoading` and `renderWhileError` :grinning:

PR #265, opened 2021-06-24
prettier in ci

cbarber:


package.json
```diff
@@ -66,7 +66,7 @@
     "lint": "./node_modules/.bin/eslint --ext js --ext jsx app/javascript/",
     "test": "jest",
     "relative-to-alias": "jscodeshift --extensions jsx -t ./config/transforms/relative_to_alias_path.js ./ ",
-    "format": "prettier --write '{app/javascript/,__mocks__}/**/*.{js,jsx}' package.json",
+    "prettier:format": "prettier --write '{app/javascript/,__mocks__}/**/*.{js,jsx}' package.json",
```
chewnoill, line 69: nit: no reason to expose prettier in this script, `yarn format` is simpler

chewnoill (approved):

package.json
cbarber: Convention. `yarn check` already exists so `prettier:format` looks similar to `prettier:check`

aaddaamm (approved):

PR #267, opened 2021-06-24
feat:add description of wind gusts

sandipa12:
Displays the description of wind gust if the wind speed is greater than 15 mph.
Demo of how it displays the wind gust information:
Screen Shot 2021-06-28 at 11 17 45 AM

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,3 +21,5 @@ services:
       - REDIS_URL=redis://redis:6379/1
   redis:
     image: redis:4.0
+    ports: 
+      - "6379:6379"


bbazyk (commented):
Are there plans for more commits where this PR will display gusts to the widget? Or is this ticket specifically for retrieval from openWeather ?
P.S. I'm excited for all the changes we've made to the weather widget so far - I think the visuals and information organization has been appreciated by anyone passing the helios projector screen in the pvd office.

app/javascript/components/widgets/weather/current-weather.jsx
```diff
@@ -32,6 +34,7 @@ const CurrentWeather = ({ weather, useLargeIcon }) => {
         <SmallSkyIcon icon={icon} />
       )}
       {description}
+      {Gust}
```
chewnoill, line 37: Better to stick to regular JSX syntax here I think:

```jsx
Gust {weather.current.windGusts}
```

app/javascript/components/widgets/weather/current-weather.jsx
```diff
@@ -22,8 +23,16 @@ const IconWrapper = styled.div`
   justify-content: center;
 `;
 
+function isGusty(weather) {
```
chewnoill, line 26: convention is that a function that starts with `is` returns a boolean.

app/javascript/components/widgets/weather/current-weather.jsx
chewnoill: Maybe we can inline this?
```jsx
{isGusty(weather)}
```

chewnoill: or even better 
```jsx
{weather.current.windGust > 15 && ` , gusty wind at ${weather.current.windGust} mph`}
```

app/lib/clients/weather_client.rb
```diff
@@ -1,5 +1,5 @@
 class Clients::WeatherClient
-  WEATHER_API_URL = 'https://api.openweathermap.org/data/2.5/onecall?lat=%{lat}&lon=%{lon}&appid=%{key}'.freeze
+  WEATHER_API_URL = 'https://api.openweathermap.org/data/2.5/onecall?lat=%{lat}&lon=%{lon}&appid=%{key}&units=imperial'.freeze
```
chewnoill, line 2: unclear why we're touching this. Maybe put it in its own commit? I think this is a separate idea.

sandipa12: The original WEATHER_API_URL has standard unit which gives the data in m/s. Adding units= imperial gives the data in mph.

app/javascript/components/widgets/weather/current-weather.jsx
cbarber: Nit: prefer extracting/naming components over inlining code in JSX.

app/lib/clients/weather_client.rb
cbarber: This should be a separate fix commit.

PR #270, opened 2021-06-28
feat:display information for gusty wind

sandipa12:
Display the information for gusty weather if the wind speed > 15 mph.
Demo of how the gusty wind is displayed:
Screen Shot 2021-06-29 at 11 18 39 AM

### DIFF
--- a/app/javascript/components/widgets/weather/current-weather.jsx
+++ b/app/javascript/components/widgets/weather/current-weather.jsx
@@ -12,6 +12,7 @@ const getCurrentWeather = gql`
         description
         icon
       }
+      windGust
     }
   }
 `;
@@ -22,6 +23,13 @@ const IconWrapper = styled.div`
   justify-content: center;
 `;
 
+const WeatherDescription = ({ baseDescription, weather }) => {
+  if (weather.current.windGust > 15) {
+    return `${baseDescription}, gusty wind at ${weather.current.windGust} mph`;
+  }
+  return baseDescription;
+};
+
 const CurrentWeather = ({ weather, useLargeIcon }) => {
   const { description, icon } = weather.current.weather;
   return (
@@ -31,7 +39,7 @@ const CurrentWeather = ({ weather, useLargeIcon }) => {
       ) : (
         <SmallSkyIcon icon={icon} />
       )}
-      {description}
+      <WeatherDescription baseDescription={description} weather={weather} />
     </IconWrapper>
   );
 };
--- a/app/lib/clients/weather_client.rb
+++ b/app/lib/clients/weather_client.rb
@@ -1,5 +1,6 @@
 class Clients::WeatherClient
-  WEATHER_API_URL = 'https://api.openweathermap.org/data/2.5/onecall?lat=%{lat}&lon=%{lon}&appid=%{key}'.freeze
+  WEATHER_API_URL = 'https://api.openweathermap.org/data/2.5/onecall?lat=%{lat}&lon=%{lon}&appid=%{key}&units=imperial'
+                    .freeze
 
   def self.forecast(location)
     key = "openweather-response-#{location.latitude}-#{location.longitude}"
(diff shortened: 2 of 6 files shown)


tmsavage:
Cool feature. In your demo, are you just trying to show what the display looks like for gusty wind even though 7 mph is less than 15 mph?

app/lib/clients/weather_client.rb
```diff
@@ -1,5 +1,5 @@
 class Clients::WeatherClient
-  WEATHER_API_URL = 'https://api.openweathermap.org/data/2.5/onecall?lat=%{lat}&lon=%{lon}&appid=%{key}'.freeze
+  WEATHER_API_URL = 'https://api.openweathermap.org/data/2.5/onecall?lat=%{lat}&lon=%{lon}&appid=%{key}&units=imperial'.freeze
```
cbarber, line 2: Lets split this into a separate commit.

app/javascript/components/widgets/weather/current-weather.jsx
```diff
@@ -32,6 +33,8 @@ const CurrentWeather = ({ weather, useLargeIcon }) => {
         <SmallSkyIcon icon={icon} />
       )}
       {description}
+      {weather.current.windGust > 15 &&
+        ` , gusty wind at ${weather.current.windGust} mph`}
```
cbarber, line 37: Lets extract this into a component.

Nit: the extra space before the comma seems incorrect.

cbarber (changes requested):

app/javascript/components/widgets/weather/current-weather.jsx
```diff
@@ -32,6 +33,8 @@ const CurrentWeather = ({ weather, useLargeIcon }) => {
         <SmallSkyIcon icon={icon} />
       )}
       {description}
+      {weather.current.windGust > 15 &&
```
fallynlogan, line 36: What is this comparison doing? Will the wind gust information only be displayed if it is above 15?

sandipa12: yes, the information is displayed only for the speeed>15

sandipa12:
@tmsavage that's true.

app/javascript/components/widgets/weather/current-weather.jsx
```diff
@@ -32,6 +40,7 @@ const CurrentWeather = ({ weather, useLargeIcon }) => {
         <SmallSkyIcon icon={icon} />
       )}
       {description}
+      {Gust(weather)}
```
allisonlucas, line 43: Since Gust is a component, we should include it as such: `<Gust weather={weather} />`. `description` is not a React Component but a string variable, which is why it is included the way it is. It might be helpful to read through the React documentation's overview, which explains React Components and props: https://reactjs.org/tutorial/tutorial.html#overview.

chewnoill: could be inlined, easiest thing to do I think
```jsx
{ weather.current.windGust > 15 &&
 `, gusty wind at ${weather.current.windGust} mph` }
```

allisonlucas: Sandipa originally had it inline, but Craig suggested it be pulled out into a component: https://github.com/mojotech/helios2/pull/270#discussion_r660001746

chewnoill: I'm okay going back to what it was, creating a a whole component for a single span of un-styled text doesn't seem important.

Or if we just renamed the function so it didn't look like it was supposed to be a Component, that just never made it all the way. I imagine if it was named `getWindGustDescription(weather)` I wouldn't bat an eye.

chewnoill (approved):
🚀

cbarber (approved):
Nice job! :+1:

app/javascript/components/widgets/weather/current-weather.jsx
```diff
@@ -31,7 +39,7 @@ const CurrentWeather = ({ weather, useLargeIcon }) => {
       ) : (
         <SmallSkyIcon icon={icon} />
       )}
-      {description}
+      <WeatherDescription baseDescription={description} weather={weather} />
```
allisonlucas, line 42: This is great, nice job! 👍

PR #271, opened 2021-06-28
The side panel timer begins before the first widget is loaded

bbazyk:
ticket id #178304916
(This branch was created off of chore: WidgetController refactor #252)

### DIFF
--- a/app/javascript/components/get-weather-query.jsx
+++ b/app/javascript/components/get-weather-query.jsx
@@ -43,10 +43,13 @@ export const WeatherQuery = ({
   WeatherFrag,
   queryName,
   cityName,
+  startTimer,
 }) => (
   <Query
     query={getLocationWeather(WeatherFrag, queryName)}
     variables={{ cityName }}
+    onCompleted={startTimer}
+    onError={startTimer}
   >
     {({ loading, error, data, subscribeToMore }) => {
       if (loading) {
@@ -56,6 +59,7 @@ export const WeatherQuery = ({
         console.error(error); // eslint-disable-line
         return <DisconnectedMessage />;
       }
+
       return (
         <Subscription
           location={data.location}
@@ -96,6 +100,7 @@ WeatherQuery.propTypes = {
   }).isRequired,
   queryName: PropTypes.string.isRequired,
   cityName: PropTypes.string.isRequired,
+  startTimer: PropTypes.func.isRequired,
 };
 
 export default WeatherQuery;
--- a/app/javascript/components/widget-display.jsx
+++ b/app/javascript/components/widget-display.jsx
@@ -43,6 +43,7 @@ export const WidgetDisplay = ({
   location,
   loading,
   error,
+  startTimer,
 }) => {
   const { byIdOrFirst: current, enabled: enabledWidgets } = location.widgets;
   const WidgetElement = widgetElements[current.name].Panel;
@@ -54,7 +55,7 @@ export const WidgetDisplay = ({
       tabIndex="0"
     >
       <FullPanel>
-        <WidgetElement cityName={cityName} />
+        <WidgetElement startTimer={startTimer} cityName={cityName} />
       </FullPanel>
       <SidePanel
         widgets={enabledWidgets}
@@ -90,6 +91,7 @@ WidgetDisplay.propTypes = {
   }).isRequired,
   loading: PropTypes.bool,
   error: PropTypes.bool,
+  startTimer: PropTypes.func.isRequired,
 };
 
 WidgetDisplay.defaultProps = {
--- a/app/javascript/components/widget-transition-controls.jsx
+++ b/app/javascript/components/widget-transition-controls.jsx
@@ -83,13 +83,6 @@ const WidgetTransitionControls = ({
   });
 
   useEffect(() => {
-    dispatch({
-      type: 'startTimer',
-      msDurationRemaining: widgetTransitionTime,
-    });
-  }, [current.id]);
-
-  useEffect(() => {
     const timerId = setInterval(() => {
       if (!isStarted(state)) {
         return;
@@ -144,6 +137,10 @@ const WidgetTransitionControls = ({
     }
   };
 
+  const startTimer = () => {
+    dispatch({ type: 'startTimer', msDurationRemaining: widgetTransitionTime });
+  };
+
   return (
     <WrappedComponent
       keyToWidgetAction={keyToWidgetAction}
@@ -153,6 +150,7 @@ const WidgetTransitionControls = ({
       location={location}
       loading={loading}
       error={error}
+      startTimer={startTimer}
     />
   );
 };
--- a/app/javascript/components/widget-transition-controls.jsx
+++ b/app/javascript/components/widget-transition-controls.jsx
@@ -13,6 +13,8 @@ export const selectShouldRequestPrefetch = state =>
     prefetchDelay;
 
 export const selectIsPaused = state => state.startTimestamp === null;
+export const isStarted = state =>
+  state.startTimestamp !== null && state.msRemainingSinceLastStart !== null;
 
 const resolveWidgetId = (enabledWidgets, currentWidgetId, { type }) => {
   const index = enabledWidgets.findIndex(w => w.id === currentWidgetId);
--- a/app/javascript/components/widget-transition-controls.jsx
+++ b/app/javascript/components/widget-transition-controls.jsx
@@ -75,7 +75,7 @@ const WidgetTransitionControls = ({
   const widgetTransitionTime = current.durationSeconds * 1000;
 
   const [state, dispatch] = useReducer(reducer, {
-    startTimestamp: Date.now(),
+    startTimestamp: null,
     msRemainingSinceLastStart: null,
     prefetchRequestedYet: false,
   });
@@ -89,6 +89,9 @@ const WidgetTransitionControls = ({
 
   useEffect(() => {
     const timerId = setInterval(() => {
+      if (!isStarted(state)) {
+        return;
+      }
       if (selectIsPaused(state)) {
         return;
       }
--- a/app/javascript/components/widget-transition-controls.jsx
+++ b/app/javascript/components/widget-transition-controls.jsx
@@ -44,7 +44,7 @@ const reducer = (state, action) => {
     case 'startTimer':
       return {
         msRemainingSinceLastStart: action.msDurationRemaining,
-        startTimestamp: selectIsPaused(state) ? null : now,
+        startTimestamp: now,
         prefetchRequestedYet: false,
       };
     case 'tryPrefetch':
--- a/app/javascript/components/widget-transition-controls.jsx
+++ b/app/javascript/components/widget-transition-controls.jsx
@@ -9,6 +9,8 @@ export const selectShouldRequestWidget = state =>
 
 export const selectShouldRequestPrefetch = state =>
   !state.prefetchRequestedYet &&
+  !!state.msRemainingSinceLastStart &&
+  !!state.startTimestamp &&
   state.msRemainingSinceLastStart - (Date.now() - state.startTimestamp) <
     prefetchDelay;
 
--- a/app/javascript/components/widget-transition-controls.jsx
+++ b/app/javascript/components/widget-transition-controls.jsx
@@ -41,7 +41,7 @@ const reducer = (state, action) => {
       };
     case 'startTimer':
       return {
-        msRemainingSinceLastStart: action.msDurationSeconds,
+        msRemainingSinceLastStart: action.msDurationRemaining,
         startTimestamp: selectIsPaused(state) ? null : now,
         prefetchRequestedYet: false,
       };
@@ -81,7 +81,7 @@ const WidgetTransitionControls = ({
   useEffect(() => {
     dispatch({
       type: 'startTimer',
-      msDurationSeconds: widgetTransitionTime,
+      msDurationRemaining: widgetTransitionTime,
     });
   }, [current.id]);
 
--- a/app/javascript/components/widget-transition-controls.test.jsx
+++ b/app/javascript/components/widget-transition-controls.test.jsx
@@ -81,6 +81,24 @@ describe('shouldRequestPrefetch', () => {
       }),
     ).toEqual(true);
   });
+  it('returns false, msRemaining as null should not trigger request prefetch', () => {
+    expect(
+      selectShouldRequestPrefetch({
+        msRemainingSinceLastStart: null,
+        startTimestamp: Date.now() - 21000,
+        prefetchRequestedYet: false,
+      }),
+    ).toEqual(false);
+  });
+  it('returns false, startTimestamp as null should not trigger request prefetch', () => {
+    expect(
+      selectShouldRequestPrefetch({
+        msRemainingSinceLastStart: 20000,
+        startTimestamp: null,
+        prefetchRequestedYet: false,
+      }),
+    ).toEqual(false);
+  });
 });
 
 describe('shouldRequestWidget', () => {
--- a/app/javascript/components/widget-transition-controls.test.jsx
+++ b/app/javascript/components/widget-transition-controls.test.jsx
@@ -2,6 +2,7 @@ import {
   selectShouldRequestPrefetch,
   selectShouldRequestWidget,
   selectIsPaused,
+  isStarted,
 } from '@components/widget-transition-controls';
 
 describe('isPaused', () => {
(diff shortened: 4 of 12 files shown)


app/javascript/components/widget-controller.jsx
```diff
@@ -33,234 +29,58 @@ const getWidgets = gql`
...
-      );
-    }
-    if (keyCode === downKey) {
-      this.nextWidget();
-    }
-    if (keyCode === upKey) {
-      this.prevWidget();
-    }
-  };
+  const fetchPolicy = newWidgetId === 0 ? 'network-only' : 'cache-first';
```
sandipa12, line 35: what does this line do ?

bbazyk: I'm curious too - I believe this was a line added in tucker's branch? @tucker-weed

tucker-weed: The fetch policy is maintained from the previous version of WidgetController - my guess about what is going on here is that the first ever query run by the WidgetController (when `newWidgetId` is 0) should make a network request, whereas all subsequent queries should check if needed query information is in the cache before making a network request. Not 100% sure about this though so if anyone else knows what's really going on here, let us know.

cbarber: Filling in the missing piece. Its because we prefetch with a direct call to `client.query`. The flow:
* Page load / First request. `<Query`  uses network-only to get the latest data to make sure we're not stale. Updates cache for first widget.
* 1.5 seconds before transition, call `client.query`. This is always a network request. Updates cache for second widget.
* Transition time. Update state to next widget. `<Query` is not `cache-first`. `<Query` has a cache hit from the update from `client.query` 1.5 seconds ago.
* Repeat the previous two steps.

If we only use `network-only` without the prefetch, our transitions are delayed by the server response time. This is compounded by also loading assets after so it was visibly slow.
If we only use `cache-first` our very first request will be our last widget from page close with old time/weather data.

app/javascript/components/get-weather-query.jsx
```diff
@@ -43,49 +43,56 @@ export const WeatherQuery = ({
...
-    variables={{ cityName }}
-  >
-    {({ loading, error, data, subscribeToMore }) => {
-      if (loading) {
-        return <LoadingMessage />;
-      }
-      if (error) {
+  resetTimer,
+}) => {
+  return (
```
cbarber, line 48: Do we need to switch from the implicit return to an explicit return? If so, lets make that a refactor that precedes the actual change

app/javascript/components/get-weather-query.jsx
```diff
@@ -43,49 +43,56 @@ export const WeatherQuery = ({
...
-  <Query
-    query={getLocationWeather(WeatherFrag, queryName)}
-    variables={{ cityName }}
-  >
-    {({ loading, error, data, subscribeToMore }) => {
-      if (loading) {
-        return <LoadingMessage />;
-      }
-      if (error) {
+  resetTimer,
```
cbarber, line 46: Maybe a better name? `resetTimer` had me thinking we reset the timer back to 0 when we're completed/errored but I'm guessing this call starts the transition timer.

app/javascript/components/widget-display.jsx
```diff
@@ -72,23 +72,19 @@ export const WidgetDisplay = ({
...
-  }, [current.id]);
+  const resetTimer = () => {
+    dispatch({ type: 'startTimer', durationSeconds: widgetTransitionTime });
+  };
 
   useEffect(() => {
     const timerId = setInterval(() => {
       const { durationRemaining, start } = state;
-      if (isPaused(state)) {
+      if (isPaused(state) || !durationRemaining) {
```
cbarber, line 82: What does `!durationremaining` mean here? Could we name/extract that as a selector so its self documenting?

app/javascript/components/widgets/twitter/panel.jsx
```diff
@@ -61,54 +61,16 @@ export const TabBar = () => (
...
-export default () => (
-  <Query query={getMojoTweets}>
+const Twitter = ({ resetTimer }) => (
+  <Query
+    query={getMojoTweets}
+    onCompleted={() => {
+      resetTimer();
+    }}
+    onError={() => {
+      resetTimer();
```
cbarber, line 71: Do we need to wrap resetTimer in another function here?

app/javascript/components/get-weather-query.jsx
jennyckaplan: Maybe back to `startTimer`, since that's the action that this function is dispatching?

chewnoill:
> _:beep: :boop:_ needs a rebase
> 
> :robot:

app/javascript/components/widget-display.jsx
```diff
@@ -72,23 +72,19 @@ export const WidgetDisplay = ({
...
   });
 
-  useEffect(() => {
-    dispatch({
-      type: 'startTimer',
-      durationSeconds: widgetTransitionTime,
-    });
-  }, [current.id]);
+  const resetTimer = () => {
+    dispatch({ type: 'startTimer', durationSeconds: widgetTransitionTime });
```
chewnoill, line 76: nit: not really a reset action, since we're not starting on mount anymore.

app/javascript/components/widget-display.jsx
chewnoill: 👍 this is a good idea. I think the idea we want is:

```js
function selectIsTimerRunning(state){
  return !isPaused(state) && state.durationRemaining
}
```

chewnoill (commented):
Couple nits, nothing major. Kind of sucks that we have to pass these callback functions around so much. Adding Redux is looking better and better.

app/javascript/components/widgets/twitter/panel.jsx
```diff
@@ -61,54 +61,16 @@ export const TabBar = () => (
...
-          <React.Fragment key={tweet.createdAt}>
-            <Tweet tweet={tweet} isPrimary={false} />
-            <TabBar />
-          </React.Fragment>
-        ))}
-      </PreviousWrapper>
-    </PanelWrapper>
-  );
-};
-
```
tucker-weed, line 82: Nice, I think combining this Twitter component with its parent component below is much clearer because the parent component isn't named (the parent component was the anonymous function `export default () => ( ...`)

bbazyk:
ready for re-review! @chewnoill @cbarber

app/javascript/components/widget-display.jsx
```diff
@@ -101,6 +103,7 @@ export const LoadingDisplay = ({ cityName, loading }) => (
       <LoadingMessage />
     </FullPanel>
     <SidePanel
+      startTimer={() => {}}
```
cbarber, line 106: Why are these noops?

app/javascript/components/widgets/numbers/panel.jsx
```diff
@@ -134,8 +134,17 @@ SubscribedEvents.propTypes = {
...
 
-const Numbers = () => (
-  <Query query={getEventCounts} variables={{ after: getStartOfWeek() }}>
+const Numbers = ({ startTimer }) => (
+  <Query
+    query={getEventCounts}
+    variables={{ after: getStartOfWeek() }}
+    onCompleted={() => {
+      startTimer();
+    }}
```
cbarber, line 143: Do we need to wrap startTimer in an anonymous function?

app/javascript/components/widgets/twitter/panel.jsx
```diff
@@ -61,54 +61,8 @@ export const TabBar = () => (
...
-        link: PropTypes.string,
-      }),
-      user: PropTypes.shape({
-        name: PropTypes.string.isRequired,
-        handle: PropTypes.string.isRequired,
-        avatar: PropTypes.string.isRequired,
-      }).isRequired,
-    }),
-  ).isRequired,
-};
```
cbarber, line 108: This looks like a refactor in the middle of a feature.
* Why are we consolidating these two components?
* If we need to consolidate the components, we should split them into separate commits.

cbarber (changes requested):

app/javascript/components/widgets/twitter/panel.jsx
bbazyk: I've removed all aspects of the refactor now besides references to startTimer, and renaming of the default component - it's in a separate commit from the other changes.

cbarber: If you're undoing changes you no longer, they should be squashed in to the originating commit. Otherwise the edit/undo is just noise.

If you're renaming the default component, that's a refactor. And it should precede your feature commit as a separate unit of work.

app/javascript/components/widget-transition-controls.jsx
```diff
@@ -13,6 +13,8 @@ export const selectShouldRequestPrefetch = state =>
     prefetchDelay;
 
 export const selectIsPaused = state => state.startTimestamp === null;
+export const isStarted = state =>
+  state.startTimestamp !== null && state.msRemainingSinceLastStart !== null;
```
cbarber, line 17: Can you add some unit tests for this?

app/javascript/components/get-weather-query.jsx
```diff
@@ -96,6 +100,11 @@ WeatherQuery.propTypes = {
   }).isRequired,
   queryName: PropTypes.string.isRequired,
   cityName: PropTypes.string.isRequired,
+  startTimer: PropTypes.func,
+};
+
+WeatherQuery.defaultProps = {
+  startTimer: () => {},
```
cbarber, line 107: IIRC, this noop is used for the weatherCorner. Could we explicitly pass a `startTimer={() => {}}` instead of relying on the default? It makes it seem more intentional imo.

app/javascript/components/widget-transition-controls.test.jsx
```diff
@@ -14,6 +15,12 @@ describe('isPaused', () => {
   });
 });
 
+describe('msRemaining', () => {
+  it('returns true if the milliseconds remaining is any non-null value', () => {
+    expect(isStarted({ msRemainingSinceLastStart: null })).toEqual(false);
+  });
```
cbarber, line 21: So we're describing the `isStarted` method. Our `it(...)` doesn't match up with the `toEqual` assertion.

Its fine to squash the unit test into the `isStarted` addition.

cbarber (commented):
This is a great improvement. Thank you splitting up the commits so they're easier to understand.

I think there is a little more work left to be done with unit testing `isStarted`. We also talked about adding null checks and testing for the other selectors Tucker wrote to help catch future bugs.

app/javascript/components/widget-transition-controls.test.jsx
cbarber: I'd like to see the unit tests in the commit that introduce `isStarted`

app/javascript/components/widget-transition-controls.test.jsx
```diff
@@ -54,6 +81,16 @@ describe('shouldRequestPrefetch', () => {
...
   });
+
+  it('returns true, msRemaining as null will still trigger request prefetch (possible error from editing useEffect execution stop)', () => {
+    expect(
+      selectShouldRequestPrefetch({
+        msRemainingSinceLastStart: null,
+        startTimestamp: Date.now() - 21000,
+        prefetchRequestedYet: false,
+      }),
+    ).toEqual(true);
```
cbarber, line 92: I disagree with this assertion. I think `selectShouldRequestPrefetch` should either return false (or throw an exception) if either `msRemainingSinceLastStart` or `startTimestamp` are null. Both are an invalid state for calling this function.

cbarber: I think adding the null checking/throwing and this test should be a separate commit that precedes your feature commit. Something like:
* refactor: rename the default twitter export
* fix: add isStarted check to prevent incorrectly calling `selectShouldRequestPrefetch` with a null value
* fix: disallow calling `selectShouldRequestPrefetch` with null `msRemaining`/`startTimestamp`
* feat: start the sidebar timer after the widget's data is loaded

cbarber:
@bbazyk this isn't passing CI and is not properly rebased (empty commit, tests still split into the wrong commits, fix commit doesn't actually fix the bug). Ping me when you have it fixed.

app/javascript/components/widget-transition-controls.jsx
```diff
@@ -41,8 +41,8 @@ const reducer = (state, action) => {
       };
     case 'startTimer':
       return {
-        msRemainingSinceLastStart: action.msDurationSeconds,
-        startTimestamp: selectIsPaused(state) ? null : now,
+        msRemainingSinceLastStart: action.msDurationRemaining,
```
jennyckaplan, line 48: I'm confused by this commit - what does it mean to change `msDurationSeconds` to `msDurationRemaining`, and why are we getting rid of `selectIsPaused(state)` for determining the `startTimestamp`? and by not dispatching the `startTimer` action within `useEffect` of this component, aren't we making a functionality change and not a refactor?

bbazyk: The variable name 'milliseconds duration seconds' must just have been a typo in a previous refactor, and I thought msDurationRemaining was more appropriate anyway. For 'selectIsPaused(state) ? null : now', Craig and I decided it's a fair assumption that a widget transition will never occur during a pause, and have to start the timer. I decided on labeling the commit a refactor since the startTimer dispatch wasn't removed from this PR - but instead moved to the final feature commit which creates and passes startTimer.

app/javascript/components/widget-transition-controls.jsx
```diff
@@ -41,8 +41,8 @@ const reducer = (state, action) => {
...
+        startTimestamp: now,
```
jennyckaplan, line 49: Should this be `Date.now()`?

app/javascript/components/widget-transition-controls.jsx
jennyckaplan: Ok, I think this commit would benefit from a more descriptive commit message. Regarding the `selectIsPaused(state)` change, can't we transition widgets using the arrow keys, even when paused? So widget transitions can occur while paused? Would this result in a timer starting on the next widget while paused if transitioning via arrow keys?

bbazyk: Yes, the timer does start on the next widget, but it is not paused. I've tested transitioning widgets from a paused widget, and couldn't find any other possible errors caused by this action. When helios is paused, and then manually transitioned, the next widget loads just as fine before calling startTimer, and then proceeds unpaused. I guess it depends if we want to consider 'pausing, transitioning, and helios automatically unpausing itself' a bug or a feature?

bbazyk: I do consider this to be convenient though.

jennyckaplan: This commit is then changing the behavior (if we're paused, we continue to stay paused until resuming, even if transitioning widgets), so it should be in a different commit if that is your intention in changing that functionality. I personally think we should keep the current functionality and stay paused until we resume.

bbazyk (commented):
The setInterval 'refactor-reversion' commit seeks to reverse changes made by an earlier refactor. These changes produced a bug (exposed by implementing startTimer) which could cause widgets to skip over themselves on loading.

bbazyk:
ready for review ! @cbarber

jennyckaplan (approved):
LGTM

app/javascript/components/widget-transition-controls.jsx
```diff
@@ -75,7 +75,7 @@ const WidgetTransitionControls = ({
   const widgetTransitionTime = current.durationSeconds * 1000;
 
   const [state, dispatch] = useReducer(reducer, {
-    startTimestamp: Date.now(),
+    startTimestamp: null,
```
planetmcd, line 80: What bug is this fixing?

bbazyk: startTimestamp's value isn't meant to be set in this reducer (and wasn't until a recent refactor PR which accidentally changed it). These changes produced a bug (exposed by implementing startTimer) which could cause widgets to skip over themselves.

planetmcd: This would be good to include in the commit message.  A format like this:

```
Main Description

Detail for a review that would be helpful for them to understand why you making the changes you are are without them having to run it and recreate.
```

bbazyk: This does look a lot better. I think I'm going to write a lot more commits like this going forward. Thanks, Mike

planetmcd:
:shipit:

PR #272, opened 2021-06-30
add Boulder SVG and its styling based on location

fallynlogan:
Add Boulder SVG which displays when Boulder is the URL path (ie http://localhost:5000/Boulder)

Screen Shot 2021-07-02 at 1 26 24 PM


app/javascript/components/widgets/weather/semi-circle.jsx
```diff
@@ -1,12 +1,20 @@
...
+import pvdImage from '@images/buildings.png';
+import bldImage from '@images/boulder.svg';
 import { yValue, xValue } from '@lib/circle-math';
 import { timeDiffInMinutes, timeAndDateForTimezone } from '@lib/datetime';
 import { colors } from '@lib/theme';
 import styled, { keyframes } from 'styled-components';
 import MoonPhase from '@weather/moon-phase';
 
+const getCityImage = () => {
+  if (window.location.pathname.includes('Boulder')) return bldImage;
```
cbarber, line 12: You'll want to avoid reaching into `window.location` and use props instead. This:
* minimize using browse specific calls so we can test components. You would need to mock `window.location.pathname` to run this component in a test
* allows Server Side Rendering where `window` is not available. Not necessary here but good to be aware of.

app/javascript/components/widgets/weather/semi-circle.jsx
```diff
@@ -125,24 +133,24 @@ export class SemiCircle extends React.Component {
     100% { offset-distance : ${arcEndPercent}% }
     `;
 
+    const styleSemiCircle = () => {
+      if (nightMode) return colors.lightBlue;
+      if (cityImage === bldImage) return colors.lightYellow;
```
cbarber, line 138: This should be its own refactor commit to extract the function and then add the boulder condition as part of the boulder commit.

app/javascript/components/widgets/weather/semi-circle.jsx
```diff
@@ -125,24 +133,24 @@ export class SemiCircle extends React.Component {
...
           </radialGradient>
         </defs>
         <defs>
           <linearGradient id="strokeGradient">
             <stop offset="2%" stopColor={colors.black} />
-            <stop
-              offset="50%"
-              stopColor={nightMode ? colors.blue : colors.pink}
-            />
+            <stop offset="50%" stopColor={styleSemiCircle()} />
```
cbarber, line 153: There's a subtle bug here with the color refactor. Maybe its worthwhile to define our day/night colors per component?

cbarber (changes requested):
Nice! I think the commits could use some restructuring and you have a regression with the original colors. I *think* we have snapshot testing now so this might be a good time to add some snapshots for this component before refactoring/adding features.

The image a little smaller than the pvd city line.

app/javascript/components/widgets/weather/semi-circle.jsx
```diff
@@ -1,12 +1,20 @@
...
 import { colors } from '@lib/theme';
 import styled, { keyframes } from 'styled-components';
 import MoonPhase from '@weather/moon-phase';
 
+const getCityImage = () => {
+  if (window.location.pathname.includes('Boulder')) return bldImage;
+  return pvdImage;
+};
+
+const cityImage = getCityImage();
```
chewnoill, line 16: This is a problem. This evaluates what the cityImage is at load time. I think we need to evaluate this **every time the route changes**.

app/assets/images/boulder.svg
```diff
@@ -0,0 +1 @@
...s="cls-2" x="876.95" y="292.66" width="7.78" height="6.02"/><rect class="cls-2" x="796.64" y="307.89" width="7.78" height="6.02"/><rect class="cls-2" x="812.27" y="307.63" width="7.78" height="6.02"/><rect class="cls-2" x="828.45" y="307.76" width="7.78" height="6.02"/><rect class="cls-2" x="845.08" y="307.5" width="7.78" height="6.02"/><rect class="cls-2" x="861.25" y="307.43" width="7.78" height="6.02"/><rect class="cls-2" x="876.95" y="307.66" width="7.78" height="6.02"/><rect class="cls-2" x="796.64" y="322.89" width="7.78" height="6.02"/><rect class="cls-2" x="812.27" y="322.63" width="7.78" height="6.02"/><rect class="cls-2" x="828.45" y="322.76" width="7.78" height="6.02"/><rect class="cls-2" x="845.08" y="322.5" width="7.78" height="6.02"/><rect class="cls-2" x="861.25" y="322.43" width="7.78" height="6.02"/><rect class="cls-2" x="876.95" y="322.66" width="7.78" height="6.02"/><rect class="cls-10" x="897.22" y="278.66" width="7.78" height="6.02"/><rect class="cls-10" x="896.37" y="292.66" width="7.78" height="6.02"/><rect class="cls-10" x="896.37" y="307.66" width="7.78" height="6.02"/><rect class="cls-10" x="896.37" y="322.66" width="7.78" height="6.02"/></g></g></g></svg>
```
tucker-weed, line 1: Did you generate this svg from an image file?

fallynlogan: I made it in illustrator and that lets you export as an SVG!

app/javascript/components/widgets/weather/semi-circle.jsx
fallynlogan: This should be fixed now!

app/javascript/components/widgets/weather/semi-circle.jsx
fallynlogan: This should be fixed now!

app/javascript/components/widgets/weather/semi-circle.jsx
fallynlogan: This should be fixed now!

app/javascript/components/widgets/weather/semi-circle.jsx
fallynlogan: This should be fixed now!

app/javascript/lib/theme.js
```diff
@@ -5,7 +5,9 @@ export const colors = {
   lightPink: '#FFA7C2',
   pink: '#FE6B6E',
   teal: '#5fc19f',
+  lightYellow: '#FAE7A7',
   yellow: '#fbd273',
+  yellow2: '#CDBD64',
```
cbarber, line 10: `yellow2` doesn't really tell us much so I'd like to avoid. I'm curious, where do we use `pink`, `yellow`, etc through-out the application? Are they actually shared color codes or are the specific to a single component? Would it make more sense to define those colors in their specific components rather than co-locating them?

app/javascript/components/widgets/weather/semi-circle.jsx
```diff
@@ -1,12 +1,30 @@
...
+offset-path: path('${p => p.arcPath}');
+animation: ${p => p.arcAnimation} 1000ms 1 normal forwards ease-in-out;
+r: ${p => p.ballRadius};
+fill: ${colors.pink}`;
+
+const BldStyle = styled.circle`
+offset-path: path('${p => p.arcPath}');
+animation: ${p => p.arcAnimation} 1000ms 1 normal forwards ease-in-out;
+r: ${p => p.ballRadius};
+fill: ${colors.yellow2}`;
```
cbarber, line 26: Nit: you could still have one component and pass in the fill color as a property.

app/javascript/components/widgets/weather/semi-circle.jsx
```diff
@@ -125,16 +144,26 @@ export class SemiCircle extends React.Component {
...
+
     const styleRadialGradient = () => {
       if (nightMode) return colors.lightBlue;
+      if (cityImage === bldImage) return colors.lightYellow;
       return colors.lightPink;
     };
 
     const styleLinearGradient = () => {
       if (nightMode) return colors.blue;
+      if (cityImage === bldImage) return colors.yellow2;
```
cbarber, line 157: How does BLD look in night mode?

app/javascript/lib/theme.js
fallynlogan: `pink` & `yellow` are both used in `panel.jsx` (for styling the slack and github information) and `pink` is also used in `semi-circle.jsx` to style the pvd buildings. `blue` and `lightBlue` are only used in the semi-circle component. I think it makes sense to define `blue`,`lightBlue`,`yellow`, and `yellow2` (with a more descriptive name) in their specific components and leave `pink` as shared. What do you think about that?

app/javascript/components/widgets/weather/semi-circle.jsx
fallynlogan: That makes a lot of sense, this should be fixed now!

app/javascript/components/widgets/weather/semi-circle.jsx
fallynlogan: Heres a screenshot of what it looks like. I think it looks similar to PVD in night mode, but can make adjustments if needed! 

Screen Shot 2021-07-06 at 12 04 28 PM

app/javascript/lib/theme.js
cbarber: Sounds good to me.

app/javascript/components/widgets/weather/semi-circle.jsx
cbarber: :+1:

app/javascript/components/widgets/weather/semi-circle.jsx
```diff
@@ -1,12 +1,26 @@
...
+  if (cityName === 'Boulder') return bldImage;
+  return pvdImage;
+};
+
+const getCityStyle = fillStyle => {
+  return styled.circle`
+  offset-path: path('${p => p.arcPath}');
+  animation: ${p => p.arcAnimation} 1000ms 1 normal forwards ease-in-out;
+  r: ${p => p.ballRadius};
+  fill: ${fillStyle}`;
```
cbarber, line 21: The `${p => p....}` above is the properties pattern from styled components. We can just extend that instead of adding another function call.

fallynlogan: This should be fixed now!

app/javascript/components/widgets/weather/semi-circle.jsx
```diff
@@ -34,6 +34,9 @@ const getOrbPercent = (endTime, totalTime, offset) => {
   return Math.floor(rangePercent);
 };
 
+const blue = '#6bd9fe';
+const lightBlue = '#8096ff';
```
chewnoill, line 44: We should keep colors in the theme object. If we need to reference them we should `import` the values.

Here is an example
https://github.com/mojotech/helios2/blob/ffebc0004c15fd93eca79bc6007db6d0c9397229/app/javascript/components/panel-wrapper.jsx#L1-L13

app/javascript/components/widgets/weather/semi-circle.jsx
```diff
@@ -207,6 +221,6 @@ const SunCircle = styled.circle`
 offset-path: path('${p => p.arcPath}');
 animation: ${p => p.arcAnimation} 1000ms 1 normal forwards ease-in-out;
 r: ${p => p.ballRadius};
-fill: ${colors.pink}`;
+fill: ${p => p.fillStyle}`;
```
chewnoill, line 224: I think for svgs we can `fill: currentColor` and we will keep the color from the parent element.

https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentcolor_keyword

app/javascript/components/widgets/weather/semi-circle.jsx
```diff
@@ -36,6 +42,8 @@ const getOrbPercent = (endTime, totalTime, offset) => {
 
 const blue = '#6bd9fe';
 const lightBlue = '#8096ff';
+const yellow = '#fbd273';
+const lightYellow = '#FAE7A7';
```
chewnoill, line 46: lets keep all new colors in the theme also

cbarber: I asked that they be split out. These are more related to the specific component than an actual theme which lead to confusing/unhelpful names like `yellow2`.

app/javascript/components/widgets/weather/semi-circle.jsx
cbarber: https://github.com/mojotech/helios2/pull/272#discussion_r664732264

app/javascript/components/widgets/weather/semi-circle.jsx
chewnoill: Colors should be consistent across the entire app. It doesn't make sense to have single use colors. Can we change the PVD palette to match the Boulder one?

`lightYellow` seems fine, certainly better than `yellow2`. Another option would be to name the color after its function, and not what it looks like (primary/secondary/etc)

cbarber: Check out the original changes. There's `yellow`, `lightYellow` and `yellow2`. Fallyn discovered most of the colors are in fact specific to this component so there's not much advantage to co-locating them for the sake of co-locating them. We could namespace in the theme with something like:
```
colors: {
  city: {
    yellow: ...
    lightYellow: ...
  }
  yellow: ...
}
```
but now we're back to why should we co-locate colors that only related to the city image in a theme file?

Changing the color palette of PVD feels out of scope of creating a new BLD image.

chewnoill: The purpose of colocating theme colors is so that they can be changed easily. Maybe plug them into the theme with the city name, and we can drop all the extra logic. How many colors does this component need?

```js
colors: {
  Boulder: [yellow1, yellow2, yellow3],
  Providence: [...]
  ...
}
```

Would be nice to extract the colors from this SVG also, but it looks like our other images are PNGs anyway so it wouldn't be super helpful.

cbarber: We're back to colocating colors that are used in one place. How about:
```
const Boulder = { /* colors */ };
const Default = { /* colors */ };
```
at the top of the component?

cbarber: And those would be something like:
```
const Boulder = { sun: "...", moon: "...", foreground: "....", background: "...."  };
```

chewnoill: > We're back to colocating colors

Yes, exactly this. This is what I'm asking for.

fallynlogan: This should be fixed now!

cbarber (approved):
LGTM nice work :+1:

chewnoill (approved):
🚀

PR #276, opened 2021-07-12
 feat:display an error message when widget is  empty array

sandipa12:
Display an error message "No widget enabled" if a widgets query for a location returns with an empty array of widgets.
https://www.pivotaltracker.com/story/show/178526431
Screen Shot 2021-07-13 at 11 46 30 AM

### DIFF
--- a/app/javascript/components/widget-controller.jsx
+++ b/app/javascript/components/widget-controller.jsx
@@ -2,10 +2,12 @@ import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import gql from 'graphql-tag';
 import { Query } from 'react-apollo';
+import { isEmpty, pathOr } from 'ramda';
 import {
   WidgetDisplay,
   LoadingDisplay,
   DisconnectedDisplay,
+  WidgetDisabledDisplay,
   getWidgetDisplay,
 } from '@components/widget-display';
 import WidgetTransitionControls from '@components/widget-transition-controls';
@@ -71,6 +73,11 @@ export const WidgetController = ({ client, cityName }) => {
           // eslint-disable-next-line
           console.error(error);
           return <DisconnectedDisplay cityName={cityName} error={error} />;
+        }
+
+        const widget = pathOr([], ['location', 'widgets', 'enabled'], data);
+        if (isEmpty(widget)) {
+          return <WidgetDisabledDisplay />;
         }
 
         return (
--- a/app/javascript/components/widget-display.jsx
+++ b/app/javascript/components/widget-display.jsx
@@ -2,7 +2,11 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import gql from 'graphql-tag';
 import styled from 'styled-components';
-import { DisconnectedMessage, LoadingMessage } from '@messages/message';
+import {
+  DisconnectedMessage,
+  LoadingMessage,
+  WidgetDisabledMessage,
+} from '@messages/message';
 import FullPanel from '@components/full-panel';
 import SidePanel, { getSidePanel } from '@components/side-panel';
 import Twitter from '@widgets/twitter';
@@ -149,4 +153,18 @@ DisconnectedDisplay.defaultProps = {
   error: false,
 };
 
+export const WidgetDisabledDisplay = () => (
+  <Wrapper>
+    <FullPanel>
+      <WidgetDisabledMessage />
+    </FullPanel>
+    <SidePanel
+      widgets={[]}
+      selectedWidgetId={0}
+      showWeather={false}
+      totalTime={0}
+      isPaused={false}
+    />
+  </Wrapper>
+);
 export default withFragment(WidgetDisplay);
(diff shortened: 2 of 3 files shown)


app/javascript/components/widget-display.jsx
```diff
@@ -147,4 +151,18 @@ DisconnectedDisplay.defaultProps = {
   error: false,
 };
 
+export const WidgetDisabledDisplay = () => (
+  <Wrapper>
+    <FullPanel>
+      <WidgetDisabledMessage />
+    </FullPanel>
+    <SidePanel
```
allisonlucas, line 159: Do we need to display the SidePanel? Can you include a screenshot of what the "No widget enabled" message looks like?

fallynlogan (approved):
Looks good to me! :)

app/javascript/components/widget-display.jsx
allisonlucas: We paired on this and determined that the SidePanel should be included :+1:. Please do add a screenshot to the PR description once @cbarber lets you know how to get an empty array returned for the widgets!

sandipa12:
@fallynlogan , the second commit appeared when I tried updating the branch with the base branch. I should probably fixup.

app/javascript/components/widget-display.jsx
sandipa12: @allisonlucas  sure.

app/javascript/components/widget-controller.jsx
```diff
@@ -72,6 +73,9 @@ export const WidgetController = ({ client, cityName }) => {
           console.error(error);
           return <DisconnectedDisplay cityName={cityName} error={error} />;
         }
+        if (data.location.widgets.enabled === []) {
+          return <WidgetDisabledDisplay />;
```
cbarber, line 77: :thinking: Can we ever leave the `WidgetDisabledDisplay` path if widgets are enabled for the city?

app/javascript/components/widget-controller.jsx
```diff
@@ -72,6 +73,9 @@ export const WidgetController = ({ client, cityName }) => {
           console.error(error);
           return <DisconnectedDisplay cityName={cityName} error={error} />;
         }
+        if (data.location.widgets.enabled === []) {
```
cbarber, line 76: :thinking: we probably should access this path safely in case data is null / empty. Have a look at ramdajs `path` / `pathOr`

cbarber (changes requested):

chewnoill (approved):

app/javascript/components/widget-controller.jsx
chewnoill: this is a good idea

app/javascript/components/widget-controller.jsx
```diff
@@ -73,6 +75,20 @@ export const WidgetController = ({ client, cityName }) => {
...
+        if (
+          R.pathOr(
+            data.location.widgets.enabled === [],
+            ['NullPointerhandler', 'condition'],
+            {
+              NullPointerhandler: {
+                conditon: data.location.widgets.enabled === [],
+              },
+            },
+          )
```
cbarber, line 87: `pathOr` is intended to accept the arguments: defaultValue, arrayPath, object. An example:
```
const something = { a: { b: { c: 6 } } };
console.log(pathOr(2, ['a', 'b', 'c'], something)); // 6
console.log(pathOr(2, ['a', 'b', 'd'], something)); // 2
console.log(pathOr(2, ['a', 'b', 'c'], null)); // 2
```
The purpose is to safely access values in an nested object that may or may not contain null/undefined. The code above can still have a null pointer exception if data is null.

app/javascript/components/widget-controller.jsx
```diff
@@ -2,10 +2,12 @@ import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import gql from 'graphql-tag';
 import { Query } from 'react-apollo';
+import * as R from 'ramda';
```
cbarber, line 5: Nit: you can just import `pathOr` by destructuring instead of all functions:
`import { pathOr } from 'ramda';`

app/javascript/components/widget-controller.jsx
```diff
@@ -73,6 +75,10 @@ export const WidgetController = ({ client, cityName }) => {
           return <DisconnectedDisplay cityName={cityName} error={error} />;
         }
 
+        if (pathOr(true, ['location', 'widgets', 'enabled'], data) === true) {
```
cbarber, line 78: This will return false if data is `{ location: { widgets: { enabled: [] } } }`.

cbarber (approved):
Nice work with ramda. I'd still like WidgetDisabledDisplay to have some mechanism to continue polling the widgets from the server to handle widgets being added (rather than requiring a page refresh), but don't want to hold this PR up. I'll make a separate ticket for that functionality.

PR #277, opened 2021-07-12
chore: rewrite withFragment

tmsavage:
The beginning of rewriting withFragment. First component to work with bathroom.jsx

### DIFF
--- a/app/javascript/components/bathroom.jsx
+++ b/app/javascript/components/bathroom.jsx
@@ -52,8 +52,4 @@ Bathroom.defaultProps = {
   error: false,
 };
 
-Bathroom.fragments = {
-  location: getBathroomCode,
-};
-
-export default withFragment(Bathroom);
+export default withFragment(Bathroom, { location: getBathroomCode });
--- a/app/javascript/components/widgets/weather/sunrise-sunset.jsx
+++ b/app/javascript/components/widgets/weather/sunrise-sunset.jsx
@@ -14,7 +14,7 @@ import { parseTime, timeDiffInMinutes } from '@lib/datetime';
 import { WhiteText } from '@components/typography';
 import withFragment from '@hocs/with-fragment';
 import SemiCircle from '@weather/semi-circle';
-import WeatherEffect from '@weather/weather-effect';
+import WeatherEffect, { getWeatherEffect } from '@weather/weather-effect';
 
 const containerHeight = '344px';
 
@@ -51,18 +51,17 @@ const SunsetLabel = styled.div`
   z-index: 2;
 `;
 
-const getSunriseSunsetWeather = gql`
+export const getSunriseSunsetWeather = gql`
+  ${getWeatherEffect}
   fragment SunriseSunsetWeather on Weather {
-    moonPhase
     ...WeatherEffect
   }
-
-  ${WeatherEffect.fragments.weather}
 `;
 
-const getSunriseSunsetLocation = gql`
+export const getSunriseSunsetLocation = gql`
   fragment SunriseSunsetLocation on Location {
     timezone
+    moonPhase
     cityName
     solarCycles {
       type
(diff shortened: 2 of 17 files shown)


app/javascript/components/bathroom.jsx
```diff
@@ -1,16 +1,15 @@
...
+// import withFragment from './hocs/with-fragment';
+import withFragmentModify from './hocs/with-fragment-modify';
 
-const getBathroomCode = gql`
-  query getBathroomCode($cityName: String!) {
-    location(cityName: $cityName) {
-      bathroomCode
-    }
+export const getBathroomCode = gql`
+  fragment Bathroom on Location {
```
cbarber, line 11: :thinking: Did we miss updating the bathroom code component in the last pass?

cbarber: I checked `master` and this change is already there. This might need a rebase

app/javascript/components/bathroom.jsx
```diff
@@ -1,16 +1,15 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Query } from 'react-apollo';
 import gql from 'graphql-tag';
 import styled from 'styled-components';
 import { WhiteSubTitle, WhiteTitleLarge } from '@components/typography';
 import { LoadingMessage, ErrorMessage } from '@messages/default-messages';
+// import withFragment from './hocs/with-fragment';
+import withFragmentModify from './hocs/with-fragment-modify';
```
cbarber, line 8: Lets change the definition of  the existing `withFragment` module rather than adding a new one with a slightly different name.

app/javascript/components/hocs/with-fragment-modify.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+
+  // TODO: add renderWhileLoading & renderWhileError
+  const render = WrappedComponent.propTypes;
+  switch(render){
+    case render.loading:
+      return Enhanced;
+    case render.error:
+      return null;
+  }
```
cbarber, line 35: Don't worry about `loading` / `error` conditions here.

app/javascript/components/bathroom.jsx
tmsavage: Bathroom was updated in the last pass. A rebase is necessary.

app/javascript/components/bathroom.jsx
tmsavage: Yes, this is just a draft PR and does not yet contain these fixes. Creating a new HOC was for testing purposes for the bathroom component only. I have a working version of `withFragment` and am implementing it with the remaining components.

app/javascript/components/hocs/with-fragment-modify.jsx
tmsavage: Removed 👍🏼

fallynlogan (commented):
Why are we updating withFragment?

tmsavage:
> Why are we updating withFragment?

We are rewriting withFragment to make it clear what its purpose is as a HOC. This ultimately removes a function that copies over statics between different components. Instead, withFragment directly receives fragments from one component, processes it, and exports.

app/javascript/components/widgets/weather/sunrise-sunset.jsx
```diff
@@ -71,8 +70,7 @@ const getSunriseSunsetLocation = gql`
 `;
 
 export const SunriseSunset = ({ location, weather }) => {
```
bencates, line 72: remove `weather` here since we're not using it anymore

app/javascript/components/widgets/weather/sunrise-sunset.jsx
```diff
@@ -128,21 +126,17 @@ export const SunriseSunset = ({ location, weather }) => {
...
     solarCycles: PropTypes.arrayOf(
       PropTypes.shape({
         type: PropTypes.string.isRequired,
         time: PropTypes.string.isRequired,
       }),
     ).isRequired,
   }).isRequired,
-  weather: PropTypes.shape({
-    moonPhase: PropTypes.number.isRequired,
-  }).isRequired,
```
bencates, line 140: I think you'll still need `weather: PropTypes.shape({})`

bencates (approved):

PR #279, opened 2021-07-12
feat: add users table

tucker-weed:
Description

1. Adds users table migration
2. Creates skeleton definition for user graphql type

### DIFF
--- a/db/migrate/20210712211847_create_users.rb
+++ b/db/migrate/20210712211847_create_users.rb
@@ -1,0 +1,10 @@
+class CreateUsers < ActiveRecord::Migration[5.2]
+  def change
+    create_table :users do |t|
+      t.string :user_name, null: false
+      t.text :public_key, null: false
+
+      t.timestamps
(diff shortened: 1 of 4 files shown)


db/migrate/20210712211847_create_users.rb
```diff
@@ -0,0 +1,10 @@
+class CreateUsers < ActiveRecord::Migration[5.2]
+  def change
+    create_table :users do |t|
+      t.string :user_name, null: false
+      t.string :public_key, null: false
```
cbarber, line 5: Nit: in sqlite3 and postgresql, the rails migration type `string` defaults to `varchar(255)`. `text` defaults to a text blob which can be any size string. public_key will likely be greater than 255 characters.

Historically, varchar vs char vs blob had performance considerations but in modern databases like postgresql the performance is the same and its just a size constraint.

cbarber (approved):
Nice work :+1:

PR #280, opened 2021-07-13
add mapbox interactive map for RIPTA transit api widget

bbazyk:
ticket id #166190214
Screenshot from 2021-07-13 00-40-29
Currently adds widget for Providence buses, contains mapbox gl map which displays pins based on json data from the RIPTA Transit API - http://realtime.ripta.com:81/ 
This preliminary PR has plans to include:
>Custom svgs to represent buses + the title of their route
>An additional api retrieval and description of important service alerts
>A more zoomed in view of just the Kennedy Plaza buses outside of the PVD Office (less clutter for the viewer)
>A reworked display wherein arrival and departure times of the buses takes center stage over the mapbox map (which still exists for visual purposes)


app/javascript/components/widget-controller.jsx
```diff
@@ -6,15 +6,12 @@ import {
...
   DisconnectedDisplay,
-  getWidgetDisplay,
 } from '@components/widget-display';
 import WidgetTransitionControls from '@components/widget-transition-controls';
 
 const getWidgets = gql`
-  ${getWidgetDisplay}
   query getWidgets($id: Int!, $cityName: String!) {
     location(cityName: $cityName) {
-      ...WidgetDisplay
```
cbarber, line 17: Why was the fragment deleted from the fragment?

app/javascript/components/widgets/buses/panel.jsx
```diff
@@ -0,0 +1,92 @@
...
+  }, []);
+
+  return (
+    <div>
+      <MapStyle />
+      <div className="sidebar">
+        Longitude: {lng} | Latitude: {lat} | Zoom: {zoom}
+      </div>
+      <MapDimensions ref={mapContainer} className="map-container">
+        {' '}
```
cbarber, line 72: Why are these spaces here?

app/javascript/components/widgets/buses/panel.jsx
```diff
@@ -0,0 +1,92 @@
+import React, { useRef, useEffect, useState } from 'react';
+import styled from 'styled-components';
+import mapboxgl from '!mapbox-gl';
+import PropTypes from 'prop-types';
+import geoJson from './chicago-parks.json';
+import MapStyle from './globalMapStyle';
+import vehiclepositions from './vehiclepositions.json';
+
+mapboxgl.accessToken = process.env.MAPBOX_TOKEN;
```
cbarber, line 9: We shouldn't expose API tokens in the frontend like this.

app/javascript/components/widgets/buses/panel.jsx
```diff
@@ -0,0 +1,92 @@
...
+      .setLngLat([
+        feature.vehicle.position.longitude,
+        feature.vehicle.position.latitude,
+      ])
+
+      .addTo(map);
+    return () => marker.remove();
+  });
+
+  return <div ref={markerref} className="marker" />;
```
cbarber, line 30: It looks like we're using `useEffect` to call the mapgl library. How does the ref relatet to the options object the `Marker` method accepts? Why do we connect that to a div? Do we need react here at all?

app/javascript/components/widgets/buses/panel.jsx
```diff
@@ -0,0 +1,92 @@
...
+    </>
+  );
+};
+
+const Buses = () => {
+  const mapContainer = useRef(null);
+  const [mapState, setMap] = useState(null);
+  const [lng] = useState(-71.415);
+  const [lat] = useState(41.825);
+  const [zoom] = useState(15);
```
cbarber, line 50: There is no way to update `lng`, `lat`, or `zoom`. Do we need this state?

app/javascript/components/widgets/buses/panel.jsx
```diff
@@ -0,0 +1,92 @@
...
+  useEffect(() => {
+    const map = new mapboxgl.Map({
+      container: mapContainer.current,
+      style: 'mapbox://styles/mapbox/streets-v11',
+      center: [lng, lat],
+      zoom,
+    });
+
+    map.addControl(new mapboxgl.NavigationControl(), 'top-left');
+    setMap(map);
```
cbarber, line 61: Why are we constructing a map in state?

db/seeds.rb
```diff
@@ -86,6 +86,17 @@
...
+  Widget.find_or_initialize_by(
+    name: "Buses",
+    location: location
+  ) do |r|
+    r.enabled = true
+    r.duration_seconds = 20
+    r.position = 5
+    r.sidebar_text = 'Bus Routes and Positions'
+    r.show_weather = true
+  end.save!
```
cbarber, line 99: How are you adding buses to the database? Did you miss adding a migration and the model? It doesn't look like the frontend is connected to the rails graphql API for this change. I think it would make sense to split out the frontend and back end changes into separate tickets.

.env.local.sample
```diff
@@ -14,3 +14,4 @@ ADMIN_EMAIL=admin@example.com
 ADMIN_PASSWORD= password
 BAMBOOHR_SUBDOMAIN=<Ask a team mate>
 BAMBOOHR_API_KEY=<Ask a team mate>
+MAPBOX_TOKEN=
```
cbarber, line 17: What is the mapbox_token used for?

cbarber (changes requested):

db/seeds.rb
bbazyk: I'm not adding them to the database yet - and I should've made that clear in the PR description. Right now there's a json file taken from an arbitrary time from the above RIPTA transit API site, I was using that formatting to create pins. As it stands, the map cannot update itself.

.env.local.sample
bbazyk: Mapbox requires the user to have a token to use its API at all. The map won't render without one.

bbazyk: Or maybe you're asking why I added part of that assignment line to the env - which is because I thought all keys and tokens were referenced in that file even if the user doesn't have them explicitly in their .env (is that wrong?)

app/javascript/components/widgets/buses/panel.jsx
bbazyk: No, I'll remove it as part of the next commit

app/javascript/components/widget-controller.jsx
bbazyk: I probably accepted the wrong changes during a rebase? :confused: Thanks for pointing that out, I'll fix it in the next commit

app/javascript/components/widget-controller.jsx
```diff
@@ -64,13 +61,13 @@ export const WidgetController = ({ client, cityName }) => {
     >
       {({ loading, error, data }) => {
         if (loading) {
-          return <LoadingDisplay cityName={cityName} loading={loading} />;
+          return <LoadingDisplay cityName={cityName} />;
```
jennyckaplan, line 64: Why is this loading prop deleted here?

app/javascript/components/widget-controller.jsx
```diff
@@ -64,13 +61,13 @@ export const WidgetController = ({ client, cityName }) => {
...
         }
 
         if (error) {
           // eslint-disable-next-line
           console.error(error);
-          return <DisconnectedDisplay cityName={cityName} error={error} />;
+          return <DisconnectedDisplay cityName={cityName} />;
```
jennyckaplan, line 70: Same with deleting this `error` prop, what is happening here?

app/javascript/components/widgets/buses/chicago-parks.json
```diff
@@ -0,0 +1,104 @@
+{
```
jennyckaplan, line 1: This Chicago JSON file should be removed.

app/views/layouts/application.html.erb
```diff
@@ -7,6 +7,7 @@
 
     <%= stylesheet_link_tag    'application', media: 'all', 'data-turbolinks-track': 'reload' %>
     <%= javascript_include_tag 'application', 'data-turbolinks-track': 'reload' %>
+    <link href="https://api.tiles.mapbox.com/mapbox-gl-js/v1.11.1/mapbox-gl.css" rel="stylesheet" />
```
jennyckaplan, line 10: Is this necessary with the `globalMapStyle.jsx` you set up with @ericgibb?

package.json
```diff
@@ -32,6 +34,7 @@
     "react-dom": "17.0.2",
     "react-highlight-words": "0.17.0",
     "react-hls-player": "^3.0.1",
+    "react-map-gl": "^6.1.16",
```
jennyckaplan, line 37: This should be removed since you pivoted to using `mapbox-gl` instead.

jennyckaplan:
Looks like you may have squashed previous commits, so those old messages stayed in your primary commit message - these should be removed. Also, is this map interactive and scrollable? I think for now we might want to have it fixed just around the office, since Helios doesn't accept any user input (at the moment).

.env.local.sample
cbarber: That's right, but the intent is to use the rails API as layer to keep API keys out of the frontend. Leaking them through to the frontend isn't going to be viable. 

Is there a limit for the requests or a pricing plan for the API? How do we go about making new API keys?

bbazyk: The Tokens API I used for free comes at a cap of 100 requests/min, which if exceeded simply returns a "HTTP 429 Too Many Requests" response from the API - and throttles further interaction. Anybody can create a mapbox token api key with only their email address - which is a lot better than the alternative of using the google maps API (my original implementation) where key users must sign up by providing their card number.

bbazyk: Should I include some instructions in the helios README?

cbarber: :thinking: The issue is the API key with the 100 req/min limit will be available in our bundle for anyone to access. If they do that, and hit our cap, our frontend stops working. What is the API used for? Can we pass the request through the rails server so we don't expose the API key?

bbazyk: I will try to pass the request through the rails server, and I intend for (like with the openWeather API key?) everyone to create their own token. I hope I'm understanding your suggestion correctly - does this solve all of the issues?

cbarber: This would solve exposing the API key issue.

Can you link to the details for the API key request limits? There's usually a monthly cap in addition to a req/min cap.

bbazyk: https://docs.mapbox.com/api/accounts/tokens/#list-tokens (bottom of this page for request limits)
https://docs.mapbox.com/accounts/guides/invoices/ - the beginning of this one has additional useful information

app/javascript/components/widget-controller.jsx
```diff
@@ -70,7 +70,7 @@ export const WidgetController = ({ client, cityName }) => {
         if (error) {
           // eslint-disable-next-line
           console.error(error);
-          return <DisconnectedDisplay cityName={cityName} error={error} />;
+          return <DisconnectedDisplay cityName={cityName} loading={loading} />;
```
sandipa12: Why have you changed it to loading? Don't we display "loading"  only when it's loading?

app/views/layouts/application.html.erb
bbazyk: According to the API, yes : https://docs.mapbox.com/mapbox-gl-js/api/
_Include the JavaScript and CSS files in the <head> of your HTML file. The CSS file is required to display the map and make elements like Popups and Markers work._
- <link href='https://api.mapbox.com/mapbox-gl-js/v2.3.1/mapbox-gl.css' rel='stylesheet' />

Without this line included, the map doesn't render properly - and the 'markers' (bus pins) won't appear at all.

app/javascript/components/widget-controller.jsx
bbazyk: Another product of my awful rebase - thanks for pointing it out, I'll  restore the files and lines that didn't need any editing.

PR #281, opened 2021-07-13
fix null value issue for windGust

sandipa12:
The response from openweather API for current weather may have a null value for windGust.  There was an error that mentioned "cannot return null value for null-nullable field" since  our graphql API does not allow a null value in our schema.
https://www.pivotaltracker.com/story/show/178831309


cbarber (approved):
Nice quick fix. Thank you.

I mentioned in the ticket that there are likely other nullable values in the openweather api response. To close the loop on this issue could you also check for values that could be missing and mark them as nullable in the graphql types?

chewnoill (approved):
🚀

PR #282, opened 2021-07-13
fix:mark null in graphql types for nullable values in the openweather api response

sandipa12:
Follow-up task into the other nullable fields from openweathermap API. 
https://www.pivotaltracker.com/story/show/178831309


cbarber (approved):
LGTM :+1:

PR #283, opened 2021-07-14
feat: add employees graphql endpoint

fallynlogan:
Adding employees information from bambooHR API to graphql API endpoint

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -31,4 +31,14 @@ class Types::QueryType < Types::BaseObject
     events = events.with_source(type) if type
     events
   end
+
+  field :employee_events, Types::EmployeeEventCollectionType do
+    description "Employee events from bambooHR"
+  end
+
+  def employee_events
+    # dummy return value for now because cannot return null
+    # will be filled in with database calls in a later PR
+    42
+  end
 end
--- a/app/graphql/types/employee_event_collection_type.rb
+++ b/app/graphql/types/employee_event_collection_type.rb
@@ -1,0 +1,17 @@
+class Types::EmployeeEventCollectionType < Types::BaseObject
+  field :birthday_employees, [Types::EmployeeType]
+  field :anniversary_employees, [Types::EmployeeType]
+  field :time_off_employees, [Types::EmployeeType]
+
+  def birthday_employees
+    Clients::BamboohrClient.find_employees.find_birthday_employees
+  end
+
+  def anniversary_employees
+    Clients::BamboohrClient.find_employees.find_anniversary_employees
+  end
+
+  def time_off_employees
+    Clients::BamboohrClient.find_employees.find_time_off_employees
+  end
+end
--- a/app/graphql/types/employee_type.rb
+++ b/app/graphql/types/employee_type.rb
@@ -1,0 +1,6 @@
+class Types::EmployeeType < Types::BaseObject
+  field "id", ID
+  field "display_name", String, method: :displayName
+  field "is_photo_uploaded", String, method: :isPhotoUploaded
+  field "photo_url", String, method: :photoUrl
+end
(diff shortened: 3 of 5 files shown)


bbazyk:
Looks good! I just have a question about the plan for the events widget - are the goals the same as they were in the original PR with the display illustration? I can see we're still working to show birthdays, anniversaries and PTOs - is the "today in history" feature still a plan for later?

app/graphql/types/employees_type.rb
```diff
@@ -0,0 +1,15 @@
...
+  field "is_photo_uploaded", String
+  field "photo_url", String
+end
+
+class Types::EmployeesType < Types::BaseObject
+  field "anniversary_employees", [Types::EmployeeType]
+  field "birthday_employees", [Types::EmployeeType]
+  field "time_off_employees", [Types::EmployeeType]
+  field "all_employees", [Types::EmployeeType]
+end
```
bencates, line 15: This looks like it's redundant given the `employees(type: String)` query. It should probably be removed.

app/graphql/types/query_type.rb
```diff
@@ -31,4 +31,20 @@ def events(after: nil, type: nil)
     events = events.with_source(type) if type
     events
   end
+
+  field :employees, [Types::EmployeesType] do
```
bencates, line 35:
```suggestion
  field :employees, [Types::EmployeeType] do
```

app/graphql/types/query_type.rb
```diff
@@ -31,4 +31,20 @@ def events(after: nil, type: nil)
...
+  end
+
+  def employees(type: nil)
+    puts type
+    puts Clients::BamboohrClient.find_employees.find_birthday_employees()
+    case type
+    when 'Anniversary' then Clients::BamboohrClient.find_employees.find_anniversary_employees()
+    when 'Birthday' then Clients::BamboohrClient.find_employees.find_birthday_employees()
+    when 'TimeOff' then Clients::BamboohrClient.find_employees.find_time_off_employees()
+    else Clients::BamboohrClient.find_employees()
```
bencates, line 47: This should probably be `when nil then ...`. If the user mistypes the `type` argument (e.g. `Timeoff` instead of `TimeOff`) it's better to produce an error rather than to show them incorrect data.

app/graphql/types/query_type.rb
```diff
@@ -31,4 +31,20 @@ def events(after: nil, type: nil)
     events = events.with_source(type) if type
     events
   end
+
+  field :employees, [Types::EmployeesType] do
+    description "Employee events from bambooHR"
+    argument :type, String, required: false
```
bencates, line 37: It's probably worth defining an [enum](https://graphql-ruby.org/type_definitions/enums) for this argument instead of accepting arbitrary strings.

app/graphql/types/query_type.rb
cbarber: I'd recommend revising the schema here. We have a bit of a naming collision with `events`. Maybe that could renamed to something more specific like `collaboration_events` and your new field could be `employee_events`. `employee_events` could have fields `anniversary`, `birthday`, `time_off`. This makes querying a little easier since the fields are discoverable and removes errors either returning a full list of employees e.g. if we had a typo `Aniversary` in the frontend it might appear to be working.

Have a look at the current `events` field and `EventCollectionType` to see how this can be done.

fallynlogan:
> Looks good! I just have a question about the plan for the events widget - are the goals the same as they were in the original PR with the display illustration? I can see we're still working to show birthdays, anniversaries and PTOs - is the "today in history" feature still a plan for later?

The "today in history" feature is not going to be used anymore. The final design that this information is going to be used for is like this:
image

app/graphql/types/query_type.rb
fallynlogan: This should be fixed now!

app/graphql/types/employees_type.rb
fallynlogan: This should be fixed now!

cbarber (approved):
LGTM. Nice work!

PR #284, opened 2021-07-15
fix : fresh page load does not show weather

sandipa12:
This bug is caused because onCompleted callbacks will be called many times . Removing it calls setNewWidgetId once and fixes the issue.

### DIFF
--- a/app/javascript/components/widget-controller.jsx
+++ b/app/javascript/components/widget-controller.jsx
@@ -55,14 +55,6 @@ export const WidgetController = ({ client, cityName }) => {
       query={getWidgets}
       variables={{ id: newWidgetId, cityName }}
       fetchPolicy={fetchPolicy}
-      onCompleted={response => {
-        const {
-          location: {
-            widgets: { byIdOrFirst: current },
-          },
-        } = response;
-        setNewWidgetId(current.id);
-      }}
     >
       {({ loading, error, data }) => {
         if (loading) {


app/javascript/components/widget-controller.jsx
```diff
@@ -36,7 +38,7 @@ const getWidgets = gql`
 export const WidgetController = ({ client, cityName }) => {
   const [newWidgetId, setNewWidgetId] = useState(0);
 
-  const fetchPolicy = newWidgetId === 0 ? 'network-only' : 'cache-first';
+  const fetchPolicy = newWidgetId === 0 ? 'cache-first' : 'network-only';
```
cbarber, line 41: I'm not following this change. `newWidgetId` will be 0 on first load, the ternary will evaluate to true, and `network-only` will be used.

sandipa12: I agree that but when I change it to cache first, the widgets did not show loading and work as expected. I ran into the fresh browser and checked it. Is there a suggestion on what I can look up to?

cbarber: This suggests that issue might be related to the apollo cache and requires further investigation to understand what the actual culprit is.

https://www.apollographql.com/docs/react/data/queries/#supported-fetch-policies

app/javascript/components/widget-controller.jsx
```diff
@@ -73,6 +75,20 @@ export const WidgetController = ({ client, cityName }) => {
...
+        if (
+          R.pathOr(
+            data.location.widgets.enabled === [],
+            ['NullPointerhandler', 'condition'],
+            {
+              NullPointerhandler: {
+                conditon: data.location.widgets.enabled === [],
+              },
+            },
+          )
```
tucker-weed, line 87: What's the advantage of using ramda in this conditional?

cbarber: In this case, ramda isn't much better than a lodash `get` since we're using it to just safely access a path in an object. The pay off with ramda is functional composition: https://github.com/ramda/ramda/wiki/Cookbook

cbarber (commented):
This fixes the issue, but is a good example of when to provide more details in the commit message beneath the commit summary. Something like:
```
fix: incognito first load should continue past loading

This bug is caused by..... Deleting `onCompleted` resolves it because...

cbarber (changes requested):
I'm looking for an explanation of the bug in the commit description. See previous comment for an example template.

cbarber:
@sandipa12 looking for a description of the issue in the commit

sandipa12:
@cbarber  I had updated the description of the issue before. Let me know if that adresses the issue.

cbarber (approved):
Thank you for the updated description

PR #285, opened 2021-07-19
feat: save employees to database

fallynlogan:
Save employees to database to retrieve employee events information faster.
Branched off of [this PR](https://github.com/mojotech/helios2/pull/283)

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -37,8 +37,7 @@ class Types::QueryType < Types::BaseObject
   end
 
   def employee_events
-    # dummy return value for now because cannot return null
-    # will be filled in with database calls in a later PR
+    Clients::BamboohrClient.find_employees.update_employees
     42
   end
 end
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -31,4 +31,14 @@ class Types::QueryType < Types::BaseObject
     events = events.with_source(type) if type
     events
   end
+
+  field :employee_events, Types::EmployeeEventCollectionType do
+    description "Employee events from bambooHR"
+  end
+
+  def employee_events
+    # dummy return value for now because cannot return null
+    # will be filled in with database calls in a later PR
+    42
+  end
 end
--- a/app/lib/clients/create_employees.rb
+++ b/app/lib/clients/create_employees.rb
@@ -5,6 +5,26 @@ class Clients::CreateEmployees
     @dates = create_dates
   end
 
+  def add_employee(employee)
+    EmployeeEvents.create!(external_id: employee['id'], display_name: employee['displayName'],
+                           birthday: employee['birthday'], hire_date: employee['hireDate'],
+                           is_photo_uploaded: employee['isPhotoUploaded'], photo_url: employee['photoUrl'],
+                           time_off_employees: find_time_off_employees)
+  end
+
+  def update_employees
+    @employees.each do |e|
+      if EmployeeEvents.exists?(external_id: e['id'])
+        EmployeeEvents.update(e['id'], display_name: e['displayName'],
+                                       birthday: e['birthday'], hire_date: e['hireDate'],
+                                       is_photo_uploaded: e['isPhotoUploaded'], photo_url: e['photoUrl'],
+                                       time_off_employees: find_time_off_employees)
+      else
+        add_employee(e)
+      end
+    end
+  end
+
   def create_dates
     (-2..2).map { |n| (Time.zone.today + n).to_formatted_s(:iso8601) }
   end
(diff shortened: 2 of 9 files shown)


schema.graphql
```diff
@@ -10,6 +10,19 @@ type Announcement {
...
+  displayName: String!
+  id: ID!
+  isPhotoUploaded: String!
+  photoUrl: String!
+}
+
+type EmployeeEventCollection {
+  anniversaryEmployees: [Employee!]!
+  birthdayEmployees: [Employee!]!
+  timeOffEmployees: [Employee!]!
```
sandipa12, line 23: I am curious about  [Employee!] ! , what does it refer to ?

fallynlogan: It is the data type that the array must consist of! For ex all of the elements in `anniversaryEmployees` must be of type `Employee`. The type `Employee` then has to have certain fields like `display_name, birthday, etc.`

app/graphql/types/query_type.rb
```diff
@@ -37,8 +37,7 @@ def events(after: nil, type: nil)
   end
 
   def employee_events
-    # dummy return value for now because cannot return null
-    # will be filled in with database calls in a later PR
+    Clients::BamboohrClient.find_employees.update_employees
```
cbarber, line 40: Should we be saving employees to the database when we query for employee events in graphql? It seems like it will still have a long request time so we probably want to save them asynchronously via sidekiq.

cbarber (commented):
I think we can restructure this a little bit to make requests to populate the database on a schedule and switch the graphql queries to request from the database.

app/lib/clients/create_employees.rb
```diff
@@ -5,6 +5,26 @@ def initialize(employees, time_off)
     @dates = create_dates
   end
 
+  def add_employee(employee)
+    EmployeeEvents.create!(external_id: employee['id'], display_name: employee['displayName'],
+                           birthday: employee['birthday'], hire_date: employee['hireDate'],
+                           is_photo_uploaded: employee['isPhotoUploaded'], photo_url: employee['photoUrl'],
+                           time_off_employees: find_time_off_employees)
```
bencates, line 12: This method is named `add_employee` but it creates an `EmployeeEvent`. That seems surprising to me.

app/lib/clients/create_employees.rb
```diff
@@ -5,6 +5,26 @@ def initialize(employees, time_off)
...
+  end
+
+  def update_employees
+    @employees.each do |e|
+      if EmployeeEvents.exists?(external_id: e['id'])
+        EmployeeEvents.update(e['id'], display_name: e['displayName'],
+                                       birthday: e['birthday'], hire_date: e['hireDate'],
+                                       is_photo_uploaded: e['isPhotoUploaded'], photo_url: e['photoUrl'],
+                                       time_off_employees: find_time_off_employees)
```
bencates, line 21: There's a bug in this. We're checking for the presence of an `EmployeeEvents` by it's `external_id`, but then trying to update it using its `id`. Those are probably not the same. In any case, I think it would be cleaner to retrieve the record and use it's `update` method instead of using `EmployeeEvents.update`:

```suggestion
      employee_event = EmployeeEvents.find_by_external_id(e['id])
      if employee_event
        employee_event.update(display_name: e['displayName'],
                                       birthday: e['birthday'], hire_date: e['hireDate'],
                                       is_photo_uploaded: e['isPhotoUploaded'], photo_url: e['photoUrl'],
                                       time_off_employees: find_time_off_employees)
```

app/models/employee_events.rb
```diff
@@ -0,0 +1,5 @@
+class EmployeeEvents < ApplicationRecord
```
bencates, line 1:
```suggestion
class EmployeeEvent < ApplicationRecord
```

Model names in rails should be singular. The filename should be `employee_event.rb` as well.

app/models/employee_events.rb
```diff
@@ -0,0 +1,5 @@
...
+  serialize :birthday_employees, Array
+  serialize :anniversary_employees, Array
```
bencates, line 3: These don't seem like they line up with the database schema. We have a `time_off_employees` column, but no `birthday_employees` or `anniversary_employees` columns.

Where do you expect these to be written to and read from?

db/migrate/20210721200625_create_employee_events.rb
```diff
@@ -0,0 +1,15 @@
+class CreateEmployeeEvents < ActiveRecord::Migration[5.2]
+  def change
+    create_table :employee_events do |t|
```
bencates, line 3: This table is named `employee_events`, but most of the columns (with the exception of `time_off_employees`) appear to be describing an employee, not an event.

bencates (changes requested):
I don't really understand what the distinction between an `Employee` and an `EmployeeEvent` is. In some places it seems like an `EmployeeEvent` is a collection of employees while in others it seems like we're using the two terms interchangeably.

Could you add some documentation describing what each of those two terms means in the context of the data model you're building?

cbarber:

PR #286, opened 2021-07-27
fix: failing circleCI tests for bamboohr_spec

fallynlogan:
@bbazyk caught a bug where in the case that there were more than one employees with a birthday that the test for birthdays in `bamboohr_spec.rb` would fail in circleCI. These fixes should resolve this!

### DIFF
--- a/spec/graphql/bamboohr_spec.rb
+++ b/spec/graphql/bamboohr_spec.rb
@@ -2,21 +2,23 @@ require 'rails_helper'
 require 'bamboozled'
 
 EMPLOYEESDATA = [
-  { "id" => "4", "displayName" => "Charlotte Abbott", "hireDate" => Time.zone.today.to_formatted_s(:iso8601),
+  { "id" => "4", "displayName" => "Charlotte Abbott", "hireDate" => "2019-07-22",
     "birthday" => "07-29", "isPhotoUploaded" => "true",
     :photoUrl => "http://blahbambooapitestblah.bamboohr.com/employees/photos/?h=e27ddc876e2b95ca0ac80ad62967506a" },
   { "id" => "5", "displayName" => "Ashley Adams", "hireDate" => "2020-12-18",
     "birthday" => "05-25", "isPhotoUploaded" => "true",
     :photoUrl => "http://blahbambooapitestblah.bamboohr.com/employees/photos/?h=efbf026fde3168eb96a3447b040f5f91" },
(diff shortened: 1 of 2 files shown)


cbarber (commented):
Could we pass the days of the week we're querying to `find_birthday_employees`, `find_time_off_employees`, `find_anniversary_employees` instead of generating new dates for each test?

bbazyk:
Why is it that this test didn't fail before? Also, if actual mojo employees' birth and hire dates are still stored as in that original date format instead of 'TODAY_SHORT or LONG' - then shouldn't the test remain specific to those types? The error I received from 'yarn run test' indicated that a retrieval for employee id #6's data returned employee #4's instead - why would testing employees with different birthdays result in that failure?

fallynlogan:
> Why is it that this test didn't fail before? Also, if actual mojo employees' birth and hire dates are still stored as in that original date format instead of 'TODAY_SHORT or LONG' - then shouldn't the test remain specific to those types? The error I received from 'yarn run test' indicated that a retrieval for employee id #6's data returned employee #4's instead - why would testing employees with different birthdays result in that failure?

@bbazyk the test failed because when I wrote the test there was only one employee returned in the `birthday_result` array in `bamboohr_spec.rb` so its test would pass. It started failing because there is another birthday in the test data that occurs on 7-29 so for the following dates (7/27, 7/28, 7/29, 7/30, and 7/31) there would be two employees returned in the array for the test (ie the employee whose birthday is set to always the current day and the one whose birthday is on 7/29) because the employee whose birthday is set to 7/29 is first in the test data it was being added to the `birthday_result` array before the employee that the test was written for so the test started failing. The current fixes in this PR should resolve this. Does that make sense? 
For Mojo Birthdays the API and bambooHR in general store birthdays as just a month and day (excluding the year) so Mojo birthdays would also be formatted this way!

spec/graphql/bamboohr_spec.rb
```diff
@@ -1,21 +1,25 @@
 require 'rails_helper'
 require 'bamboozled'
 
+TODAY_LONG = Time.zone.today.to_formatted_s(:iso8601)
+TODAY_SHORT = Time.zone.now.strftime("%m-%d")
+DATES = [TODAY_LONG].freeze
```
cbarber, line 6: Lets not shift the time period for both birthdays and today. This sets us up to have flaky tests. We can just set this to be the date you first created the test when it was working and leave the json data as static.
`DATES = ["2021-07-22"].freeze`

cbarber (approved):
Nice job!

PR #287, opened 2021-07-27
[bug] fixed time update & date display bug with HOC

tmsavage:


app/javascript/components/top-corner-controller.jsx
```diff
@@ -26,7 +35,13 @@ const Row = styled.div`
...
 `;
 
-const TopCorner = ({ showWeather, cityName, loading, error, location }) => {
+export const TopCorner = ({
+  showWeather,
+  cityName,
+  loading,
+  error,
+  location,
+}) => {
```
fallynlogan, line 44: Should this be added in its own refactor commit?

app/javascript/components/hocs/with-fragment.jsx
```diff
@@ -16,6 +23,10 @@ export default function withFragment(WrappedComponent, fragmentObject) {
       };
     }, {});
 
+    if (componentName != null) {
+      console.log('After Data Props: ', fragmentDataProps);
```
fallynlogan, line 27: What does 'After Data Props' mean?

fallynlogan (commented):
LGTM just a few questions!

app/javascript/components/top-corner-controller.jsx
tmsavage: I don't think so. This is a very minor "refactor" and is probably not worth a separate commit. But maybe a mentor has another opinion.

app/javascript/components/hocs/with-fragment.jsx
tmsavage: This statement was just meant to check the differences of props before and after `fragmentDataProps` was computed.

app/javascript/components/side-panel.jsx
```diff
@@ -84,4 +86,4 @@ SidePanel.defaultProps = {
   error: false,
 };
 
-export default withFragment(SidePanel, { location: getSidePanel });
+export default withFragment(SidePanel, { location: getSidePanel }, 'foo');
```
bencates, line 89: I think this change was included by mistake.

bencates (approved):
Pull out that extra `'foo'` and then you're good to go!

PR #288, opened 2021-07-28
feat: [WIP] display anticipated weatherinfo

sandipa12:
This ticket is to display anticipated precipitation info in text if there is a higher probability of rainfall in next hour 
Here's how the precipMessage is displayed when precipitation probability is more than 50%. Working on investigating if the data is accurate for the  next hour .

Screen Shot 2021-07-28 at 1 21 24 PM

PT: https://www.pivotaltracker.com/story/show/178576211

### DIFF
--- a/app/graphql/types/weather_type.rb
+++ b/app/graphql/types/weather_type.rb
@@ -70,6 +70,7 @@ class Types::WeatherHourlyDataType < Types::WeatherObject
   field :clouds, Int
   field :uvi, Int
   field :visibility, Int
+  field :snow, Float, null: true
 end
 
 class Types::WeatherMinutelyDataType < Types::WeatherObject
@@ -104,6 +105,7 @@ class Types::WeatherType < Types::WeatherObject
   field "hourly", [Types::WeatherHourlyDataType]
   field "daily", [Types::WeatherDailyDataType]
   field "offset", Int
+  field "precip_message", String
 
   field "moonPhase", Float, deprecation_reason: "Moved to Location"
 
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -11,13 +11,49 @@ class Location < ApplicationRecord
   def self.primary(city_name = ENV['PRIMARY_CITY_NAME'])
     find_by(city_name: city_name)
   end
-
-  def weather
+   
+  def weather_detail(weather_info, time_split)
     response = ::Clients::WeatherClient.forecast(self).dup
-    response['moonPhase'] = moon_phase
-    response.freeze
-    response
-  end
+    weather_info.select {
+      |key, value |
+        case value.to_f
+      when value.to_f >= 0.25 then
+      response['precip_message'] = "possible to #{key} #{time_split}"
+      when value.to_f >= 0.50 then
+      response['precip_message'] = "Likely to #{key} #{time_split}"
+      when value.to_f >= 0.75 then
+      response['precip_message'] = "#{key} #{time_split}"
+      else
+        response['precip_message'] = ""
+      end
+    }
+    end
+    
+    def weather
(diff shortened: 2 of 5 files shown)


app/graphql/types/weather_type.rb
```diff
@@ -33,7 +33,7 @@ def initialize(value, context, **kwargs, &block)
 
   field "id", Integer
   field "main", String
-  field "description", String, resolve: ->(obj, _args, _ctx) { obj["description"].to_s.humanize }
+  field "description", String, resolve: ->(obj, _args, _ctx) { obj["description"].to_s.humanize}
```
bbazyk, line 36: Just curious what this line does (I guess I'm not very familiar with the syntax)

bencates: https://apidock.com/rails/v5.2.3/ActiveSupport/Inflector/humanize

app/models/location.rb
```diff
@@ -15,6 +15,69 @@ def self.primary(city_name = ENV['PRIMARY_CITY_NAME'])
   def weather
     response = ::Clients::WeatherClient.forecast(self).dup
     response['moonPhase'] = moon_phase
+    
+   response['current']['weather'][0]['hourly'] = response['hourly']
+
+   weather_info = {"rain" => "response['hourly'][2]['pop']" , "snow" => "response['hourly'][2]['snow']"}
```
bencates, line 21: I don't think we want the quotes around the values. We want `"rain"` to be the value in the response, not the literal string `"response['hourly'][2]['pop']"`.

app/models/location.rb
```diff
@@ -15,6 +15,69 @@ def self.primary(city_name = ENV['PRIMARY_CITY_NAME'])
...
+           when value.to_f >= 0.50 then
+            response['precip_message'] = "Likely to #{key} tonight"
+           when value.to_f >= 0.75 then
+           response['precip_message'] = "#{key} tonight"
+           else    
+           response['precip_message'] = "possible to #{key} tonight"
+         end
+       }  
+       end
+     end
```
bencates, line 77: It looks like the only difference between your various `case` statements is the time frame. It might be worth pulling out a helper method that takes `value.to_f` and a `time_frame` string to reduce some of the duplication.

Thatl'll also help to make the flow through this complicated conditional easier to understand. Right now there's at least one bug - it doesn't handle the case where it's not going to rain or snow in the next 24 hours - but it's hard to tell that as it's currently written.

bencates (commented):
This is looking better.

There's still a lot going on in that `weather` method. The conditional logic here is pretty nuanced and since it's getting results from an API it'll be hard to verify that all the edge cases work correctly.

This function and whatever helpers we end up pulling out of it would really benefit from some unit test coverage. We already have a test file for this class set up in `spec/models/location_spec.rb`, I think it would be worth our while to add some tests to cover the code you're working on. You can run that test on your machine with the following command: `bin/rspec spec/models/location_spec.rb`.

There's a ton of resources on the internet for learning how to use rspec. [This one](https://www.tutorialspoint.com/rspec/index.htm) looks like as good a place as any to start, though. Don't be afraid to play around with it; that's what tests are for.

app/models/location.rb
```diff
@@ -11,13 +11,49 @@ class Location < ApplicationRecord
...
+             time_split = "afternoon"
+           when 18..21 then
+             time_split = "evening"
+           end
+         end
+       end
+      weather_detail(weather_info,time_split)
+      response.freeze
+      response
+      end
```
tucker-weed, line 56: nitpick: there are some indentation issues on this function

Also, the spreads `2..24`, `12..18`, and `18..21` seem to be magic numbers. Perhaps comments could be added to better explain what is going on here?

cbarber:

PR #289, opened 2021-07-29
fix: added unique key prop for each displayed cam URL

bbazyk:
Fix console warning: "Warning: Each child in a list should have a unique "key" prop." for traffic widget panel


cbarber (approved):
LGTM

PR #290, opened 2021-07-30
feat: add cypress (and pvd and bld smoke tests)

tucker-weed:
Description:

1. Adds cypress dev dependency
2. Adds smoke tests for pvd and bld pages


cbarber (commented):
Nit: I think the first four commits can be a single `add cypress`. I can see the case being made for separating "add <framework>" and "add stuff that runs in <framework>"

I'd like to see a README update in the last commit mentioning the cypress tests depend on an up to date `rake db:seed` and a server running locally.

tucker-weed:
> Nit: I think the first four commits can be a single `add cypress`. I can see the case being made for separating "add " and "add stuff that runs in "
> 
> I'd like to see a README update in the last commit mentioning the cypress tests depend on an up to date `rake db:seed` and a server running locally.

@cbarber working on these changes now

.gitignore
```diff
@@ -44,3 +44,7 @@ yarn-debug.log*
 .DS_Store
 
 *.rdb
+
+# Ignore cypress assets
+/cypress/videos/*
+/cypress/screenshots/*
```
bencates, line 50: nitpick: we don't need the wildcard (`*`) for these. We can ignore the directory itself.

cypress/integration/smoke-test-bld.spec.js
```diff
@@ -0,0 +1,27 @@
+describe('main page, boulder smoke test', () => {
+  beforeEach(() => {
+    cy.fixViewport();
+    cy.visit('http://localhost:5000/Boulder');
+    cy.waitForInitialContent();
+  });
```
bencates, line 6: This and `smoke-test-pvd.spec.js` are almost identical. The only difference is the URL and the first test assertion.

It might be cleaner to abstract the location out and use a single test file:

```suggestion
describe('main page, smoke test', () => {
  ['Boulder', 'Providence'].forEach(location => {
    beforeEach(() => {
      cy.fixViewport();
      cy.visit(`http://localhost:5000/${location}`);
      cy.waitForInitialContent();
    });

    // etc...
```

tucker-weed: Was wondering how to consolidate the tests - this makes a lot of sense.

bencates: Actually, I think you'll have to have the `describe` inside the `forEach` for the `beforeEach` to work correctly. My apologies.

cbarber (approved):
Nice work :+1:

PR #291, opened 2021-08-02
Fix css issues on helios (separate active admin css from app/assets)

bbazyk:
Active admin css styling was overriding styling of side panel for helios. This PR's simple fix is to separate the css into vendor/assets/stylesheets (automatically still referenced by application.css in the original app/assets/stylesheets directory).

Ticket ID: #179089416

Screenshot from 2021-08-02 16-41-24
Screenshot from 2021-08-02 16-43-21

(active admin interface was tested, and the styling remains the same)
The deletion of the css file is git tracked in the commit, but its addition back into the vendor directory isn't.


cbarber (changes requested):
This is what I'm seeing running your branch locally:
Screenshot from 2021-08-03 08-55-35

bbazyk:
I realized that my original explanation for the absence of the css in this commit (it not being git-tracked) was correct, but missed the point. I since edited the .gitignore file to ensure that the css file, and its contained subdirectories aren't missed by github. You should be able to pull this and see the same difference as my included screenshots @cbarber (it helps to have the images in two tabs, and flip between them to notice that every piece of text and icon was shifted up and down after the active admin PR)

.gitignore
```diff
@@ -36,8 +36,15 @@ yarn-debug.log*
...
 
-# Ignore docker related files
-/vendor
+# Ignore docker related files, but don't ignore active admin's css specifications
+vendor/*
+!vendor/assets
+vendor/assets/*
+!vendor/assets/stylesheets
+vendor/assets/stylesheets/*
+!vendor/assets/stylesheets/active_admin.scss
```
jennyckaplan, line 45: This might be a cleaner solution: https://stackoverflow.com/a/18702826. Can we try something like this out?

cbarber (commented):
I like Jen's proposed alternative

.gitignore
```diff
@@ -36,8 +36,13 @@ yarn-debug.log*
...
 .env*
 
-# Ignore docker related files
-/vendor
+# Ignore docker related files, but don't ignore active admin's css specifications
+vendor/**
+!vendor/assets/
+!vendor/assets/stylesheets/
+!vendor/assets/stylesheets/**
+!vendor/assets/stylesheets/active_admin.scss
```
bencates, line 44: 1. This is redundant. `stylesheets/**` matches everything in the folder, you don't need to explicitly mention `active_admin.scss` as well.

   Based on the comment, I think we probably would only need `!vendor/assets/stylesheets/active_admin.scss` to explicitly include that one file.

2. I don't think `.gitignore` is the right place to handle this edge case in the first place. `git add` has a [`--force` flag](vendor/assets/stylesheets/active_admin.scss) to explicitly add an ignored file. Once a file has been committed, git will continue to track it even if `.gitignore` says otherwise.

cbarber (approved):
LGTM

PR #293, opened 2021-08-10
Bb/improve database seeding

bbazyk:
Ticket id: 178579007

rake db:seed
-should be able to be run multiple times.
-add new, randomish, events whenever the seed script is run

Replaced active record method 'find_or_initialize_by' by defining 'update_or_new' in the seeds file. This new method has the added benefit of always saving an object's fields - thereby allowing for inserts and field edits to be recognized during reseeding.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -148,6 +148,23 @@ update_or_new!(TrafficCam, title: "28th and Iris") do |r|
   r.feed_format = "video"
 end
 
+# Add random one-time events (will not be displayed on helios after this week)
+
+update_or_new!(Event, external_id: "Tucker's final helios PR :(") do |r|
+  r.source = 'github_pull'
+  r.created_at = Date.parse('2021-10-6T10:09:08.07')
+end
+
+update_or_new!(Event, external_id: "PR: fix to allow multiple seedings") do |r|
+  r.source = 'github_pull'
+  r.created_at = Date.parse('2021-10-6T10:09:08.07')
+end
+
+update_or_new!(Event, external_id: "#the-five-interns-2021 channel message") do |r|
+  r.source = 'slack_message'
+  r.created_at = Date.parse('2021-10-6T10:09:08.07')
+end
+
 if AdminUser.count.zero?
   AdminUser.create!(
     email: ENV['ADMIN_EMAIL'],
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,6 +1,11 @@
-providence = Location.find_or_initialize_by(
-  city_name: 'Providence'
-) do |r|
+def update_or_new!(klass, attributes)
+  my_record = klass.find_by(attributes) || klass.new(attributes)
+  yield(my_record)
+  my_record.save!
+  my_record
+end
+
+providence = update_or_new!(Location, city_name: 'Providence') do |r|
   r.latitude = 41.823989
   r.longitude = -71.412834
   r.time_zone = 'America/New_York'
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -1,187 +1,170 @@
...
+
+update_or_new(Event, external_id: "PR: fix to allow multiple seedings") do |r|
+  r.source = 'github_pull'
+  r.created_at = Date.parse('2021-10-6T10:09:08.07')
+end
+
+update_or_new(Event, external_id: "#the-five-interns-2021 channel message") do |r|
+  r.source = 'slack_message'
+  r.created_at = Date.parse('2021-10-6T10:09:08.07')
+end
```
bbazyk, line 167: Screenshot from 2021-08-10 13-47-23

Example events appear on helios for the week of August 10th - as expected.

db/seeds.rb
```diff
@@ -1,187 +1,169 @@
-providence = Location.find_or_initialize_by(
-  city_name: 'Providence'
-) do |r|
+def update_or_new(klass, attributes)
+  my_record = klass.find_by(attributes) || klass.new(attributes)
```
tucker-weed, line 2: nit: perhaps klass could be renamed to `record_class`? Looks good to me otherwise

bbazyk: 'class' is a reserved keyword for defining a class in ruby. Because it cannot be used as a variable name, the ruby community has taken to using 'klass'. Not for any special reason, but I believe it's almost a kind of naming convention at this point.

planetmcd: @bbazyk @tucker-weed Ben is correct.  Because of reserved word status, this is the idiomatic way of referencing a class in code. I like your instincts that this is harder to read than using a real word and something more descriptive though.  If it wasn't fairly entrenched (you can find lots of examples if you dig around the rails codebase) it would be a solid suggestion.

db/seeds.rb
```diff
@@ -148,6 +148,23 @@ def update_or_new(klass, attributes)
   r.feed_format = "video"
 end
 
+# Add random events
+
+update_or_new(Event, external_id: "Tucker's final helios PR :(") do |r|
+  r.source = 'github_pull'
+  r.created_at = Date.parse('2021-10-6T10:09:08.07')
```
planetmcd, line 155: These dates are hard coded, what are the implications of that?

db/seeds.rb
```diff
@@ -1,187 +1,152 @@
-providence = Location.find_or_initialize_by(
-  city_name: 'Providence'
-) do |r|
+def update_or_new(klass, attributes)
+  my_record = klass.find_by(attributes) || klass.new(attributes)
+  yield(my_record)
+  my_record.save!
```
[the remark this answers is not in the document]
bbazyk: save and save! both run validations, but save returns the result as a boolean, and save! raises a 'record invalid' error. Then, by 'adopt the naming conventions of save' you suggest I should rename 'update_or_new' to 'update_or_new!' ? I agree that name explains the method's full purpose better, and I'll make that change.

db/seeds.rb
bbazyk: The 'MojoTech by the Numbers' widget on helios displays events only for the week (I believe). I reached out to Will before he left for PTO to ask what he had in-mind for this half of the PR... These sample events are to be added as a display test for helios, and only appear on the widget for a temporary amount of time.

db/seeds.rb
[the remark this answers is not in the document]
planetmcd: Basically bang methods, `method_name!` will alter the state of the world in some non reversible way.  It is an idiom in Ruby.  In the case here, if a database transaction is created (I can explain that tomorrow if you haven't looked it up by then), a `save!` will close that transaction and try to commit changes.  So omitting the bang on this method might lead readers to conclude the change is non destructive.  Perhaps the non bang method `save` might be better in this case.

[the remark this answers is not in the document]
bbazyk: I've renamed update_or_new to reflect the bang save method it contains.

db/seeds.rb
bbazyk: These events serve more as tests for the Numbers widget, and their hardcoded dates may be updated by future helios editors to display test events on that day. These test records are not intended to generate current events or artificially increment the events displayed on helios for any given week.
Because they can be edited to test the functionality of the Numbers widget, these test events are created with the update_or_new! method - allowing for successful seeding of their edited fields.

cbarber:

PR #295, opened 2021-08-12
Bb/weathercorner css

bbazyk:
Fixed styling issue where weather corner's div was unintentionally taller than the time + AM/PM div.. This caused all widget's side panel elements (besides the weather widget who doesn't render weathercorner) to be shifted downwards by that height offset.

ticket id: 179167655
Description:
"Weather corner's styling doesn't apply to the weather widget (which neglects displaying weather in the top corner). This results in the rest of the weather widget's sidepanel styling to be offset."

### DIFF
--- a/app/javascript/components/time.jsx
+++ b/app/javascript/components/time.jsx
@@ -19,6 +19,7 @@ const AMPM = styled.div`
   font-size: ${fontSizes.medium};
   font-family: ${fonts.extended};
   margin-left: 11px;
+  line-height: 45px;
 `;
 
 const TimeValue = styled(Row)`
--- a/app/javascript/components/time.jsx
+++ b/app/javascript/components/time.jsx
@@ -14,7 +14,6 @@ const HourMin = styled.div`
   font-family: ${fonts.extended};
   font-weight: ${weights.regular};
   position: relative;
-  top: 8px;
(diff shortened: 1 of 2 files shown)


app/javascript/components/weather-corner.jsx
```diff
@@ -40,8 +40,7 @@ const TempText = styled.div`
 `;
 
 const SkyIconWrapper = styled.div`
-  margin-top: 35px;
-  margin-right: 24px;
+  margin-bottom: 5px;
 `;
```
bbazyk, line 44: Before this commit, the sky icon's top margin could result in its div's height growing. This just re-creates the original problem of the PR where the skyicon/temperature div was taller than the time/AMPM div.

app/javascript/components/time.jsx
```diff
@@ -13,7 +13,6 @@ const HourMin = styled.div`
   font-size: ${fontSizes.xlarge};
   font-family: ${fonts.extended};
   font-weight: ${weights.regular};
-  position: relative;
 `;
```
bbazyk, line 16: Using box model (with BottomLineRow) over positioning method

cbarber (approved):
LGTM :+1:

PR #306, opened 2021-12-20
Add a users table for developers

ctvidic:
Ticket #180678039

- Adds table migration/model for developer users accommodating slack handle, github handle, and public key

### DIFF
--- a/app/models/developer_user.rb
+++ b/app/models/developer_user.rb
@@ -1,0 +1,5 @@
+class DeveloperUser < ApplicationRecord
+  validates :public_key, presence: true
+  validates :git_handle, presence: true
+  validates :slack_handle, presence: true
+end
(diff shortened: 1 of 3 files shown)


app/models/developer_user.rb
```diff
@@ -0,0 +1,5 @@
+class DeveloperUser < ApplicationRecord
+    validates :public_key, presence: true
```
aaddaamm, line 2: CI is whining about using consistent spacing for this file, just back this spacing off to 2 spaces and it should be good.

mariana0pachon: ☝️ also, might be good to change the default on your editor

rlamsal1256 (approved):
Nice work 👍

cbarber (approved):
LGTM :rocket:

PR #307, opened 2022-01-04
Added Events endpoint with create method

lsim6580:


aaddaamm:
pipeline is mad about some linting issues, fixup those changes to make it happy and rebase off latest master while we are at it if you don't mind. 🙏🏻

app/controllers/events/event_controller.rb
```diff
@@ -0,0 +1,18 @@
+class Events::EventController < ActionController::Base
+  protect_from_forgery with: :null_session
+
+  def create
+    @event = Event.new(event_params)
```
aaddaamm, line 5: does this need to be an instance variable? or can we get away with local?

app/controllers/events/event_controller.rb
```diff
@@ -0,0 +1,20 @@
+class Events::EventController < ActionController::Base
+  protect_from_forgery with: :null_session
+
+  def create
+    event = Event.new(event_params)
+    if event.save
+      render json: { message: 'Event created' }, status: :created
+    else
+      render json: { message: event.errors.full_messages }, status: :bad_request
```
aaddaamm, line 9: i don't know enough about the event model but should we guard against `nil` in this accessor? not sure if we are guaranteed to get `errors` back on the event. Correct me if I'm wrong though.

lsim6580: I am confused what you are asking? check for nil on event.errors? to my knowledge if event.save returns false, event.errors should be initialized

aaddaamm: on second thought, can we avoid this conditional and throw if the record fails to save? we have similar looking code in the rescue clause below. I was reading up on Model.save and it looks like `save` will return a boolean and `save!` will throw an exception if it fails. Maybe we could switch to using the bang method and rely on the rescue to catch any errors?

lsim6580: Why not make use of the more readable errors that .save would allow? Exception messages can be messy imo.

app/controllers/events/event_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+    end
+  rescue StandardError => exception
```
lsim6580, line 11: @aaddaamm this will catch any of the weird save issues that we need

cbarber (approved):
LGTM. Nice work!

PR #310, opened 2022-02-02
fix: replace Dean St cam

malssid:
ID: #181116708

Replaced (broken) Dean street camera with Broadway which is in the same area.


glassesaaron (approved):
Looks good to me!

cpjolicoeur:
:( poor Dean St.  The camera is working again at Dean btw.

malssid:
Oh good news. I'll add it back 🥳

PR #311, opened 2022-02-02
fix: Remove extra 'images from' URL from cam page

malssid:
ID: #181116668

The Providence traffic cam page was showing 2 "images from" URLs for http and https. It will only show the https URL from now on.

### DIFF
--- a/app/javascript/components/widgets/traffic/panel.jsx
+++ b/app/javascript/components/widgets/traffic/panel.jsx
@@ -99,7 +99,7 @@ const Traffic = ({ cityName, startTimer }) => (
       const [firstColumn, secondColumn] = splitAt(columnSize, trafficCams);
       const getUniqUrls = compose(
         uniq,
-        map(cam => new URL(cam.url).origin),
+        map(cam => new URL(cam.url).hostname),
       );
 
       return (


app/javascript/components/widgets/traffic/panel.jsx
```diff
@@ -117,9 +117,9 @@ const Traffic = ({ cityName, startTimer }) => (
             </Column>
           </Row>
           <Notice>Images from:</Notice>
-          {getUniqUrls(trafficCams).map(url => (
-            <Notice key={url}>{url}</Notice>
-          ))}
+          {getUniqUrls(trafficCams).map(url => {
+            return url.includes('https') && <Notice key={url}>{url}</Notice>;
```
cbarber, line 121: I should have been clearer in the ticket. Maybe we can get rid of the schema portion and just show the domain?

chewnoill: maybe try 
```js
new URL(url).hostname
```
https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

malssid: ok

cbarber:
@malssid this looks good. Just needs a rebase and squash/fixup the two commits into one.

aaddaamm (approved):

PR #312, opened 2022-02-04
feat: blocks fall again after cycle

malssid:
ID: #181114917

Each serialized block gets a random y position. This causes them to fall from the top each time the slack/github events page mounts.

### DIFF
--- a/app/javascript/components/widgets/numbers/falling-blocks.jsx
+++ b/app/javascript/components/widgets/numbers/falling-blocks.jsx
@@ -131,6 +131,14 @@ const getImageFromCache = imagePath => {
   return Promise.resolve(image);
 };
 
+export const getBlocksDroppedSinceLastMount = () => {
+  return localStorage.getItem('blocksDroppedSinceLastMount') === 'true';
+};
+
+export const setBlocksAddedSinceLastMount = blocksDropped => {
+  localStorage.setItem('blocksDroppedSinceLastMount', blocksDropped);
+};
+
 class Scene extends React.Component {
   imageCache = {};
 
@@ -164,6 +172,7 @@ class Scene extends React.Component {
     if (isPresent(world)) {
       this.restoreWorld(world);
     }
+
     Composite.allBodies(this.engine.world)
       .filter(body => !body.isStatic)
       .forEach(block => this.sleepStartEvent(block));
@@ -176,6 +185,7 @@ class Scene extends React.Component {
       world,
       data: { loading },
     } = this.props;
+
     if (prevProps.data.loading && !loading && isPresent(world)) {
       this.restoreWorld(world);
     }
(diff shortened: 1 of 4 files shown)


aaddaamm:
Is it possible to get a GIF or like a video link of it working?

malssid:
> Is it possible to get a GIF or like a video link of it working?
Here you go.
https://user-images.githubusercontent.com/67285754/152576709-9de0f616-f58d-4cbd-94c7-47cc0949c612.mov

chewnoill:
Does it still make sense to save the block state at all? Would be nice to drop all the local cache stuff if we don't need it anymore

malssid:
> Does it still make sense to save the block state at all? Would be nice to drop all the local cache stuff if we don't need it anymore

I actually misunderstood Craig's story and implemented it incorrectly. 

To check if no events have occurred in one entire cycle (mounting and then unmounting) we probably don't need to keep track of the previous block state (locally) anymore. We talked about it today and think setting a boolean in local storage to keep track of whether an event has occured "since last unmount" makes the most sense.

malssid:
Here is a video demonstrating what happens when blocks are idle for an entire cycle. https://streamable.com/bd3mg5

app/javascript/components/widgets/numbers/falling-blocks.jsx
```diff
@@ -344,6 +366,7 @@ class Scene extends React.Component {
     // randomly pick a point from one wall edge to the other
     // so that they don't fall directly on top of each other.
     if (block) {
+      setBlocksAddedSinceLastMount(true);
```
chewnoill, line 369: oh I see, we are setting it from in here also. 

It might be easier to clear the world state, on unmount, if we haven't added anything new since it was mounted; clearing existing world cache when we don't need it, forces it to rebuild the next time it gets mounted. Could help avoid leaking more state into local storage

cbarber: Has anything been added was surprisingly tricky:
* Dropped balls on mount (server counts > local cache counts)
* Dropped balls while mounted (subscription event from server)
* New balls to drop while unmounted (server counts > local counts again)

chewnoill: Ah, I see, that does seem tricky. Is it worth persisting it with the rest of the `world` state then?

cbarber: Maybe. I can't remember the reasoning we arrived at to prefer localStorage over local graphql storage. Do you remember @malssid

malssid: I don't completely remember why. I think I chose it just because it's something I've used in the past.It might make sense to just store it with the world state.

cbarber (approved):
LGTM :shipit:

PR #318, opened 2022-02-17
Upgrade gems first pass

cbarber:
Bumping ruby gems now that the client/server are split

This is the first pass before digging into individually pinned gems.


malssid (approved):
LGTM

PR #321, opened 2022-02-17
feat: add/configure dotenv for elixir

malssid:
You can access environment variables with `Dotenv.get("<ENV_VARIABLE>")`


server-phoenix/lib/helios/application.ex
```diff
@@ -7,6 +7,11 @@ defmodule Helios.Application do
 
   @impl true
   def start(_type, _args) do
+    unless Mix.env == :prod do
+      Dotenv.load
```
cbarber, line 11: Does this find the `.env.local` file in the parent directory?

server-phoenix/mix.exs
```diff
@@ -50,7 +50,8 @@ defmodule Helios.MixProject do
       {:jason, "~> 1.2"},
       {:plug_cowboy, "~> 2.5"},
       {:absinthe, "~> 1.7"},
-      {:absinthe_plug, "~> 1.5"}
+      {:absinthe_plug, "~> 1.5"},
+      {:dotenv, "~> 3.0.0", only: [:dev, :test]}
```
cbarber, line 54: I'm unclear how this works in production if we have something like `Dotenv.get("<ENV_VARIABLE>")` in code.

malssid: I can add the library for all environments by leaving out the "only" option. Are these env variables loaded into heroku directly for production? (through dashboard settings or the cli)

cbarber: They are. My question is how will `Dotenv.get` function in the code in production if its only available for `:dev` and `:test`?

server-phoenix/lib/helios/application.ex
cbarber: I'm getting this locally:
```
iex(1)> Dotenv.get("WEATHER_API_KEY")
nil
```

malssid: That library might only work for .env files, not .env.local

server-phoenix/mix.exs
malssid: I can remove the "only" option like this:
`{:dotenv, "~> 3.0.0"}`
so then it'll be available in all environments.

chewnoill:
LGTM, I think this is the right thing to do. There is some language/documentation problems though. Not clear if these have been addressed based on PR description/discussion

* ENV should reference `System.get_env("<ENV_VARIABLE>")` and it should be the same way for dev/test/prod. It doesn't make sense to reference them through `Dotenv.get` because this is not loaded in prod.

* If we need to do something with `.env.local` (`cp .env.local .env`) before we can run we should update the readme so that is clear

malssid:
Are you suggesting we use `System.put_env(key, value)` to load them in before reading them with `System.get_env(key)`?

chewnoill:
Yes, I expect that was the whole point of using [Dotenv.load](https://github.com/avdi/dotenv_elixir/blob/master/lib/dotenv.ex#L112)

malssid:
ok yeah, I just tested `System.get_env("WEATHER_API_KEY")` and it works. I'll also update the readme so people know to copy the sample .env.local to a .env.

malssid:
Never mind. It looks like the readme already tells them to copy .env.local.sample to a .env.

cbarber:
* Lets rename `.env.local.sample` to `.env.sample` as well
* webpack.config.js has a hardcoded path to `.env.local` that will need to be updated

README.md
```diff
@@ -34,7 +34,8 @@ Prequisites:
 - `gem install foreman`
 - `bundle install`
 - `yarn install`
-- `cp .env.local.sample .env` -- fill in env vars where applicable
+- `cp .env.sample .env` -- dotenv loads variables from a **.env** in your project's root directory. Fill in env vars where applicable
+- `cp .env.local .env` -- for people with a .env.local
```
chewnoill, line 38: This seems confusing. Do we need both of these? Can we consolidate them into one file?

malssid: Yeah, my bad. Let's just keep the first line that tells them to copy the contents from .env.sample to a .env

cbarber (approved):
Nice work! :ship: :rocket:

PR #327, opened 2022-02-22
chore: bump to rails 6.1

cbarber:
Breaking fix:
* tod needed #to_s for where query


aaddaamm:
to-the-future-time-travel

PR #328, opened 2022-02-23
docs: update README to include nvm instructions

linalim:


cbarber:
Could you add a commit including `.nvmrc`?

PR #329, opened 2022-02-23
refactor: make timezone field 2 words

malssid:
The Rails schema had timezone as 1 word and since it's 2 words I switched it to "time_zone" to keep it consistent with other fields.


cbarber (approved):
LGTM :rocket:

PR #332, opened 2022-02-25
Added absinthe graphql resolvers

malssid:
- Added field resolvers
- Added sample JSON data for location, traffic cams, weather, and widgets.
- Changed absinthe's default middleware to look for string keys rather than atoms.


server-phoenix/lib/helios_web/schema/resolvers/event_collection.ex
```diff
@@ -0,0 +1,40 @@
...
+          @test_query_data.all
+          |> Enum.filter(fn event -> DateTime.compare(event.created_at, created_after) == :gt end)
+
+        %{created_after: created_after, type: type} ->
+          @test_query_data.all
+          |> Enum.filter(fn event -> DateTime.compare(event.created_at, created_after) == :gt end)
+          |> Enum.filter(fn event -> event.source == type end)
+
+        _ ->
+          @test_query_data.all
```
cbarber, line 31: I suspect there is a more idiomatic way to write this. Maybe using `if/2` in the pipeline or using function argument matching/extract filter functions. Asking #elixir will probably provide some good options.

server-phoenix/lib/helios_web/schema/helpers/widget_collection.ex
```diff
@@ -1,42 +1,5 @@
...
-        name: "Events",
-        durationSeconds: 20,
-        id: 6,
-        locationId: "1",
-        position: 5,
-        showWeather: true,
-        sidebarText: "Events at MojoTech"
-      }
-    ]
+
```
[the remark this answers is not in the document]
malssid: We do need the `enabled_and_available` function later on when the db is setup. Maybe I should just move it into the widget_collection resolvers module? I wanted to separate out actual field resolvers from other functions (helpers).

server-phoenix/lib/helios_web/schema/resolvers/event_collection.ex
```diff
@@ -0,0 +1,40 @@
...
+      end
+
+    %{all: new_data, count: @test_query_data.count}
```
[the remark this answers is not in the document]
malssid: new_data refers to just the event data in `all`, so since I'm not modifying the counts I just return the original `count`.

server-phoenix/lib/helios_web/schema/resolvers/event_collection.ex
malssid: This is what it looks like with argument pattern matching. I'll see what #elixir also thinks about it. 
image

server-phoenix/lib/helios_web/schema/types/unix_date_time.ex
```diff
@@ -5,7 +5,7 @@ defmodule HeliosWeb.Schema.Types.UnixDateTime do
     serialize(&serialize_time/1)
   end
 
-  defp serialize_time(%Absinthe.Blueprint.Input.Integer{value: value}) do
+  defp serialize_time(value) do
```
cbarber, line 8: Should this be a bug fix commit?

linalim (approved):

PR #333, opened 2022-02-28
Add ecto schemas and temporary sqlite3 config

linalim:
Launching IEx shell:
`iex -S mix`

Querying data examples:

- Aliasing for shortcut:
`alias Helios.{Repo, Location, ...}`

- Test out getting an instance from a model:
`locations = Repo.get(Location, 1)`

- Loading other models (use atom type):
`Repo.preload locations, :announcements`
OR for multiple preloading:
`Repo.preload locations, [:announcements, :widgets, ...]`

- SELECT, FROM, WHERE queries:
`import Ecto.Query`
`query = from(w in Widget, where: w.name == "Traffic", select: [w])
Repo.all(query)`


public/manifest.json
```diff
@@ -0,0 +1,43 @@
...
+  "media/images/commit.png": "/media/images/commit-2c662617.png",
+  "media/images/displayCommit.png": "/media/images/displayCommit-8a84c3c9.png",
+  "media/images/displayPR.png": "/media/images/displayPR-ca202dda.png",
+  "media/images/displaySlack.png": "/media/images/displaySlack-62d96504.png",
+  "media/images/logo.png": "/media/images/logo-612932ca.png",
+  "media/images/pr.png": "/media/images/pr-e808acd6.png",
+  "media/images/raincloud.png": "/media/images/raincloud-f4e49358.png",
+  "media/images/slack.png": "/media/images/slack-1765e252.png",
+  "media/images/sunmoon.png": "/media/images/sunmoon-0ecf3389.png"
+}
```
cbarber, line 43: This looks like an artifact from webpack that was missed in gitignore.

server-phoenix/config/dev.exs
```diff
@@ -2,6 +2,8 @@ import Config
 
 # Configure your database
 
+# change this to postgres -> sqlite3 temporarily (ecto setup)
+# iex -S mix (starts the phoenix)
```
cbarber, line 6: We can probably drop these comments for a note in the commit message about sqlite3 being a temporary solution

cbarber (commented):
Looking really good. I think you can squash these commits into a single commit adding the initial database modals. Maybe include an example in the PR description on how to fire up iex and query some of the data.

server-phoenix/lib/helios/announcement.ex
```diff
@@ -0,0 +1,31 @@
+defmodule Helios.Announcement do
+  use Ecto.Schema
+  import Ecto.Changeset
+  alias Helios.Announcement
+  import Ecto.Query
+
+  @primary_key {:id, :id, autogenerate: true}
+  schema "announcements" do
+    field :publish_on, :utc_datetime
```
malssid, line 9: The absinthe announcement type has this field as a `:datetime`, which is a built in scalar. I notice that they defined a `UTCDateTimeType` scalar on the rails end to use for the `publish_on` field. Then in schema.rb `publish_on` is a `t.datetime`. Do you know why the types are different in schema.rb and the graphql type definition (announcement_type.rb)?

server-phoenix/config/dev.exs
```diff
@@ -10,6 +15,10 @@ import Config
 #   show_sensitive_data_on_connection_error: true,
 #   pool_size: 10
 
+
+# Configure Time zone database
+config :elixir, :time_zone_database, Tzdata.TimeZoneDatabase
```
malssid, line 20: Already in config.exs

server-phoenix/lib/helios/widget.ex
```diff
@@ -18,4 +20,18 @@ defmodule Helios.Widget do
...
+  end
+
+  def available(query, time) when is_binary(time) do
+    td = elem(Timex.parse(time, "{ISO:Extended}"),1)
+    available(query, td)
+  end
+
+  def available(query, time) do
+      tod = DateTime.to_time(time)
+      from q in query, where: (q.start <= ^tod or is_nil(q.start)) and (q.stop >= ^tod or is_nil(q.stop))
```
cbarber, line 35: Nit: indentation here. @malssid how are you checking elixir formatting?

cbarber (commented):
Just a nit pick on indentation. There may be other formatting issues that I'd rather address now in local development and then move towards CI checking it for us (separate PR).

server-phoenix/lib/helios/widget.ex
malssid: I haven't been running `mix format` but it's something we should start doing.

server-phoenix/lib/helios/announcement.ex
```diff
@@ -15,4 +16,16 @@ defmodule Helios.Announcement do
 
     belongs_to :location, Helios.Location
   end
+
+  def with_location(query, location) do
+    from q in query, where: q.location == ^location
```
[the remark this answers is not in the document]
linalim: I think the pin operator is used here to make sure that location refers to the location from the params, and not a new variable which can hold other values.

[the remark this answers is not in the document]
malssid: Ah ok. Makes sense.

server-phoenix/mix.exs
```diff
@@ -55,7 +55,8 @@ defmodule Helios.MixProject do
       {:dotenv, "~> 3.0.0", only: [:dev, :test]},
       {:tzdata, "~> 1.1"},
       {:httpoison, "~> 1.7"},
-      {:sun_times, "~> 0.1.2"}
+      {:sun_times, "~> 0.1.2"},
+      {:timex, "~> 3.7.6"}
```
cbarber, line 59: Should the lock file have changed?

server-phoenix/lib/helios/announcement.ex
```diff
@@ -1,18 +1,31 @@
...
     field :message, :string
-    field :people, :string
+    field :people, :string, presence: true
     field :company, :string
-    field :announcement_id, :string
-    field :created_at, :utc_datetime
-    field :updated_at, :utc_datetime
+    field :announcement_id, :string, presence: true
+    field :created_at, :utc_datetime, presence: true
+    field :updated_at, :utc_datetime, presence: true
```
cbarber, line 15: It looks like you squashed these changes into the scopes commit instead of the schema commit.

server-phoenix/lib/helios/widget.ex
```diff
@@ -1,21 +1,38 @@
...
+  def available(query, time) when is_binary(time) do
+    case DateTime.from_iso8601(time) do
+      {:ok, datetime, _} -> available(query, datetime)
+      _ -> {:error, "Datetime string is not in the iso format"}
+    end
+  end
+
+  def available(query, time) do
+    tod = DateTime.to_time(time)
+    from q in query, where: (q.start <= ^tod or is_nil(q.start)) and (q.stop >= ^tod or is_nil(q.stop))
```
cbarber, line 36: :thinking: I think the return values for overloaded functions with pattern matching should probably be the same. This one returns a query and the other might return a tuple or a query.

cbarber (approved):
Awesome work! :shipit:

PR #338, opened 2022-03-03
Added SunTimes and unit test for solar cycles

malssid:
- Added SunTimes to calculate solar cycles (sunset/sunrise)
- Added a unit test to show that there will be slight differences when comparing SunTimes sunset/sunrise values to the [NOAA solar calculator](https://gml.noaa.gov/grad/solcalc/)


cbarber (commented):
I think this can be one commit since we're moving away from the API call to a calculation. More of a feature than a refactor since a refactor would keep behavior the same.

cbarber (approved):
LGTM :ship:

PR #339, opened 2022-03-03
Connect elixir backend to frontend (with static weather and traffic cam data)

malssid:
- The widget_collection resolver `by_id_or_first` now falls back to the first widget's `id` if the one passed in doesn't exist. 
- Added a traffic cam resolver and static JSON data for them.
- The default absinthe middleware can now handle key-value pairs where the value is a list.
- Configured CORS on the backend and added an "/api/graphql" endpoint for the frontend.
- Fixed inconsistencies in ruby and elixir related to argument and field names/types.


server-phoenix/lib/helios_web/schema/resolvers/event_collection.ex
```diff
@@ -20,11 +20,11 @@ defmodule HeliosWeb.Schema.Resolvers.EventCollection do
 
         %{created_after: created_after} ->
           @test_query_data["all"]
-          |> Enum.filter(fn event -> DateTime.compare(event["created_at"], created_after) == :gt end)
+          |> Enum.filter(fn event -> DateTime.compare(event["created_at"], elem(DateTime.from_iso8601(created_after), 1)) == :gt end)
 
         %{created_after: created_after, type: type} ->
           @test_query_data["all"]
-          |> Enum.filter(fn event -> (DateTime.compare(event["created_at"], created_after) == :gt) and (event["source"] == type) end)
+          |> Enum.filter(fn event -> (DateTime.compare(event["created_at"], elem(DateTime.from_iso8601(created_after), 1)) == :gt) and (event["source"] == type) end)
```
cbarber, line 27: I don't think this belongs in the commit `fix: field/argument name inconsistencies`

.env.sample
```diff
@@ -14,4 +14,4 @@ ADMIN_EMAIL=admin@example.com
 ADMIN_PASSWORD= password
 BAMBOOHR_SUBDOMAIN=<Ask a team mate>
 BAMBOOHR_API_KEY=<Ask a team mate>
-BACKEND_URI=http://localhost:5001
+BACKEND_URI=http://localhost:4000/api
```
cbarber, line 17: I'd leave the sample alone until we're ready to convert entirely to phoenix.

server-phoenix/lib/helios_web/router.ex
```diff
@@ -22,7 +22,7 @@ defmodule HeliosWeb.Router do
 
   scope "/api" do
     pipe_through(:api)
-
+    forward "/graphql", Absinthe.Plug, schema: HeliosWeb.Schema
```
cbarber, line 25: I'd split this out of ` feat: add cors and graphql api endpoint `. `and` in generally smell that two things are happening in a commit. This is separate from cors work.

server-phoenix/lib/helios_web/schema.ex
```diff
@@ -7,7 +7,12 @@ defmodule HeliosWeb.Schema do
   end
 
   def get_string_key(%{source: source} = res, key) do
-    %{res | state: :resolved, value: Map.get(source, key)}
+    cond do
+      is_map(source) ->
```
cbarber, line 11: @linalim and I discovered this can follow a method declaration like so:
```
  def get_string_key(%{source: source} = res, key) when is_map(source) do
    ....
  end
```

server-phoenix/lib/helios_web/schema/helpers/location.ex
```diff
@@ -6,7 +6,6 @@ defmodule HeliosWeb.Schema.Helpers.Location do
   def sunset(lat, long, offset \\ "today") do
     response = HTTPoison.get!("https://api.sunrise-sunset.org/json?lat=#{lat}&lng=#{long}&formatted=0&date=#{offset}")
     data = Jason.decode!(response.body)
-    IO.inspect(data)
```
cbarber, line 9: Nit: removing this probably in its own "cleanup" commit.

server-phoenix/lib/helios_web/schema/resolvers/widget_collection.ex
```diff
@@ -17,6 +17,8 @@ defmodule HeliosWeb.Schema.Resolvers.WidgetCollection do
   end
 
   def by_id_or_first(_parent, %{id: id}, _info) do
-    {:ok, Enum.find(Jason.decode!(WidgetCollectionData.enabled), fn widget -> widget["id"] === id end)}
+    decodedData = Jason.decode!(WidgetCollectionData.enabled)
+    idExists = Enum.find(decodedData, fn widget -> widget["id"] === id end)
```
cbarber, line 21: Nit: I think convention in elixir is snake case for variable names.

A unit test for this bug fix would also be great to prevent regressions when we connect the database.

server-phoenix/lib/helios_web/schema/queries/event_collection.ex
```diff
@@ -6,7 +6,7 @@ defmodule HeliosWeb.Schema.Queries.EventCollection do
   object :event_collection_queries do
     field :events, :event_collection do
       description("MojoTech slack/github events")
-      arg(:created_after, :datetime)
+      arg(:created_after, :string)
```
chewnoill, line 9: nit: could we rename this arg `:after` to match the ruby version? then we could drop all the client/ruby changes?

malssid: I should've noted it. `after` is a reserved word in elixir so that's why I chose to go with `created_after`

server-phoenix/lib/helios_web/schema.ex
```diff
@@ -7,7 +7,7 @@ defmodule HeliosWeb.Schema do
   end
 
   def get_string_key(%{source: source} = res, key) do
-    %{res | state: :resolved, value: Map.get(source, key)}
+      %{res | state: :resolved, value: Map.get(source, key)}
```
cbarber, line 10: Did you mean to indent this line here? It looks like a format issue.

server-phoenix/lib/helios_web/schema/resolvers/widget_collection.ex
```diff
@@ -5,18 +5,20 @@ defmodule HeliosWeb.Schema.Resolvers.WidgetCollection do
...
+    num_widgets = Enum.count(decoded_data)
+    case Enum.find_index(decoded_data, fn widget -> widget["id"] === current_widget_id end) do
       nil ->
-        {:ok, Enum.at(decodedData, 0)}
-      x when x === numWidgets - 1 -> {:ok, Enum.at(decodedData, 0)}
+        {:ok, Enum.at(decoded_data, 0)}
+      x when x === num_widgets - 1 -> {:ok, Enum.at(decoded_data, 0)}
       x ->
-        {:ok, Enum.at(decodedData, x + 1)}
+        {:ok, Enum.at(decoded_data, x + 1)}
```
cbarber, line 15: This commit mixed in variable renames in `next/3`

server-phoenix/lib/helios_web/schema/types/weather.ex
```diff
@@ -95,7 +101,7 @@ defmodule HeliosWeb.Schema.Types.Weather do
     field(:wind_gust, :float)
     field(:wind_deg, non_null(:integer))
     field(:clouds, non_null(:float))
-    field(:uvi, non_null(:integer))
+    field(:uvi, non_null(:float))
```
cbarber, line 104: Was this a bug? Does rails have the wrong type as well?

server-phoenix/mix.exs
```diff
@@ -55,6 +55,7 @@ defmodule Helios.MixProject do
       {:tzdata, "~> 1.1"},
       {:httpoison, "~> 1.7"},
       {:cors_plug, "~> 3.0"}
+      {:sun_times, "~> 0.1.2"}
```
cbarber, line 58: This looks like a merge commit with syntax errors. I think you need a comma above this line here. Lets prefer rebase to avoid this.

server-phoenix/lib/helios_web/schema/types/weather.ex
malssid: Rails has it as an integer. I was getting an error on elixir because it was coming in as a float. I'll have to update the rails side as well.

server-phoenix/lib/helios_web/schema/resolvers/widget_collection.ex
malssid: oops... Those should be in a separate commit.

server-phoenix/lib/helios_web/schema.ex
malssid: I need to get in a habit of running `mix format`. The only problem is I'll have to run it for individual files or it will most likely change a bunch of them and cause merge conflicts with ones that Lina might be working on.

cbarber: The commit message here is `fix: remove type checking on default absinthe middleware` and this commit just indents this line (appears to be incorrect). Could we drop the commit entirely?

server-phoenix/lib/helios_web/schema/types/weather.ex
```diff
@@ -48,14 +50,16 @@ defmodule HeliosWeb.Schema.Types.Weather do
     field(:wind_speed, non_null(:float))
     field(:wind_deg, non_null(:integer))
     field(:clouds, non_null(:integer))
-    field(:uvi, non_null(:integer))
+    field(:uvi, non_null(:float))
```
cbarber, line 53: Why did this change in ` feat: add resolver for weather_info_data `? Does it need to be in the next commit?

cbarber (commented):
Looks like a couple minor fixups and then this should be good

server-phoenix/lib/helios_web/schema/types/weather.ex
malssid: Yes! I must've missed that when splitting up the commit.

cbarber (approved):
Awesome work! :shipit:

PR #340, opened 2022-03-08
Specify node version in package json

cbarber:


package.json
```diff
@@ -107,8 +110,8 @@
     "relative-to-alias": "jscodeshift --extensions jsx -t ./config/transforms/relative_to_alias_path.js ./ ",
     "prettier:format": "prettier --write '{app/javascript/,__mocks__}/**/*.{js,jsx}' package.json webpack.config.js",
     "prettier:check": "prettier --check '{app/javascript/,__mocks__}/**/*.{js,jsx}' package.json webpack.config.js",
-    "build": "webpack",
-    "serve": "webpack-dev-server"
+    "build": "./node_modules/.bin/webpack",
```
chewnoill, line 113: nit: I think yarn will reach into this bin folder automatically. Should only need this if you're building with npm, in which case we have the wrong lock file.

chewnoill (approved):

package.json
cbarber: I'm going to remove this commit. We initially saw `$ webpack` was not found so I had incorrectly assumed yarn path manipulation wasn't working in heroku.

PR #343, opened 2022-03-09
fix: proptypes with incorrect casing of timeZone

malssid:
The time was defaulting to UTC due to propTypes and prop names using `timezone` instead of `timeZone` on the frontend.

### DIFF
--- a/app/javascript/components/date.jsx
+++ b/app/javascript/components/date.jsx
@@ -25,7 +25,7 @@ export class Date extends React.Component {
     loading: PropTypes.bool,
     error: PropTypes.bool,
     location: PropTypes.shape({
-      timezone: PropTypes.string,
+      timeZone: PropTypes.string,
     }).isRequired,
   };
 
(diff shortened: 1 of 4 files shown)


app/graphql/types/location_type.rb
```diff
@@ -4,7 +4,7 @@ class Types::LocationType < Types::BaseObject
   field "id", ID
   field "latitude", Float
   field "longitude", Float
-  field "time_zone", String
+  field "time_zone", String, method: :time_zone
```
cbarber, line 7: I don't think this is necessary.

cbarber (commented):
Nit on the commit message. This is fixing both propTypes and props which is an important distinction.

app/graphql/types/location_type.rb
malssid: Yeah, I don't think it is either.

cbarber (approved):

PR #345, opened 2022-03-10
Modify weather resolver to get data from OpenWeather API

malssid:
- Removed the deprecated moonphase field from weather type (it is part of location now)
- Removed an extra `Rails.cache.fetch` call from weather resolver
- Configured ConCache to cache the response from OpenWeather. This library uses ETS (Erlang Term Storage) to provide a key/value store.
- Modified the weather resolver to pull in data from OpenWeather api (instead of using test data).
- Each widget (in test data) now has location eagerly loaded in so the widget weather resolver can pass location data to the weather client.


server-phoenix/lib/helios_web/schema/resolvers/location.ex
```diff
@@ -12,8 +12,10 @@ defmodule HeliosWeb.Schema.Resolvers.Location do
     {:ok, Jason.decode!(LocationData.location(System.get_env("PRIMARY_CITY_NAME")))}
   end
 
-  def weather(_parent, _args, _info) do
-    {:ok, Jason.decode!(WeatherData.weather)}
+  def weather(parent, _args, _info) do
+    response = WeatherClient.forecast(parent)
+    IO.inspect(Jason.decode!(response)["daily"])
```
cbarber, line 17: Looks like a debug line snuck in.

cbarber (commented):
I changed the base branch to `ma/connect-elixir-to-frontend` so this would be easier to review. Nice atomic commits here. Just a minor debug line to remove and then this is good to go.

server-phoenix/lib/helios_web/schema/resolvers/location.ex
malssid: They're always camouflaging within the code  😡

PR #347, opened 2022-03-11
fix: fields to ecto timestamps in schemas

linalim:
Ecto.Schema.timestamps automatically fills in 'inserted_at' and 'updated_at' columns. I changed the 'inserted_at' column name to 'created_at' to match the database schema, and the type from default 'naive_datetime' to 'utc_datetime'.


malssid (approved):
LGTM!

PR #348, opened 2022-03-11
Modify weather resolver to get data from OpenWeather API

malssid:
- Removed the deprecated moonphase field from weather type (it is part of location now)
- Removed an extra Rails.cache.fetch call from weather resolver
- Configured ConCache to cache the response from OpenWeather. This library uses ETS (Erlang Term Storage) to provide a key/value store.
- Modified the weather resolver to pull in data from OpenWeather api (instead of using test data).
- Each widget (in test data) now has location eagerly loaded in so the widget weather resolver can pass location data to the weather client.


server-phoenix/lib/helios_web/schema/resolvers/widget.ex
```diff
@@ -1,13 +1,13 @@
...
+  alias HeliosWeb.Clients.WeatherClient
   alias HeliosWeb.TestData.TrafficCams, as: TrafficCamsData
 
   def day_announcements(_parent, _args, _info) do
     {:ok, []}
   end
 
-  def weather(_parent, _args, _info) do
-    {:ok, Jason.decode!(WeatherData.weather)}
+  def weather(parent, _args, _info) do
```
cbarber, line 9: I'm wondering if we should be specifying guards here to hint at the parent type.

malssid: `parent` comes in as a map (of location in this case), so I'm not sure if it's necessary.

cbarber: Won't parent be whatever the resolver sets it? `parent` being `Location` doesn't seem obvious in the widget resolver.

malssid: Ohhh... I'm sorry, I was reading this as the Location weather resolver. We loaded location onto each widget (in test data) which is how the resolver above can access the latitude and longitude of the location. 

Back to your original question, do you mean checking if parent has a `location` key?

cbarber (approved):

PR #349, opened 2022-03-11
Formatting and checking format in config.yml

linalim:
Whenever a PR is merged into master branch, will be dropping the commit, format it, then push it back.


cbarber (approved):
Nice work. Just need to keep this rebased.

PR #364, opened 2022-03-28
fix: change tweet's created_at format from string to datetime

linalim:
Changes due to tweet object having field `created_at` as type `:datetime` but api response outputs a utc datetime string.

### DIFF
--- a/server-phoenix/lib/helios_web/clients/tweet.ex
+++ b/server-phoenix/lib/helios_web/clients/tweet.ex
@@ -1,4 +1,6 @@
 defmodule HeliosWeb.Clients.Tweet do
+  use Timex
+
   defstruct [:text, :media, :interactions, :status, :user, :created_at]
 
   defmodule InteractionQuery do
@@ -20,8 +22,13 @@ defmodule HeliosWeb.Clients.Tweet do
       interactions: interactions(tweet),
       status: status(tweet),
       user: user(tweet),
-      created_at: tweet.created_at
+      created_at: created_at(tweet)
     }
+  end
+
+  def created_at(t) do
+    t.created_at
+    |> Timex.parse!("%a %b %d %T %z %Y", :strftime)
   end
 
   def attributes(t) do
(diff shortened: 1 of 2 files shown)


server-phoenix/lib/helios_web/clients/tweet.ex
```diff
@@ -20,10 +22,15 @@ defmodule HeliosWeb.Clients.Tweet do
       interactions: interactions(tweet),
       status: status(tweet),
       user: user(tweet),
-      created_at: tweet.created_at
+      created_at: created_at(tweet)
```
cbarber, line 25: Does this pass in the unit tests? Are we running the unit tests in CI?

linalim: I completely forgot about updating unit tests! I updated them now, so we're passing them. I also added them in CI.

server-phoenix/test/helios_web/clients/tweet_test.exs
```diff
@@ -235,7 +235,7 @@ defmodule HeliosWeb.Clients.TweetTest do
   test "from_api retweet status" do
     assert Tweet.from_api(test_tweet_retweet()) ===
              %HeliosWeb.Clients.Tweet{
-               created_at: "Fri Mar 18 15:56:28 +0000 2022",
+               created_at: ~U[2022-03-18 15:56:28Z],
```
cbarber, line 238: Its fine to squash the unit test change with the type change since they're tightly related e.g. tests won't pass in the commit with the type change.

cbarber (commented):
Strange compilation error in CI:
```
Unchecked dependencies for environment test:
* uuid (Hex package)
  could not find an app file at "_build/test/lib/uuid/ebin/uuid.app". This may happen if the dependency was not yet compiled or the dependency indeed has no app file (then you can pass app: false as option)
** (Mix) Can't continue due to errors on dependencies
```

malssid (commented):
nit: I noticed that rails shows tweets **and** retweets from @mojotech. Should we include retweets as well in phoenix?

cbarber:
> nit: I noticed that rails shows tweets **and** retweets from @mojotech. Should we include retweets as well in phoenix? IIRC it was only showing tweets.

We should but I'm fine with that being a separate PR/issue.

cbarber (approved):
LGTM :rocket:

PR #365, opened 2022-03-29
feat: configure frontend to work with both languages

malssid:
To switch between languages you now only have to change `BACKEND_URL` and `BACKEND_LANGUAGE` in your `.env`


app/javascript/lib/phoenix-graphql-backend.js
```diff
@@ -0,0 +1,46 @@
...
+import { hasSubscription } from '@jumpn/utils-graphql';
+import { HttpLink } from 'apollo-link-http';
+import { split } from 'apollo-link';
+import Cookies from 'js-cookie';
+import * as AbsintheSocket from '@absinthe/socket';
+import { createAbsintheSocketLink } from '@absinthe/socket-apollo-link';
+
+const webSocketURL = new URL(process.env.BACKEND_URL);
+
+webSocketURL.protocol = 'ws:';
```
cbarber, line 12: :100: nice way to switch the protocol without find/replace

app/javascript/components/app.jsx
```diff
@@ -2,78 +2,26 @@ import 'regenerator-runtime/runtime';
...
-
-const absintheLink = split(
-  operation => hasSubscription(operation.query),
-  webSocketLinkAbsinthe,
-  authedHttpLink,
-);
+if (process.env.BACKEND_LANGUAGE === 'ruby') {
+  link = railsLink;
+} else if (process.env.BACKEND_LANGUAGE === 'elixir') {
+  link = absintheLink;
```
cbarber, line 23: NIt: we could change these to requires for tree shaking but that can be an improvement for another day.

app/javascript/components/app.jsx
```diff
@@ -2,78 +2,26 @@ import 'regenerator-runtime/runtime';
...
+}
```
cbarber, line 24: Should we else here and fail loudly? Or just default to `ruby` for now?

cbarber (commented):
Looks great. Small comment on `if/else` on `process.env.BACKEND_LANGUAGE`

app/javascript/lib/phoenix-graphql-backend.js
malssid: Thanks!

app/javascript/components/app.jsx
malssid: I was looking into that... I was trying out `import()` from here [https://javascript.info/modules-dynamic-imports](https://javascript.info/modules-dynamic-imports), but I was having trouble actually getting the default import.

app/javascript/components/app.jsx
malssid: I was thinking we default to ruby for now.

cbarber (approved):
LGTM. Nice work :shipit:

PR #366, opened 2022-03-29
feat: added unit test in CI

linalim:


cbarber:
Looks like phoenix CI is failing attempting to connect to redis. Maybe we could disable that backend when testing?

cbarber (commented):
Nit: we should order the unit test fix and the redis disable before the CI commit. This is to guarantee any commit passes CI.

PR #367, opened 2022-03-30
Add announcements functionality to Slack controller

malssid:
## Setup Instructions
1. **First** follow the instructions here #354 to set up "Event Subscriptions".
2. Add the bot to the channels you want it to listen to messages in. This doesn't necessarily have to be the #helios channel, but it makes sense to communicate with the bot in the same channel you'll receive replies in.
3. Enable "Incoming Webhooks" in your Slack app settings, generate a webhook url for the #helios channel, and then update  `SLACK_WEBHOOK_URL` in your .env. This is the channel where you will see the bot's replies whether your announcement was saved or if there was an error in the command you entered.

### Add an announcement using this format:
`@Helios guests: <message> to <people> from <company> on <publish date and time> in <office location>`

- ### The `<office location>` should be either `Providence` or `Boulder`
- ### The `<publish date and time>` should look like this: 
✅ August 23rd 2019 at 09:00 am
  - ### Not like this (needs the suffix after the day)
❌ August 23 2019 at 09:00 am
  - ### Or this (needs the leading zero if hour is a single digit)
❌ August 23rd 2019 at 9:00 am

### DIFF
--- a/app/javascript/components/widgets/guests/panel.jsx
+++ b/app/javascript/components/widgets/guests/panel.jsx
@@ -6,7 +6,11 @@ import gql from 'graphql-tag';
 import styled from 'styled-components';
 import { parseDate, isInFutureToday } from '@lib/datetime';
 import { colors, weights, fontSizes, spacing } from '@lib/theme';
-import { LoadingMessage, DisconnectedMessage } from '@messages/message';
+import {
+  LoadingMessage,
+  DisconnectedMessage,
+  NoAnnouncementsMessage,
+} from '@messages/message';
 
 const subscribeAnnouncementPublished = gql`
   subscription onAnnouncementPublished {
@@ -117,8 +121,12 @@ const Guests = ({ startTimer, cityName }) => (
         return <DisconnectedMessage />;
       }
 
-      if (loading || data.location.dayAnnouncements.length === 0) {
+      if (loading) {
         return <LoadingMessage />;
+      }
+
+      if (data.location.dayAnnouncements.length === 0) {
+        return <NoAnnouncementsMessage />;
       }
 
       return (
--- a/app/javascript/components/widgets/guests/panel.jsx
+++ b/app/javascript/components/widgets/guests/panel.jsx
@@ -15,7 +15,7 @@ const subscribeAnnouncementPublished = gql`
       people
       company
       publishOn
-      id
+      announcementId
     }
   }
 `;
@@ -28,6 +28,7 @@ const getLocationAnnouncements = gql`
         people
         company
         publishOn
+        announcementId
       }
     }
   }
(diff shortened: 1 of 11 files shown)


app/javascript/components/widgets/guests/panel.jsx
```diff
@@ -158,6 +162,7 @@ const Guests = ({ cityName }) => (
 );
 
 Guests.propTypes = {
+  startTimer: PropTypes.func.isRequired,
```
cbarber, line 173: This probably belongs in the previous commit enabling the announcement widget. Likely above too.

server-phoenix/lib/helios/announcement.ex
```diff
@@ -20,11 +20,11 @@ defmodule Helios.Announcement do
     from q in query, where: q.location == ^location
   end
 
-  def happen_today(query, _time_zone) do
+  def happen_today(query) do
     today = DateTime.utc_now()
```
cbarber, line 24: I think we set this to midnight if the method is answering the question "does this announcement happen today?"

cbarber: Lets add a unit test

server-phoenix/lib/helios/announcement.ex
```diff
@@ -20,11 +20,11 @@ defmodule Helios.Announcement do
     from q in query, where: q.location == ^location
   end
 
-  def happen_today(query, time_zone) do
-    {:ok, today} = DateTime.now(time_zone)
+  def happen_today(query) do
+    today = DateTime.new!(Date.utc_today(), ~T[00:00:00], "Etc/UTC")
+    # When DB is changed to Postgres (UTC -> Naive Time zone), the following code will be used:
```
cbarber, line 25: We switched the database to postgresql. We should probably test this out again

malssid: It still works. There seems to be a problem with the weather poller though.

server-phoenix/lib/helios_web/controllers/web_hooks/slack_controller.ex
```diff
@@ -1,33 +1,175 @@
...
+            source: :slack_message,
+            external_id: event["event_ts"]
+          })
+          |> publish_event
+        else
+          {:ok, "Event Skipped"}
+        end
+
+      "app_mention" ->
+        case parse_announcement(event) do
```
chewnoill, line 56: seems like a lot of stuff is happening here. Maybe we could split it up?

server-phoenix/lib/helios/announcement.ex
cbarber: I'm not sure about that. I have weather working locally.

malssid: Oh ok. Must've been something with the env variables.

server-phoenix/lib/helios_web/controllers/web_hooks/slack_controller.ex
malssid: Yeah, I can do that. You're talking about splitting up the commit, right?

PR #369, opened 2022-04-04
Add Kaffy (admin interface) and authentication to backend

malssid:
- Users can now log into the admin dashboard by going to `/admin/login`.
- The dashboard can be accessed at `/admin`.
- Users can view, edit and delete fields in the database.


server-phoenix/mix.exs
```diff
@@ -63,7 +64,8 @@ defmodule Helios.MixProject do
       {:absinthe_phoenix, "~> 2.0.0"},
       {:uuid, "~> 1.1"},
       {:exq, "~> 0.16.1"},
-      {:quantum, "~> 3.0"}
+      {:quantum, "~> 3.0"},
+      {:kaffy, git: "https://github.com/malssid/kaffy.git"}
```
sethkrasnianski, line 68: Was it ever discussed to dogfood our own [torch](https://github.com/mojotech/torch) package instead of using kaffy? Or is this more of an experiment?

chewnoill: 😞

cathyzoller:
@malssid @cbarber @linalim  I'm working on adding a postgres db for phoenix.. eventually I think we should have the new elixir rewrite housed in a separate repo.  With that in mind, do you think the admin interface could be to use [Torch](https://github.com/mojotech/torch), as mentioned above?

cathyzoller:
@malssid I see the admin_user feature was generated with `mix phx.gen.auth`.  Once we have the postgres db, we won't need to add the tokens to sqlite, but rather run the migration for the postgres db.

malssid:
> @malssid I see the admin_user feature was generated with `mix phx.gen.auth`. Once we have the postgres db, we won't need to add the tokens to sqlite, but rather run the migration for the postgres db.

oh ok

PR #370, opened 2022-04-05
Fix weather client caching a bad response

malssid:
The weather client was caching a bad response if there was an error with OpenWeatherAPI. 
- Weather will now resolve to a GraphQL error if the response is bad (ex: API key is invalid)
- I also changed ApolloClient's errorPolicy to `none` so the `error` parameter in `useQuery` can be populated with errors. Before it was set to `all`, which renders partial data even if there's an error.
See here for more information: https://www.apollographql.com/docs/react/v2/data/queries/#inspecting-error-states

### DIFF
--- a/app/javascript/components/app.jsx
+++ b/app/javascript/components/app.jsx
@@ -52,11 +52,11 @@ const client = new ApolloClient({
   defaultOptions: {
     query: {
       fetchPolicy: 'network-only',
-      errorPolicy: 'all',
+      errorPolicy: 'none',
     },
     watchQuery: {
       fetchPolicy: 'network-only',
-      errorPolicy: 'all',
+      errorPolicy: 'none',
     },
   },
 });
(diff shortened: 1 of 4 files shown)


app/javascript/components/app.jsx
```diff
@@ -52,11 +52,11 @@ const client = new ApolloClient({
   defaultOptions: {
     query: {
       fetchPolicy: 'network-only',
-      errorPolicy: 'all',
+      errorPolicy: 'none',
     },
     watchQuery: {
       fetchPolicy: 'network-only',
-      errorPolicy: 'all',
+      errorPolicy: 'none',
```
cbarber, line 59: NIt: I think its important to distinguish this as a separate commit commenting on the mis-configuration of the apollo client for graphql errors. I link to the docs in the commit message would also be nice.

malssid: Sure, I'll separate it out and link to the docs in the commit message.

linalim (approved):
LGTM :)

cbarber (commented):
Look like `change apolloclient errorPolicy to 'none'` is changing the wrong thing

malssid:
> Look like `change apolloclient errorPolicy to 'none'` is changing the wrong thing

oh yikes.. Idk what happened.

PR #372, opened 2022-04-08
feat: weather poller worker added

linalim:
- Uses Agent to update args of field weather_published. When channel subscribed, the longitude-latitude pair is added to the list of states. When channel unsubscribed, it is removed.
- Implements Absinthe Socket Decorator and Absinthe Channel Decorator, which references code from: [this link](https://github.com/absinthe-graphql/absinthe_phoenix/issues/39#issuecomment-982310059)
- Currently, handle_in("unsubscribe") handles unsubscribing from all channels. Handling unsubscribing from each channel has not been implemented.
- To test out weather poller worker:
  - subscribe to weatherPublished using {longitude, latitude} pair in Graphiql
  - unsubscribe by exiting out of Graphiql

### DIFF
--- a/server-phoenix/config/config.exs
+++ b/server-phoenix/config/config.exs
@@ -66,5 +66,6 @@ config :cors_plug,
 
 config :helios, Helios.Scheduler,
   jobs: [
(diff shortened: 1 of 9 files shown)


cbarber (approved):
:shipit:

PR #376, opened 2022-05-14
Switch migrations to phoenix

cbarber:


cathyzoller (approved):
Hands on works well with brand-new helios_dev database.  Maybe add something to the readme for those with existing postgres schema_migrations to drop the database first and create a new one before running migrations.

`psql -U postgres -c 'DROP DATABASE IF EXISTS helios_dev;'`
`docker-compose up -d`

PR #379, opened 2022-05-18
fix: weather client sending too many api requests

malssid:
Weather client now checks the cache before making an API request.


cbarber (approved):
:rocket:

PR #385, opened 2022-05-24
Fix for #182249098 bug, involving PVD showing night time during the day

LeoSipowicz:
beginTime and endTime had swapped values which caused the sunset and sunrise times and sides to be flipped as well as the moon/sun to be opposite what it should be. Example below.
Before:
Screen Shot 2022-05-24 at 10 58 07 AM
After:
Screen Shot 2022-05-24 at 10 57 41 AM


cbarber:
This seems to resolving the symptom and not the root problem. I suspect the server response changed in the transition from rails to phoenix. It also appears that the sun icon is no longer on the arc path in your screen shots.

Lets open this back up.

PR #386, opened 2022-05-24
Revert "Fix for #182249098 bug, involving PVD showing night time during the day"

LeoSipowicz:
Reverts mojotech/helios2#385


cbarber (approved):
Approving this in case you want to revert just this commit here and work in a different branch for the server fix. Up to you.

PR #387, opened 2022-05-24
chore: bump to erlang 24.3.3

cbarber:
ubuntu has a compilation compatibilty issue with >=24.1.7 and <24.3.3


cathyzoller (approved):
Compatibility checked here: https://github.com/elixir-lang/elixir/blob/main/lib/elixir/pages/compatibility-and-deprecations.md#compatibility-between-elixir-and-erlangotp

looks good 👍

PR #391, opened 2022-05-27
Changed DB and GraphQL schemas, renaming created_at columns to inserted_at

force10269:
Because of the transition from Rails to Phoenix, we are switching from supporting the use of the created_at column for timestamps to inserted_at. Changes are reflected in both the postgres DB with a migration, as well as in the GraphQL schema for the Event and Tweet schemas. The Phoenix GraphQL schema is now being dumped into ./app/javascript/schema.json as opposed to the Rails schema


isometriks (changes requested):
Can we squash the 2nd commit into the one before it?

force10269:
> Can we squash the 2nd commit into the one before it?

Resolved!

isometriks (approved):

server-phoenix/priv/repo/migrations/20220525214039_change_column_names.exs
```diff
@@ -0,0 +1,26 @@
...
+        ALTER TABLE developer_users RENAME COLUMN created_at TO inserted_at;
+        ALTER TABLE events RENAME COLUMN created_at TO inserted_at;
+        ALTER TABLE locations RENAME COLUMN created_at TO inserted_at;
+        ALTER TABLE traffic_cams RENAME COLUMN created_at TO inserted_at;
+        ALTER TABLE users RENAME COLUMN created_at TO inserted_at;
+        ALTER TABLE widgets RENAME COLUMN created_at TO inserted_at;
+      END IF;
+    END $$;")
+  end
+end
```
cbarber, line 26: We can skip `active_admin_comments`. 

`rename/3` would leverage the ecto migration DSL and be reversible:
https://hexdocs.pm/ecto_sql/Ecto.Migration.html#rename/3

cbarber (commented):
I think we might have better diffs on `schema.json` if you use the `--pretty` option.

What do you think of this for the commit structure?
* chore: switch graphql schema to phoenix
  Do you updates to README, circleci, and dump schema.json before making any changes.
* chore: switch created_at to inserted_at
  Updates both frontend and backend references. Dumps the altered version of schema.json. `--pretty` would make the `created_at` renames stand out.

server-phoenix/priv/repo/migrations/20220525214039_change_column_names.exs
force10269: Resolved!

server-phoenix/lib/helios/announcement.ex
```diff
@@ -19,7 +19,7 @@ defmodule Helios.Announcement do
     field :people, :string
     field :company, :string
     field :announcement_id, :string
-    timestamps(inserted_at: :created_at, type: :utc_datetime)
+    timestamps(inserted_at: :inserted_at, type: :utc_datetime)
```
cathyzoller, line 22: Since the default timestamp in Phoenix is inserted_at, and the default type is :utc_datetime, could we just use `timestamps()` here instead?

server-phoenix/lib/helios/developer_user.ex
```diff
@@ -12,7 +12,7 @@ defmodule Helios.DeveloperUser do
     field :git_handle, :string
     field :slack_handle, :string
 
-    timestamps(inserted_at: :created_at, type: :utc_datetime)
+    timestamps(inserted_at: :inserted_at, type: :utc_datetime)
```
cathyzoller, line 15: same as above, also following instances of timestamps.

force10269:
> From Elixir forum, comparing utc_datetime & naive_datetime use for timestamps:
> 
> "Both :naive_datetime and :utc_datetime will map to the same database type by default, e.g. in PostgreSQL they will use timestamp (without timezone). The only difference is that the first one will only take in NaiveDateTimes and the second one will 1) assume that everything read from the DB is in UTC, and 2) only accept and read DateTimes.
> 
> What you should use in your schema and migrations depends on your existing data in the database. If you know the data is UTC (which seems to be the case), you can rewrite them as :utc_datetime. If your data is not UTC, you will need to use :naive_datetime instead."
> 
> Are current date_times UTC or does it matter?

After resetting the db and testing any difference between using `timestamps()` and `timestamps(..., type: utc_datetime)`

`timestamps()`
Screen Shot 2022-05-31 at 11 19 55 AM

`timestamps(..., type: utc_datetime)`
Screen Shot 2022-05-31 at 10 51 58 AM

It seems like there is no notable difference between these two now after the migration that renames all pertinent columns to inserted_at. We can opt for the use of `timestamps()` for simplicity

server-phoenix/lib/helios/developer_user.ex
force10269: Should be simplified now to `timestamps()` for all relevant files

cathyzoller (approved):
looks good to me! 👍

PR #394, opened 2022-06-01
Make Empty and Error pages periodically poll the server to refresh data

RVRX:
Page no longer needs to be refreshed if there was an error communicating to the DB, it will try again to refetch the data on error. This also means the webpage can be loaded before the DB server is up, and discover the widgets once DB is alive. Page can also now be opened with all widgets disabled, as enabling a widget after the fact will now work as well. 

Though note that disabling an *active* widget will still require a page refresh.

DB re-check timeout is 5 seconds. (i.e., if there is an error connecting to the DB, Apollo/React will try again every two seconds).

### DIFF
--- a/app/javascript/components/widget-controller.jsx
+++ b/app/javascript/components/widget-controller.jsx
@@ -56,12 +56,17 @@ export const WidgetController = ({ client, cityName }) => {
       variables={{ id: newWidgetId, cityName }}
       fetchPolicy={fetchPolicy}
     >
-      {({ loading, error, data }) => {
+      {({ loading, error, data, refetch }) => {
         if (loading) {
           return <LoadingDisplay cityName={cityName} loading={loading} />;
         }
 
         if (error) {
+          setTimeout(async () => {
+            await client.resetStore();
+            await refetch();
+          }, 5000);
+
           // eslint-disable-next-line
           console.error(error);
           return <DisconnectedDisplay cityName={cityName} error={error} />;


RVRX:
Marking as draft while I check out if removing cache prevents proper preloading of widgets

cbarber:
> Though note that disabling an active widget will still require a page refresh.

This is surprising. Do you know why that happens?

RVRX:
> > Though note that disabling an active widget will still require a page refresh.
> 
> This is surprising. Do you know why that happens?

Okay it appears I was slightly mistaken, it's just when there's **one widget left when the timer ends** that it will reach an unrecoverable state (i.e. require a refresh), no matter if that widget has since been disabled. Generally speaking, disabling a widget, even one currently on screen, will work fine.

app/javascript/components/widget-controller.jsx
```diff
@@ -56,12 +56,18 @@ export const WidgetController = ({ client, cityName }) => {
...
+      {/* eslint-disable-next-line no-shadow */}
+      {({ client, loading, error, data, refetch }) => {
         if (loading) {
           return <LoadingDisplay cityName={cityName} loading={loading} />;
         }
 
         if (error) {
+          setTimeout(async () => {
+            await client.resetStore();
+            await refetch();
```
LeoSipowicz, line 68: Is it possible to get stuck at these awaits, and the DisconnectedDisplay is never displayed?

RVRX: As it is asynchronous I believe it should show the `<DisconnectedDisplay/>` no matter what as it will not wait for Ln67+68 before returning.
https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#working_with_asynchronous_functions

app/javascript/components/widget-controller.jsx
```diff
@@ -56,12 +56,18 @@ export const WidgetController = ({ client, cityName }) => {
       variables={{ id: newWidgetId, cityName }}
       fetchPolicy={fetchPolicy}
     >
-      {({ loading, error, data }) => {
+      {/* eslint-disable-next-line no-shadow */}
+      {({ client, loading, error, data, refetch }) => {
```
cbarber, line 60: Looks like we disable `no-shadow` here for `client`. Is this the same `client` object that is passed into `WidgetController` above?

cbarber (commented):
Nice work on a tricky issue :100: 

The commit's body might be a good place to explain the nuance of `client.resetStore`.

app/javascript/components/widget-controller.jsx
RVRX: Yup it's the same object! Maybe I should have listened to the eslint warning 🙃 Figured it was some special Apollo call, but it works fine just reference the parent object.

Pushed new changes to reflect.

cbarber (approved):
Awesome work :ship:

PR #396, opened 2022-06-03
Fix Mojotech by the numbers commit count

RVRX:
Tracker now counts commits to both master & main.
Tests added as well.


RVRX:
tests pass fine locally ;(
image
The DB does need to be up if that has anything to do with the CI tests failing?

server-phoenix/lib/helios_web/controllers/web_hooks/github_controller.ex
```diff
@@ -32,7 +32,7 @@ defmodule HeliosWeb.WebHooks.GithubController do
   end
 
   defp github_push(payload) do
-    if payload["ref"] == "refs/heads/master" do
+    if Enum.member?(["refs/heads/master", "refs/heads/main"], payload["ref"]) do
```
cbarber, line 35: Ah! Nice catch.

We have some example github payloads in: `app/lib/simulate/`. I took a peek in them and it looks like github sends both a `default_branch` and `master_branch` value. Not sure what the difference is but I imagine they could future proof matching any github mainline branch is named.

I'm also wondering if we should just treat every `git push` as new commits.

cbarber (commented):
It looks like the tests are failing because of the CORSPlug configuration from .env. You can use boolean short circuiting to give that method call a sane default value when it returns nil.

Its fine to squash tests into the commit with the related bug fix.

server-phoenix/test/helios_web/controllers/web_hooks/github_controller_test.exs
```diff
@@ -0,0 +1,151 @@
+defmodule HeliosWeb.WebHooks.GithubControllerTest do
+  use HeliosWeb.ConnCase, async: true
+  use ExUnit.Case
+  import Ecto.Query
```
isometriks, line 4: This import isn't used anymore

server-phoenix/lib/helios_web/controllers/web_hooks/github_controller.ex
isometriks: Would treating every `git push` as new commits count force pushes for rebases etc inflating the numbers? Also missed the default/master branch fields, that does seem like a good idea.

server-phoenix/test/helios_web/controllers/web_hooks/github_controller_test.exs
cbarber: Leo is working through the warnings if you want to put this off. I'm fine with dealing with it here as well.

isometriks:
Can we just do

```
config :cors_plug,
  origin: ["*"]
```

In test config setup so CORS is disabled for tests?

isometriks:
I guess we could place an origin header as well: https://github.com/mschae/cors_plug/blob/b2d6d0f6b7ed40b4e692736010de45b848573215/lib/cors_plug.ex#L175

RVRX:
To comment on new changes:
* CORS causing CI test failure was fixed in 50355ca (per https://github.com/mojotech/helios2/pull/396#issuecomment-1147562672) and c1b7f76 (bug applicable outside this issue).
* Per https://github.com/mojotech/helios2/pull/396#discussion_r890250863, Helios now counts any changes to branches marked as 'default' in github.

PR #399, opened 2022-06-03
chore: Added FRONTEND_URL env var to Cypress and webpack

force10269:
This is done so that when we change the FRONTEND_URL in .env, Cypress runs its tests with the new url, and it is also changed for accessing locally (e.g. localhost:5000 to localhost:3000)

### DIFF
--- a/cypress/plugins/index.js
+++ b/cypress/plugins/index.js
@@ -11,6 +11,7 @@
 
 // This function is called when a project is opened or re-opened (e.g. due to
 // the project's config changing)
+require('dotenv').config();
 
 /**
  * @type {Cypress.PluginConfig}
@@ -19,4 +20,6 @@
 module.exports = (on, config) => {
   // `on` is used to hook into various events Cypress emits
   // `config` is the resolved Cypress config
-}
+  // eslint-disable-next-line no-param-reassign
+  return { ...config, env: { ...process.env, ...config.env } };
+};
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -8,6 +8,10 @@ const HtmlWebpackPlugin = require('html-webpack-plugin');
 
 require('dotenv').config({ path: path.resolve(__dirname, '.env') });
 
+const frontendUrl = new URL(
+  process.env.FRONTEND_URL || 'http://localhost:3000',
+);
+
 module.exports = {
   mode: 'development',
   output: {
@@ -85,13 +89,13 @@ module.exports = {
     quiet: false,
     disableHostCheck: true,
     host: '0.0.0.0',
-    port: 3000,
+    port: frontendUrl.port,
     https: false,
     hot: false,
     contentBase: path.resolve(__dirname, './server-phoenix/priv/static'),
     inline: true,
     useLocalIp: false,
-    public: 'localhost:3000',
+    public: frontendUrl.host,
     publicPath: '/',
     historyApiFallback: {
       disableDotRule: true,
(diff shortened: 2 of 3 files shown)


webpack.config.js
```diff
@@ -85,13 +85,13 @@ module.exports = {
     quiet: false,
     disableHostCheck: true,
     host: '0.0.0.0',
-    port: 3000,
+    port: `${process.env.FRONTEND_URL}`.replace(/\D/g, ''),
```
cbarber, line 88: Maybe `Url` is available here to do this without string replace? https://developer.mozilla.org/en-US/docs/Web/API/URL/port

cypress/plugins/index.js
```diff
@@ -19,4 +20,7 @@
 module.exports = (on, config) => {
   // `on` is used to hook into various events Cypress emits
   // `config` is the resolved Cypress config
+  // eslint-disable-next-line no-param-reassign
+  config.env.FRONTEND_URL = process.env.FRONTEND_URL;
+  return config;
```
cbarber, line 25: Could we splat `process.env` into the export in a new `config.cypress.js` file? https://docs.cypress.io/guides/guides/environment-variables#Option-1-configuration-file That would let us share `.env` every where.

webpack.config.js
force10269: Did this! Went with your suggestion for using `const frontendUrl = process.env.FRONTEND_URL || 'http://localhost:3000'` to avoid circleCI failure because of `new URL(...)` throwing an error when it builds in git since .env isn't available

webpack.config.js
```diff
@@ -85,13 +86,13 @@ module.exports = {
...
     host: '0.0.0.0',
-    port: 3000,
+    port: new URL(frontendUrl).port,
     https: false,
     hot: false,
     contentBase: path.resolve(__dirname, './server-phoenix/priv/static'),
     inline: true,
     useLocalIp: false,
-    public: 'localhost:3000',
+    public: frontendUrl,
```
cbarber, line 95: Just double checking this handles the `http://` prefix.

cypress/plugins/index.js
```diff
@@ -1,4 +1,4 @@
-/// <reference types="cypress" />
+// / <reference types="cypress" />
```
LeoSipowicz, line 1: Is this extra space a typo?

force10269: For some reason my linter was complaining about this bit, but it was fine the way it was before, so now I changed it back.

webpack.config.js
```diff
@@ -85,13 +86,13 @@ module.exports = {
     quiet: false,
     disableHostCheck: true,
     host: '0.0.0.0',
-    port: 3000,
+    port: new URL(frontendUrl).port,
```
cathyzoller, line 89: If you were to assign `new URL(frontendUrl)` to a constant on line 11, then you'll avoid duplicate instantiation of a `new URL` construct.  Then just use the const on lines 89 and 95.

cathyzoller (changes requested):
Looks good for the most part!  Just a couple of small things.

cypress/plugins/index.js
force10269: Went ahead and changed that!

webpack.config.js
force10269: Neater this way for sure. Resolved

cbarber (approved):
Awesome work :ship:

cbarber:
@cathyzoller looks like your requested changes were addressed

cathyzoller (approved):
Looks Great Korry!

PR #400, opened 2022-06-06
Fix elixir build warnings #182248985

LeoSipowicz:
Deletion of unreferenced Alias and imports. Also fixed /0 functions being called incorrectly without parentheses to remove ambiguity and clear warnings thrown on compile.


cbarber (approved):
Awesome! Thank you for cleaning these up :100:

PR #401, opened 2022-06-08
feat: capture Slack and GitHub webhook source authors

RVRX:
GitHub and Slack webhooks will now record author of the event for webhooks.
_Pair-programmed with @priscilatrevino 👍🏻 _


server-phoenix/test/helios_web/controllers/web_hooks/slack_controller_test.exs
```diff
@@ -0,0 +1,163 @@
...
+            type: "event_callback",
+            event_id: "EvXXXXXX",
+            event_time: 1_559_841_923,
+            authed_users: ["UXXXXXXXX", "UXXXXXXXX"]
+          }
+        }
+      )
+
+    messages = get_slack_messages()
+    first_message = messages |> Enum.at(0) |> IO.inspect()
```
cbarber, line 154: Do we need `IO.inspect()`?

server-phoenix/test/helios_web/controllers/web_hooks/slack_controller_test.exs
```diff
@@ -0,0 +1,163 @@
...
+    assert Map.fetch(first_message, :source_author) == {:ok, "Priscila"}
```
cbarber, line 155: Its looks like this assertion is the only difference from `"check for data in table",`. I'm ok with multiple assertions in the test e.g. a row must be present before you assert the author is set. If you want to keep the separate tests, you could extract a setup method:
```
describe "when a slack web hook event is submitted" do
  setup [:post_slack_message]

  test ...

  test ...
end

defp post_slack_message(%{conn: conn}) do
  ...
end
```

server-phoenix/test/helios_web/controllers/web_hooks/slack_controller_test.exs
RVRX: was unnecessary; removed

server-phoenix/test/helios_web/controllers/web_hooks/slack_controller_test.exs
RVRX: Good point. Combined those two tests into one.

RVRX:
first force-push to e86ae35 was addressing Barber's comments, second to 41c042b was resolving some linter warnings in the tests

server-phoenix/test/helios_web/controllers/web_hooks/slack_controller_test.exs
```diff
@@ -0,0 +1,113 @@
...
+          authed_users: ["UXXXXXXXX", "UXXXXXXXX"],
+          slack: %{
+            token: "Sim-123456",
+            team_id: "TXXXXXXXX",
+            api_app_id: "AXXXXXX",
+            event: %{
+              client_msg_id: "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX",
+              type: "message",
+              text: "hi",
+              user: "UXXXXXXXX",
```
cbarber, line 39: Question for follow up work. Do we get a user identifier or a name in the above `event` map?

cbarber (approved):
Great job :ship:

PR #402, opened 2022-06-13
Wo capture images

wogara:
this ticket will check if an image was sent and make a post request to that images url with the bearer token associated with the app. to use properly, this token must be stored in the .env file.


server-phoenix/lib/helios_web/controllers/web_hooks/slack_controller.ex
```diff
@@ -1,12 +1,36 @@
...
+          fn ->
+            try do
+              response = HTTPoison.post!(url, [], headers, follow_redirect: true)
+            rescue
+              e -> IO.inspect(e)
+            end
+          end
+        )
+      else
+        Logger.info("not an image")
```
cbarber, line 32: Nit: we'll probably want to extract this as its own function in future commits. Fine for now.

cbarber (approved):
Nice job! :rocket:

PR #403, opened 2022-06-14
feat: add graphql high-score query

RVRX:
### ~~Note 1~~
_**Edit: See [comment below](https://github.com/mojotech/helios2/pull/403#issuecomment-1156899548)**_

~~The ticket originally called to bale able to run a GraphQL query of approximately the format:~~
```gql
query getTopAuthors {
  events(createdAfter: "2022-06-14T14:17:47Z") {
    byAuthor(top: 5) {
      author
      count
    }
  }
}
```
~~I've instead provided a query that is used like so:~~
```gql
query getTopAuthors {
  events(top: 5, createdAfter: "2022-06-14T14:17:47Z") {
    byAuthor {
      sourceAuthor
      count
    }
  }
}
```
~~The big difference being that the argument is passed to `events()` not to `byAuthor()`. This is due to difficulties with getting the argument in the [`event_collection` resolver](https://github.com/mojotech/helios2/blob/master/server-phoenix/lib/helios_web/schema/resolvers/event_collection.ex). I am unsure of the feasibility of doing it the planned way; any suggestions are welcome, but note that the current implementation does appear to be working fine, just maybe not set up as would be ideal.~~

### Note 2
the function `top_commits_by_author()` in `helios/event.ex` uses an [Ecto.Fragment](https://hexdocs.pm/ecto/Ecto.Query.html#module-fragments) (raw SQL statement), as it [wasn't possible](https://angelika.me/2016/09/10/how-to-order-by-the-result-of-select-count-in-ecto/#select-with-an-alias) to do the full statement with Ecto's query syntax. So if the DB Ecto points to is changed to something with non-standard sql different there is a [small?] chance that the statement will not work - but the unit test should flag this fine

### Note 3
`schema.graphql`. My IDE (IntelliJ) re-generated that file, so there are more non-meaningful changes (indentation, spacing...) than I'd like, but I am unsure how else to generate that file from the `schema.json`. If anyone knows a better way, that'd be appreciated.

### Testing
- Use the [graphiql](http://localhost:3000/graphiql) interface running on localhost with the above query.
- A unit test was added and can be run with the usual `mix test`

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -26,15 +26,22 @@ type Announcement {
     publishOn: DateTime!
 }
 
+type AuthorContribution {
+    count: Int
+    sourceAuthor: String
+}
+
 type Event {
     externalId: String!
     id: ID!
     insertedAt: DateTime!
     source: EventSource!
+    sourceAuthor: String!
 }
 
 type EventCollection {
     all: [Event]!
+    byAuthor(top: Int): [AuthorContribution]
     count: EventCount!
 }
 
--- a/schema.graphql
+++ b/schema.graphql
@@ -1,369 +1,312 @@
+# This file was generated based on "schema.json". Do not edit manually.
+
+schema {
+    query: RootQueryType
+    subscription: RootSubscriptionType
+}
+
+"A single tab or panel of Helios"
+interface Widget {
+    durationSeconds: Int!
+    id: Int!
+    locationId: String!
+    name: String!
+    position: Int!
+    showWeather: Boolean
+    sidebarText: String
+}
+
 type Announcement {
-  announcementId: String!
-  company: String!
-  id: ID!
-  locationId: Int!
-  message: String!
-  people: String!
-  publishOn: UTCDateTime!
-}
-
-scalar DateTime
-
-type Employee {
-  displayName: String!
-  id: ID!
-  isPhotoUploaded: String!
-  photoUrl: String!
-}
-
-type EmployeeEventCollection {
-  anniversaryEmployees: [Employee!]!
-  birthdayEmployees: [Employee!]!
-  timeOffEmployees: [Employee!]!
+    announcementId: String!
+    company: String!
+    id: ID!
+    locationId: String!
+    message: String!
+    people: String!
+    publishOn: DateTime!
 }
 
 type Event {
-  insertedAt: DateTime!
-  externalId: String!
-  id: ID!
-  source: EventSource!
+    externalId: String!
+    id: ID!
+    insertedAt: DateTime!
+    source: EventSource!
 }
 
 type EventCollection {
-  all: [Event!]!
-  count: EventCount!
+    all: [Event]!
+    count: EventCount!
 }
 
 type EventCount {
-  githubCommit: Int!
-  githubPull: Int!
-  slackMessage: Int!
-}
-
-enum EventSource {
-  """
-  Github commit
-  """
-  githubCommit
-
-  """
-  Github pull request
-  """
-  githubPull
-
-  """
-  Slack message
-  """
-  slackMessage
+    githubCommit: Int
+    githubPull: Int
+    slackMessage: Int
 }
 
 type EventsWidget implements Widget {
-  durationSeconds: Int!
-  id: Int!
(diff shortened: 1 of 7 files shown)


server-phoenix/lib/helios_web/schema/resolvers/event_collection.ex
```diff
@@ -24,6 +24,8 @@ defmodule HeliosWeb.Schema.Resolvers.EventCollection do
     {:ok,
      %{
        all: Repo.all(event_query),
+       by_author:
+         event_query |> Event.top_commits_by_author(Map.get(args, :top, 3)) |> Repo.all(),
        count: Event.count(Event) |> Repo.all() |> Enum.into(%{})
```
cbarber, line 29: What if we refactored this to:
* return `{:ok, query}`
* name a sub-schema to represent the result of events. `EventCollectionAggregation`?
* extract `all` and `count` as their own fields with resolvers into `EventCollectionAggregation`. The first argument to the resolver is `parent` and should be the resolution of the parent object, in this case `{:ok, query}`

Then you could add `by_author` as a field with its own arguments and resolver.

RVRX:
Okay, some new changes: Big difference being that **Note 1** from the PR description above no longer applies. The graphQL query now works as originally wanted. You can also query with `byAuthor` without supplying the `top: #` argument, which will just return a sorted list of all the authors and their event counts.

RVRX:
Looks like I still might have some work to do here... not seeing any stats on the by the numbers page when I'm running it locally, so hold off on merging this for now.

server-phoenix/test/helios_web/schema/resolvers/event_collection_test.exs
```diff
@@ -0,0 +1,98 @@
...
+      Routes.github_path(conn, :handle),
+      %{
+        action: "opened",
+        pull_request: %{
+          url: "https://api.github.com/repos/Codertocat/Hello-World/pulls/1",
+          id: "279147437",
+          user: %{login: "RVRX"}
+        }
+      }
+    )
```
cbarber, line 24: You could create a method that just calls `Repo.insert` for an `Event` created by passed in args. No need to go to the controller for this.

cbarber (commented):
Great commit structure! Just a minor improvement in the tests for the high score query

server-phoenix/test/helios_web/schema/resolvers/event_collection_test.exs
RVRX: good suggestion! fixed and looking a lot cleaner now, added an additional test for some more coverage as well

server-phoenix/lib/helios_web/schema/resolvers/event_collection.ex
```diff
@@ -21,10 +21,18 @@ defmodule HeliosWeb.Schema.Resolvers.EventCollection do
...
+  def count(parent, _args, _info) do
     {:ok,
-     %{
-       all: Repo.all(event_query),
-       count: Event.count(event_query) |> Repo.all() |> Enum.into(%{})
-     }}
+     parent
+     |> Event.count()
+     |> Repo.all()
+     |> Enum.into(%{})}
```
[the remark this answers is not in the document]
RVRX: yup, good catch! ty

schema.graphql
```diff
@@ -1,369 +1,319 @@
-type Announcement {
-  announcementId: String!
-  company: String!
-  id: ID!
-  locationId: Int!
-  message: String!
-  people: String!
-  publishOn: UTCDateTime!
+# This file was generated based on "schema.json". Do not edit manually.
```
chewnoill, line 1: This is a big change, was the old version of this generated from the rails endpoint? Might be nice to build this once before changing the schema and commit that on its own. Just so we can have a cleaner diff on this `feat: add graphql high score query` commit

chewnoill (approved):
Looks good, nice work @RVRX!

schema.graphql
RVRX: Yup good idea, pulled out the initial regeneration into a8cfdba

PR #406, opened 2022-06-17
Update projected precipitation to output rain and snow

RVRX:
Added type (rain, snow, sleet, and thunder) to the precipitation description.

Had to pull out the hourly data points to get this, as the minutely data does not contain that level of accuracy. So I'm taking the first hour's precipitation type as the one used in the description of upcoming weather. The start/end of the precipitation is still tracked by the minutely data.

### DIFF
--- a/app/javascript/components/widgets/weather/current-weather.jsx
+++ b/app/javascript/components/widgets/weather/current-weather.jsx
@@ -17,6 +17,11 @@ export const getCurrentWeather = gql`
     minutely {
       precipitation
     }
+    hourly {
+      weather {
+        id
+      }
+    }
   }
 `;
 
@@ -49,6 +54,17 @@ export const AnticipatedPrecipitation = ({ weather }) => {
   if (precip.length === 0) {
     return '';
   }
+  let precipitationType = 'precipitation';
+  const weatherId = weather.hourly[0].weather.id;
+  if (weatherId >= 200 && weatherId <= 232) {
+    precipitationType = 'rain and thunderstorms';
+  } else if (weatherId >= 300 && weatherId <= 531) {
+    precipitationType = 'rain';
+  } else if (weatherId >= 600 && weatherId <= 602) {
+    precipitationType = 'snow';
+  } else if (weatherId >= 611 && weatherId <= 613) {
+    precipitationType = 'sleet';
+  }
   const averagePrecipitation =
     precip.reduce((acc, val) => acc + val.value, 0) / precip.length;
   let descStart = `in ${precip[0].minute} minutes`;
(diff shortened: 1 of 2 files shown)


app/javascript/components/widgets/weather/current-weather.jsx
```diff
@@ -49,6 +54,28 @@ export const AnticipatedPrecipitation = ({ weather }) => {
   if (precip.length === 0) {
     return '';
   }
+  let precipitationType = 'precipitation';
+  if (
+    weather.hourly[0].weather.id >= 200 &&
```
isometriks, line 59: Might be a little less verbose to do something like `const weatherId = weather.hourly[0].weather.id`

RVRX: introduced variable 👍🏻

app/javascript/components/widgets/weather/current-weather.test.jsx
```diff
@@ -27,12 +41,147 @@ describe('AnticipatedPrecipitation component', () => {
...
+    let byHour =
+      '[\n' +
+      '    {\n' +
+      '      "weather": {\n' +
+      '        "id": 804\n' +
+      '      }\n' +
+      '    },\n' +
+      '    {\n' +
+      '      "weather": {\n' +
+      '        "id": 804\n' +
```
cbarber, line 53: 804 is clouds according to this:
https://openweathermap.org/weather-conditions#Weather-Condition-Codes-2

Should we be showing the precipitation message in this case?

app/javascript/components/widgets/weather/current-weather.test.jsx
```diff
@@ -27,12 +41,147 @@ describe('AnticipatedPrecipitation component', () => {
...
+      '      }\n' +
+      '    }]';
+    byHour = JSON.parse(byHour);
```
cbarber, line 56: This is repeated a couple times and effectively gets an array of POJOs. Could you make a function that accepts two ids and returns an array of POJOs?

app/javascript/components/widgets/weather/current-weather.test.jsx
```diff
@@ -10,13 +10,57 @@ describe('AnticipatedPrecipitation component', () => {
...
```
chewnoill, line 25: Maybe just start off as a json object instead of a string that gets parsed?

app/javascript/components/widgets/weather/current-weather.test.jsx
chewnoill: +1

app/javascript/components/widgets/weather/current-weather.test.jsx
RVRX: Yeah that would've made a little more sense ;)
Changed.

app/javascript/components/widgets/weather/current-weather.test.jsx
RVRX: Done.

chewnoill (approved):
🚀

RVRX:
huh that wasn't a new change - just a quick rebase from master, I guess the CI got an update, lol?

RVRX:
Ahh, yup, the lint package was just bumped in a merged PR

PR #407, opened 2022-06-20
fix: esbuild backend port not grabbing assets

force10269:
After testing with the backend port, @cbarber realized that there was an issue with Phoenix being able to grab assets. The fix to this is to put `publicPath: '/assets'` into our `esbuild.config.js` file and then restructure where our `index.html` file goes. The backend port should be successfully grabbing those assets now


cbarber (approved):
Nice fix :rocket:

PR #408, opened 2022-06-20
Upgraded npm modules and resolved security vulnerabilities

force10269:
I was able to upgrade the npm modules just by using `yarn upgrade`, and then was able to resolve all security vulnerabilities by using the plugin `yarn-audit-fix` with `yarn-audit-fix --audit-level moderate/high` and checking the vulnerabilities with `yarn audit`. There were a couple of sub-dependencies that I needed to manually add into `package.json` which can be seen under `resolutions`


package.json
```diff
@@ -7,6 +7,7 @@
   "dependencies": {
     "@absinthe/socket": "^0.2.1",
     "@absinthe/socket-apollo-link": "^0.2.1",
+    "@apollo/client": "^3.6.8",
```
cbarber, line 10: Does this replace the old `apollo-client` library? If so, can you remove the old libraries?

app/javascript/components/widget-transition-controls.jsx
```diff
@@ -162,7 +162,7 @@ WidgetTransitionControls.propTypes = {
   cityName: PropTypes.string.isRequired,
   location: PropTypes.shape({
     widgets: PropTypes.shape({
-      enabled: PropTypes.array.isRequired,
+      enabled: PropTypes.arrayOf.isRequired,
```
cbarber, line 165: Lets split out the lint changes to a commit that precedes to eslint upgrade.

package.json
```diff
@@ -110,6 +110,13 @@
     "sass-loader": "^7.1.0",
     "style-loader": "^0.23.1"
   },
+  "resolutions": {
+    "live-server/**/glob-parent": "^6.0.2",
+    "eslint-plugin-graphql/**/node-fetch": "^3.2.6",
+    "eslint-plugin-graphql/**/cross-fetch": "^3.1.5",
+    "react-svg-loader/**/nth-check": "^2.1.1",
+    "eslint-plugin-graphql/**/ws": "^8.8.0"
```
cbarber, line 118: Could we bump the eslint-plugin-graphql version since its a dev dependency?

app/javascript/components/widget-transition-controls.jsx
force10269: There's actually a lot of these changes, but they are in their own commit now!

package.json
force10269: One of the tests in jest apparently relies on `@apollo/client` specifically, and we are using the `apollo-client` library. I played around with it and I couldn't find a way to make the test rely on the `apollo-client` library; it seems like there was some sub-dependency there that was specifically trying to call `@apollo/client` as modules were getting updated

chewnoill (approved):
🥇

PR #409, opened 2022-06-21
Add REST API to create commit and PR events.

RVRX:
#### **_OUTDATED, see [comment](https://github.com/mojotech/helios2/pull/409#issuecomment-1163466814)_**
**Endpoint**: `/api/v1/events/{id}`
Sending a POST request here with **body fields**:  `"event_type"` (one of either `"git_commit"` or `"git_pr"`) and `"author"` will add a new row to the DB.

### Example
```
POST localhost:4000/api/v1/events/2113728f27ae82c7b1a177c8d03f9e96e0adf246
```
```json
{
	"event_type": "git_commit",
	"author": "RVRX"
}
```
Will add:
| ... | source  | external_id | ... | source_author|
| -- | ------- | ------------| -- | --------------|
| ... | github_commit | 2113728f27ae82c7b1a177c8d03f9e96e0adf246 | ... | RVRX

#### Note 1
The ticket for the PR end of this feature noted that:
> We'll need to come up with some unique id for a branch since this isn't internal to git. Something like sha256(git remote origin + git branch name) may work but would mean we'd need to duplicate the same process in the github webhook for PRs.

My interpretation of was that this would be dealt with on the client-side/whomever is sending the POST, but if we want to add the remote and branch name as fields, and calculate the ID off that, we could.

#### Note 2
Also my interpretation of a "REST API" was me googling it and deciding it sounded just like any other API I've used/created and building it as such. Wouldn't be surprised if I incorrectly implemented something on this front.

### DIFF
--- a/server-phoenix/lib/helios_web/controllers/events_api/events_api_controller.ex
+++ b/server-phoenix/lib/helios_web/controllers/events_api/events_api_controller.ex
@@ -1,0 +1,54 @@
+defmodule HeliosWeb.EventsAPI.EventsAPIController do
+  use HeliosWeb, :controller
+  alias Helios.{Repo, Event}
+
+  def handle(conn, params) do
+    cond do
+      params["version"] == "1" -> api_v1(conn, params)
+      true -> send_resp(conn, 404, "Not Found")
+    end
+  end
+
+  defp api_v1(conn, params) do
+    case params["event_type"] do
+      "github_commit" ->
+        post_event(
+          conn,
+          Event.commits(Event),
+          :github_commit,
+          params["event_id"],
+          params["author"]
+        )
+
+      _ ->
+        send_resp(conn, 400, "Bad Request")
+    end
+  end
+
+  defp post_event(conn, source_type_query, source, external_id, source_author) do
+    unless(
+      source_type_query
+      |> Event.with_external_id(external_id)
+      |> Repo.exists?()
+    ) do
+      Repo.insert!(%Event{
+        source: source,
+        external_id: external_id,
+        source_author: source_author
+      })
+      |> publish
+
+      send_resp(conn, 201, "Created")
+    else
+      send_resp(conn, 208, "Already Reported")
+    end
+  end
+
+  defp publish(event) do
+    Absinthe.Subscription.publish(
+      HeliosWeb.Endpoint,
+      event,
+      event_published: "all"
+    )
+  end
+end
--- a/server-phoenix/lib/helios_web/controllers/events_api/events_api_controller.ex
+++ b/server-phoenix/lib/helios_web/controllers/events_api/events_api_controller.ex
@@ -20,6 +20,15 @@ defmodule HeliosWeb.EventsAPI.EventsAPIController do
           params["author"]
         )
 
+      "github_pr" ->
+        post_event(
+          conn,
+          Event.pull_requests(Event),
+          :github_pull,
+          params["event_id"],
+          params["author"]
+        )
+
       _ ->
         send_resp(conn, 400, "Bad Request")
     end
--- a/server-phoenix/lib/helios_web/router.ex
+++ b/server-phoenix/lib/helios_web/router.ex
@@ -31,6 +31,10 @@ defmodule HeliosWeb.Router do
     post("/publish", PublishController, :handle)
   end
 
+  scope "/api/v:version/events", HeliosWeb.EventsAPI do
+    post("/", EventsAPIController, :handle)
+  end
+
   # Other scopes may use custom stacks.
   # scope "/api", HeliosWeb do
   #   pipe_through :api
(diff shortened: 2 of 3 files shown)


RVRX:
Just updated the API endpoint and fields to reflect an issue that is coming up in preparation for API Auth.

the `event_id` field was moved from the URL to the JSON body as `"event_id"`.

This was done so that the JSON body will be unique and so the same signature can no longer be used to authenticate different commit ids.

server-phoenix/lib/helios_web/controllers/events_api/events_api_controller.ex
```diff
@@ -0,0 +1,44 @@
+defmodule HeliosWeb.EventsAPI.EventsAPIController do
+  use HeliosWeb, :controller
+  alias Helios.{Repo, Event}
+
+  def handle(conn, params) do
+    cond do
+      params["version"] == "1" -> api_v1(conn, params)
```
cbarber, line 7: Hah! That's some forward thinking :100:

server-phoenix/lib/helios_web/controllers/events_api/events_api_controller.ex
```diff
@@ -0,0 +1,44 @@
...
+    ) do
+      Repo.insert!(%Event{
+        source: :github_commit,
+        external_id: params["event_id"],
+        source_author: params["author"]
+      })
+      |> publish
+    end
+
+    send_resp(conn, 201, "Created")
```
cbarber, line 35: I wonder if there is value for an HTTP result if it already exists. I feel like that could be useful for debugging and probably isn't an error/failure since its a common case when pushing up a branch. 409 Conflict seems out. 303 See Other? We don't really want to redirect though. :thinking:

server-phoenix/lib/helios_web/controllers/events_api/events_api_controller.ex
```diff
@@ -34,6 +35,24 @@ defmodule HeliosWeb.EventsAPI.EventsAPIController do
...
+      |> Event.with_external_id(params["event_id"])
+      |> Repo.exists?()
+    ) do
+      Repo.insert!(%Event{
+        source: :github_pull,
+        external_id: params["event_id"],
+        source_author: params["author"]
+      })
+      |> publish
+    end
```
cbarber, line 51: This is repeated from commit. Could we extract a method that passes in the different source?

cbarber (commented):
> My interpretation of was that this would be dealt with on the client-side/whomever is sending the POST, but if we want to add the remote and branch name as fields, and calculate the ID off that, we could.

This is correct.

> Also my interpretation of a "REST API" was me googling it and deciding it sounded just like any other API I've used/created and building it as such. Wouldn't be surprised if I incorrectly implemented something on this front.

REST can be rabbit hole if go down the complexities of nesting resources, eager joining resources, and HATEOAS. These might be worthwhile to know of for REST but to be frank, its diminishing returns for the sake of API "purity". What you made is perfectly fine and straight forward for a REST API. Well done. :100:

server-phoenix/lib/helios_web/controllers/events_api/events_api_controller.ex
chewnoill: 208 Already Reported (WebDAV; RFC 5842) sounds like it makes sense

RVRX: Yup 208 looks best suited. Added.

server-phoenix/lib/helios_web/controllers/events_api/events_api_controller.ex
RVRX: Was about to say that would be a rather verbose method as `:commit` and `:pr` need their own get_all query, and that it wouldn't work out the box for other `Event` types; but TIL you can pass functions in Elixir! Updated as follows:
```elixir
defp post_event(conn, get_all_of_source_type, source, external_id, source_author) do
    unless(
      Event
      |> get_all_of_source_type.()
      |> Event.with_external_id(external_id)
      |> Repo.exists?()
    ) do
      Repo.insert!(%Event{
        source: source,
        external_id: external_id,
        source_author: source_author
      })
      |> publish
...
```

server-phoenix/lib/helios_web/controllers/events_api/events_api_controller.ex
```diff
@@ -0,0 +1,55 @@
...
+
+      _ ->
+        send_resp(conn, 400, "Bad Request")
+    end
+  end
+
+  defp post_event(conn, get_all_of_source_type, source, external_id, source_author) do
+    unless(
+      Event
+      |> get_all_of_source_type.()
```
cbarber, line 40: You can just pass in the Ecto Query from `Event.commits(Event)`:
```
post_event(
  conn,
  Event.commits(Event),
  ....

  defp post_event(conn, source_type_query, source, external_id, source_author) do
    ...
    unless(
      source_type_query
      |> Event.with_external_id(external_id)
      ...
  ```

RVRX: updated!

PR #411, opened 2022-06-22
chore: Added priv/static/index.html to server-phoenix gitignore

force10269:
This should have been added into #405 since we are now creating an `index.html` file in `server-phoenix/priv/static` with esbuild


cbarber (approved):
Nice quick fix :shipit:

PR #412, opened 2022-06-22
feat: Added a carousel to display tweets

priscilatrevino:
I was able to display Mojo's tweets in a carousel that is set on a timed interval. All five tweets can be seen before it moves onto the next widget. This is a first draft and there are definitely improvements that can made visually. Any and all feedback would be great!
*Note (things that could be improved on that I noticed so far):
- Making the first tweet (a text tweet) be centered in the carousel 
- Increasing the size of the retweet and favorite icons. Also increasing the opacity of the icons so they can be seen more easily. 

https://user-images.githubusercontent.com/58053105/175128566-efb364c8-8030-426e-a279-dabff25f3659.mov

### DIFF
--- a/app/javascript/components/widgets/twitter/tweet_profile.jsx
+++ b/app/javascript/components/widgets/twitter/tweet_profile.jsx
@@ -7,7 +7,7 @@ const ProfileWrapper = styled.div`
   display: flex;
   flex-wrap: no-wrap;
   justify-content: flex-start;
-  width: 18vw;
+  width: 15vw;
   flex-direction: row;
   align-items: flex-end;
   margin-bottom: ${(props) => (props.primary ? '48px' : `${spacing.xl}`)};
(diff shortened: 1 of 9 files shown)


app/javascript/components/widgets/twitter/tweet_profile.jsx
```diff
@@ -7,40 +7,39 @@ const ProfileWrapper = styled.div`
   display: flex;
   flex-wrap: no-wrap;
   justify-content: flex-start;
-  width: 18vw;
+  width: 15vw;
   flex-direction: row;
   align-items: flex-end;
   margin-bottom: ${(props) => (props.primary ? '48px' : `${spacing.xl}`)};
```
cathyzoller: Is this still needed since the check for `props.primary` has been removed elsewhere?

cathyzoller: nvmd... I see it is deleted in the next commit. :)

cathyzoller (approved):
Excellent work Priscilla.  Much nicer UI

PR #414, opened 2022-06-23
Wo/capture channel names events

wogara:


wogara:
need to add unit tests

PR #417, opened 2022-06-27
Wo/capture channel names events

wogara:
potentially fixed the failed test by updating the schema


server-phoenix/priv/repo/migrations/20220608170815_add_author_col_to_events.exs
```diff
@@ -4,6 +4,7 @@ defmodule Helios.Repo.Migrations.AddAuthorColToEvents do
   def change do
     alter table("events") do
       add(:source_author, :string)
+      add(:source_channel, :string)
```
chewnoill, line 7: We should not be updating existing migrations. If we need to make more changes we need to create a new migration

server-phoenix/lib/helios/slack_channel_names.ex
```diff
@@ -0,0 +1,27 @@
...
+
+  def with_id(query, channel_id) do
+    from(q in query, where: q.channel_id == ^channel_id, select: {q.channel_id})
+  end
+
+  def name_to_id(query, channel_id) do
+    from(q in query, where: q.channel_id == ^channel_id, select: q.channel_name)
+  end
+
+  # if no slack id and if slack id exists
```
cathyzoller, line 23: Go ahead and take out commented code.  If it belongs to another commit, just stage the lines you need instead of commenting.

server-phoenix/test/helios_web/controllers/web_hooks/slack_controller_test.exs
```diff
@@ -105,6 +108,337 @@ defmodule HeliosWeb.WebHooks.SlackControllerTest do
...
+                follow_redirect: true ->
+        %HTTPoison.Response{
+          body: "{
+            \"ok\": \"true\",
+            \"channel\": {
+
+            }
+        }"
+        }
+      end do
```
cathyzoller, line 218: Could you explain the syntax of this with_mock function to me, either here or in slack?

wogara: the slack controller makes an api call to slack, and the with_mock function allows me to circumvent this actual call and replace it with this "mock" call.

server-phoenix/lib/helios/slack_channel_names.ex
```diff
@@ -14,4 +14,8 @@ defmodule Helios.SlackChannelNames do
   def with_id(query \\ __MODULE__, channel_id) do
     from(q in query, where: q.channel_id == ^channel_id, select: {q.channel_id})
   end
+
+  def name_to_id(query, channel_id) do
```
chewnoill, line 18: this seems confusing, are we taking an id and returning a name? might make more sense to name this `id_to_name`

server-phoenix/lib/helios_web/controllers/web_hooks/slack_controller.ex
```diff
@@ -61,14 +61,20 @@ defmodule HeliosWeb.WebHooks.SlackController do
...
       unless Event
              |> Event.slack_messages()
              |> Event.with_external_id(event["event_ts"])
              |> Repo.exists?() do
         Repo.insert!(%Event{
           source: :slack_message,
           external_id: event["event_ts"],
-          source_author: event["user"]
+          source_author: event["user"],
+          source_channel: event_channel_name
```
chewnoill, line 77: Names are usually a display level concern. My concern here is that the name of the channel can change, and then these events would all get orphaned. 

Would it make sense to store the `id` instead? Then when somebody asks for the list of events, we can join with the SlackChannelNames table to get the name from the id?

server-phoenix/lib/helios/slack_channel_names.ex
cathyzoller: ... or could also name it name_from_id 🤷

server-phoenix/lib/helios_web/controllers/web_hooks/slack_controller.ex
cathyzoller: hmmm... this event table seems more like a logger.  Perhaps, in this case, it's ok to have orphaned names when users &/or channels are removed or changed.

cathyzoller (approved):
Looks great 👍  
Nit:  naming convention which could be changed.. If you decide to change it, then let me know & I can re-approve.

PR #419, opened 2022-06-27
feat: Google authentication added to backend port

force10269:
Make sure to run `cd server-phoenix && mix deps.get` to obtain the elixir-google-auth dependency for the Phoenix server. After running `mix phx.server` with `yarn serve` and going to your `BACKEND_URL`, it should look like this:

Screen Shot 2022-06-27 at 3 45 28 PM

After signing in with Google, it will redirect you to this page, from which you can access either the Boulder or Providence page for Helios:

Screen Shot 2022-06-27 at 3 45 36 PM

`INTEGRATIONS.md` was also updated to reflect the new changes to `.env` for the Google API keys. There, it should show steps for creating a Google Cloud application and obtaining the API keys properly.


server-phoenix/lib/helios_web/router.ex
```diff
@@ -66,6 +66,8 @@ defmodule HeliosWeb.Router do
   end
 
   scope "/", HeliosWeb do
+    pipe_through :browser
```
RVRX, line 69: will adding a pipe to `:browser` cause any unintended effects to the regular `/*path`? Maybe place this under a new scope?

force10269: Based off of the docs, I'm actually not quite sure why `pipe_through :browser` was not already in there for that scope, so I'm keeping that there, but the `GoogleAuthController` now has its own scope

chewnoill: If we're adding this for a reason outside the commit message it would make sense to put it on its own commit.

chewnoill (approved):
Nice work @force10269!

server-phoenix/lib/helios_web/router.ex
RVRX: > I'm actually not quite sure why `pipe_through :browser` was not already in there for that scope

I believe `:browser` is a builtin pipe that was auto generated, and we never _needed_ to use it, but it looks like it adds some extra security, so yeah we should probably add it to the `/*pages` scope.

cathyzoller:
In your description, in general, you'll want to add necessary steps for local testing, including any dependency updates, what url(s) should be visited etc.  EG  `cd server-phoenix && mix deps.get`.

PR #420, opened 2022-06-28
Wo/slack id to name

wogara:
created a slack channel names an id table. changed the task supervisor name and updating the slack controller to reflect that change.


server-phoenix/lib/helios/application.ex
```diff
@@ -19,6 +19,7 @@ defmodule Helios.Application do
       # Start the Ecto repository
       Helios.Repo,
       # Start the Telemetry supervisor
+      {Task.Supervisor, name: SlackImageDownloader},
```
chewnoill, line 22: nit: commit message is calling this a `rename` but it seems like its a new thing. Also not used yet?

server-phoenix/lib/helios/slack_channel_names.ex
```diff
@@ -0,0 +1,17 @@
...
+  import Ecto.Query
+
+  schema "slack_channel_names" do
+    field :channel_name, :string
+    field :channel_id, :string
+
+    timestamps()
+  end
+
+  def with_id(query, channel_id) do
```
bencates, line 14: Might make this easier to use to provide a default value for `query`

```suggestion
  def with_id(query \\ __MODULE__, channel_id) do
```

server-phoenix/lib/helios_web/controllers/web_hooks/slack_controller.ex
```diff
@@ -1,12 +1,76 @@
...
+        body = %{
+          "channel" => channel_id
+        }
+
+        request_body = URI.encode_query(body)
+
+        headers = [
+          {"Accept", "application/json"},
+          {"Authorization",
+           "Bearer #{Application.get_env(:helios, HeliosWeb.Endpoint)[:slack_bearer_token]}"},
```
bencates, line 21: This line is a mouthful, both here and in `handle/2`. Might be worth pulling out a `defp slack_bearer_token()` to improve legibility.

server-phoenix/lib/helios_web/controllers/web_hooks/slack_controller.ex
```diff
@@ -1,12 +1,76 @@
...
+
+        json = Jason.decode!(channel_response.body)
+        name = json["channel"]["name"]
+
+        Repo.insert!(%SlackChannelNames{
+          channel_id: channel_id,
+          channel_name: name
+        })
+      rescue
+        e -> IO.inspect(e)
```
bencates, line 38: This isn't really recovering from an error, it's just logging it and going on as though nothing went wrong. It's also spitting it out to stdout instead of properly logging it which may or may not get dropped in production depending on how the logs are captured and stored.

If we don't know how to programmatically recover from an error the best thing to do is probably to let it propagate up to the API layer and (hopefully) return a meaningful error message to the client. That way the user knows something broke and can react appropriately.

server-phoenix/lib/helios_web/controllers/web_hooks/slack_controller.ex
```diff
@@ -1,12 +1,76 @@
...
+        headers = [
+          Authorization:
+            "Bearer #{Application.get_env(:helios, HeliosWeb.Endpoint)[:slack_bearer_token]}"
+        ]
+
+        Task.Supervisor.start_child(
+          SlackImageDownloader,
+          fn ->
+            try do
+              response = HTTPoison.post!(url, [], headers, follow_redirect: true)
```
bencates, line 64: `response` is never used. We can just fire off `HTTPoison.post!/4`.

server-phoenix/lib/helios_web/controllers/web_hooks/slack_controller.ex
```diff
@@ -1,12 +1,76 @@
...
+            rescue
+              e -> IO.inspect(e)
```
bencates: Again, not sure what we're rescuing here. If this isn't caught it'll propagate up to the `Task.Supervisor` which will log the crash for us.

bencates (commented):
This looks pretty good! I added a few code cleanup suggestions.

cathyzoller:
Is this PR a fork of your #417 branch?  If so, please note that in the description.  Doesn't look like this should be merged before #417, so perhaps also add a do not merge label until the first pr is merged.

wogara:
> Is this PR a fork of your #417 branch? If so, please note that in the description. Doesn't look like this should be merged before #417, so perhaps also add a do not merge label until the first pr is merged.

no, the #417 PR is a fork of this branch

chewnoill (approved):
Nice work @wogara 🚀

PR #421, opened 2022-06-28
Add authentication for the REST API

RVRX:
This authentication for the Events API works via public key encryption. I **do not expect it to work with anyone's _current_ SSH keys, but any machine with OpenSSL can easily generate keys understood by the API**. There is also currently no way – to my knowledge – to add or update new users in the DB without directly interfacing with it.

### User Authentication Plug
The API now uses a new Authentication Plug, that should carry over to any other route that implements the pipeline `:requires_auth` with only a little modification (probably just adding the the `:raw_body` field via a body_reader parser).

Any request that goes through this pipeline is required to include: 
 - The `?user=<user_name>` query parameter, 
 - `http-auth` header with the base64 encoding of the signature of the body of the  request. Only requests with a valid signature that maps to the public key of user in the DB will be allowed to access the resource on the other end of the pipeline.

Keys must be PEM-format RSA key with SHA256 encryption, such as that generated by OpenSSL. Such a key can be generated with:
```shell
openssl genrsa -out helios.pem && openssl rsa -in helios.pem -pubout -out helios_pub.pem
```
It might be possible to generate such keys with ssh-keygen, but I did not stumble upon a valid output key.

### The post-commit Hook
Might have some optimizations to make in this script, but when placed in `.git/hooks` it will fire after every commit you make in the given repository.

There are two variables near the top of this script that should be modified to point to your private key and a Helios instance.

**Note**: when in a Helios development environment, with the hook pointing to your localhost you might receive some noisy console output upon committing when the repository state is such that errors are produced by Phoenix, as Phoenix will respond with an apparently unsilenceable HTML response stack-trace.


cathyzoller (commented):
Nice work... needs a rebase before approval, and I have just a couple of comments.

server-phoenix/lib/helios_web/plugs/user_authentication.ex
```diff
@@ -0,0 +1,80 @@
...
+
+    if is_valid do
+      conn
+    else
+      send_resp(conn, 401, "Unauthorized") |> halt()
+    end
+  end
+
+  defp get_user_name(conn) do
+    user_name = conn.params["user"]
```
cbarber, line 53: Could we put required fields in the header? Would this simplify the git hook?
```
Authorization: Signature user="user",signature="base64-payload"
```

server-phoenix/test/helios_web/controllers/events_api/events_api_controller_test.exs
```diff
@@ -3,148 +3,151 @@ defmodule HeliosWeb.WebHooks.EventsAPIControllerTest do
...
+      public_key: File.read!("test/helios_web/controllers/events_api/pub.pem")
+    })
+
+    # All valid except the JSON body itself
     conn_1 =
       conn
       |> put_req_header("content-type", "application/json")
+      |> put_req_header(
+        "http-auth",
+        "aXGMNYY/AetzCz2Gdmkx4VeQj5sWzEJ6cF/n0ThioysMFxnhwLjrkQJKHIjjKM6EuFu0DHQLVyvnHM6wNJylRsdT33yLFyu0+qFeAmFucCtbt7sgXsIUXHAOfKR6eTnxLjouYFW5rj3RNfcebsQZyERfPt6Jsc+A4nSk/w3M1p4PpeMOCU1IASy1eSHurpcYJlaJaRdwO5g7wbD36nygdL2lFIh/wTiOqKTo+Vys0D9a2Hcmg3dr01hNn/8sG/CLJALta3zKW763FMFJ4yJrfNgSwDB/mijl4UBBIHx0ZY0AJq8G+0lUof8gaKqAizYM9EirxH9YLaA4iI8Jo9J+Kg=="
```
cbarber, line 20: Could you add a comment at the top of the file for how to re-create these payloads from the CLI (assuming that's how you did it)?

server-phoenix/test/helios_web/controllers/events_api/pub.pem
```diff
@@ -0,0 +1,9 @@
+-----BEGIN PUBLIC KEY-----
+MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEA3ETTMJbnAeVPi+BIqogG
+hfN0CueTVGVdLF6idoNlzEoPjY0L2uqpJX4dU8B4fTf1HGMsyU8AMPSdnnMpoTvY
+ZtQxQxkVcOFwfdQjQ648RAXqIm37F5WCgchsai/L9sN1Kg8RRvC0r0Ol8B4dLGoj
+IRdx1GFLwlXa3DqkMvCSRh5zintO7vBelgCOwhyOfpbui9iXCW4mVWiP2xTB1+Gs
+pMyaCfZriRkPb6/8RMK8PIcIW878ODUUuRSZH+guxjs8c2DL7tFTvkMGczkwXXi3
+92X6GRxWT0jNGMIUPdtWmgEUXjOBuYhuLyjHz1RqPiRtDMYfJNpu7yI1VtA2Gp1r
+xQIDAQAB
+-----END PUBLIC KEY-----
```
cbarber, line 9: Lets add the private key as well in case we need to recreate any of the above payloads (assuming these are both created just for this test). If this is your actual public key, lets switch it to one we create just for testing.

server-phoenix/test/helios_web/controllers/events_api/events_api_controller_test.exs
```diff
@@ -3,148 +3,151 @@ defmodule HeliosWeb.WebHooks.EventsAPIControllerTest do
...
+        Routes.events_api_path(conn, :handle, "1", user: "rvrx"),
+        File.read!("test/helios_web/controllers/events_api/example-pr.json")
       )
 
     assert conn.status == 201
     assert length(get_commits()) == 0
     assert length(get_prs()) == 1
   end
 
-  test "POST multiple commits (w/ duplicate IDs) and PRs", %{conn: conn} do
```
cbarber, line 101: :thinking: I feel like these were useful tests. If you add the throw away private key, can you sign them programmatically to continue to use them?

post-commit
```diff
@@ -0,0 +1,72 @@
...
+# create JSON file
+touch ${tmp_json_path}
+echo -n "{
+	\"event_type\": \"github_commit\",
+	\"author\": \"${username}\",
+	\"event_id\": \"${git_sha}\"
+}" > ${tmp_json_path}
+
+# sign the JSON file (openssl sign tmp_json_path with rsa_key_path | convert to b64)
+signature=$(openssl dgst -sha256 -sign ${rsa_key_path} ${tmp_json_path} | base64)
```
cbarber, line 58: IIRC, openssl will sign STDIN in if you omit the file name. Will let you avoid the temp file.

chewnoill (approved):

server-phoenix/lib/helios_web/plugs/user_authentication.ex
RVRX: So the username is already in the JSON body, so it looks like I can effective do the same thing as JWT (and doing what barber said) by refactoring the auth to get the username from the POST body insetad of the query param.

So now...
 - user sends POST,
 - Auth looks at POST, gets the user
 - gets that user's key
 - verify the body against the signature
Which effectively is the same as JWT except we've got an unencoded body instead of an encoded (b64) header.

server-phoenix/test/helios_web/controllers/events_api/events_api_controller_test.exs
RVRX: I kinda felt like they were redundant, but was maybe biased by the chore of not wanting to sign and encrypt anymore payloads. I'll add them back

server-phoenix/lib/helios_web/plugs/user_authentication.ex
cbarber: Sounds fine. If we want to explore standards for either the auth header (Signature has an rfc) or JWTs that can be a future iteration on this so we don't scope creep.

server-phoenix/test/helios_web/controllers/events_api/pub.pem
RVRX: added

post-commit
RVRX: yup, fixed

PR #424, opened 2022-07-05
Add Torch Interface with Authentication

RVRX:
### Implement torch as admin UI with Authentication from `phx.gen.auth`.
Most of the changes and new files in this PR are automated changes, from the `torch.gen.html` command with `--web Admin`

#### Torch page @ `<helios>/admin/events`:
image

#### Same page when not logged in:
image

### Create with mix task
run
```sh
mix create_admin "foobar@mojotech.com" "aNicePassword333"
```

### Create by manually adding to the DB:
Add a user to the `admins` table, with a valid email, hashed password (see below), and make sure the **all** the date columns have a valid [NaiveDateTimes](https://hexdocs.pm/elixir/1.12.3/NaiveDateTime.html#local_now/1) as well.

Passwords can be hashed with:
```
htpasswd -bnBC 12 "" your_password | tr -d ':\n' | sed 's/$2y/$2b/'
```

## Issues
This is not a _perfect_ Torch implementation, as there are a few issues with some of our schemas and the like. For example there are some location ID's for some tables in the DB, but no mention of that field in the schema, so torch can't currently interface with that column.


cbarber:
> Tests? Gone from under 30 to 237 tests; 99% of new tests were auto generated. While they were helpful in the dev process to make sure import were re-arranged properly, not sure they are still necessary.

Are the new generated tests testing new generated code? Is that code used anywhere besides the tests? If so, the tests are valid.

RVRX:
Getting the "Looks like we're disconnected" message ;(

server-phoenix/lib/helios/locations.ex
```diff
@@ -0,0 +1,167 @@
+defmodule Helios.Locations do
```
force10269, line 1: We also need to change the alias for Location in `seeds.exs` so `mix setup` doesn't error out

RVRX: yup thats one of the errors I just found. Gotta do a better job looking through all these imports lol.

chewnoill (approved):
🚀

PR #426, opened 2022-07-08
fix: follow compiler lint warnings

RVRX:
Fixing some compiler warnings on unused variables.

### DIFF
--- a/server-phoenix/lib/helios/slack_channel_names.ex
+++ b/server-phoenix/lib/helios/slack_channel_names.ex
@@ -1,7 +1,6 @@
 defmodule Helios.SlackChannelNames do
   use Ecto.Schema
 
-  import Ecto.Changeset
   import Ecto.Query
 
   schema "slack_channel_names" do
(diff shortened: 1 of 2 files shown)


force10269 (commented):
This may be unrelated, but can we tack on a fix for the warning we get from the unused import of `Ecto.Changeset` in `slack_channel_names.ex`? I've been getting that on compilations

RVRX:
> These fixes seem to work. This may be unrelated, but can we tack on a fix for the warning we get from the unused import of `Ecto.Changeset` in `slack_channel_names.ex`? I've been getting that on compilations

Yeah that would fit in this PR. Where/when are you seeing that warning though, @force10269 not sure I've seen it? Just wanna make sure I'm getting all the current lint errors.

force10269:
> > These fixes seem to work. This may be unrelated, but can we tack on a fix for the warning we get from the unused import of `Ecto.Changeset` in `slack_channel_names.ex`? I've been getting that on compilations
> 
> Yeah that would fit in this PR. Where/when are you seeing that warning though, @force10269 not sure I've seen it? Just wanna make sure I'm getting all the current lint errors.

I saw it when I ran `mix test` for the first time I checked into your branch, so it should be there. You can also try `mix compile --force` and there should be something there

chewnoill (approved):
🚀

PR #428, opened 2022-07-14
Fix torch navbar links

RVRX:
Makes the torch navbar links absolute paths to prevent broken links when in a torch sub-page


force10269 (approved):
💯

PR #429, opened 2022-07-18
Add handtracking cursor

RVRX:
### Hand Tracking
Adds the ability to control a cursor via webcam hand-tracking. Using a pinching gesture with your thumb and pointer will act as a 'click'.
Feature is off by default, and must be enabled by adding `HAND_TRACKING_ENABLED=true` to your `.env`.

output


app/javascript/yoha-starter.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+  const video = yoha.CreateVideoElementFromStream(streamRes.stream);
+
+  // Download models.
+  const modelFiles = await yoha.DownloadMultipleYohaTfjsModelBlobs(
+    '/assets/yoha/box/model.json',
+    '/assets/yoha/lan/model.json',
+    (rec, total) => {
+      // eslint-disable-next-line no-console
+      console.debug(`[YOHA]: Download progress: ${(rec / total) * 100}%`);
```
chewnoill, line 32: how long does this download take?

chewnoill (approved):
Really cool!

app/javascript/yoha-starter.js
RVRX: Not long at all, typically <1.5 seconds.

PR #432, opened 2022-08-01
Wo/send message for feeds

wogara:
i am using a new bearer token for a bot that I created so i can control the permissions needed. I also added some logs so that I can see what the errors may be in heroku.


chewnoill (approved):
🚀

PR #437, opened 2022-08-11
fix: typo in prettier

LeoSipowicz:
A typo in prettier was causing it not to be run. Fixed the typo and called prettier:format to update files where prettier had not been run.


cbarber (approved):
Nice job! :rocket:

PR #441, opened 2022-08-16
Wo/pr count fix

wogara:
the payload from the github webhook was not parsed correctly, this hopefully fixes that


chewnoill:
> 🤖 could use a rebase

wogara:
the bigger problem here is that helios was expecting application/json data, but the github webhook was sending x-www-urlencoded data. Another problem was that the event table "external_id" was expecting a string, but the json data delivers this id as an integer.

chewnoill (approved):
🚀

